PR #19, opened 2015-01-26
Use plural key for list responses, even if the list is 0 or 1 records long

ForSpareParts:
The Ember documentation is a little ambiguous on this point, but I would expect an endpoint that returns a list to always use the same key for its content. The client making the request has no knowledge ahead of time about the number of requests, so it wouldn't know whether to look for, say, "user" instead of "users." Or, to put it another way, if I say "give me a list of all the users," I'll be expecting a list, even if the list turns out to be empty or only one record long.

Granted, Ember Data *does* handle a singular key correctly in this case, but it still seems like it would be more consistent to use the plural key -- other clients consuming the application's API may not be as savvy.

I've also added a standalone test runner (runtests.py) to make running the library's tests a little easier -- this way, a contributor won't need to create a separate Django project to run the test suite.

Looking forward to your thoughts. Also, thanks for your great work on this project so far!

### DIFF
--- a/example/tests/test_format_keys.py
+++ b/example/tests/test_format_keys.py
@@ -36,7 +36,7 @@ class FormatKeysSetTests(TestBase):
 
         user = get_user_model().objects.all()[0]
         expected = {
-            u'user': [{
+            u'users': [{
                 u'id': user.pk,
                 u'firstName': user.first_name,
                 u'lastName': user.last_name,
@@ -47,8 +47,9 @@ class FormatKeysSetTests(TestBase):
         json_content = json.loads(response.content)
         meta = json_content.get('meta')
 
-        self.assertEquals(expected.get('user'), json_content.get('user'))
-        self.assertEqual('http://testserver/user-viewset/?page=2', meta.get('nextLink'))
+        self.assertEquals(expected.get('users'), json_content.get('users'))
+        self.assertEqual('http://testserver/user-viewset/?page=2',
+            meta.get('nextLink'))
 
     def test_pluralization(self):
         """
@@ -74,3 +75,12 @@ class FormatKeysSetTests(TestBase):
 
         json_content = json.loads(response.content)
         self.assertEquals(expected.get('users'), json_content.get('users'))
+
+        #test that the key is still pluralized when there are no records for the
+        #model, as long as the endpoint serves a list
+        get_user_model().objects.all().delete()
+        response = self.client.get(self.list_url)
+        self.assertEqual(response.status_code, 200)
+
+        json_content = json.loads(response.content)
+        self.assertEqual(json_content.get('users'), [])
--- a/rest_framework_ember/utils.py
+++ b/rest_framework_ember/utils.py
@@ -65,6 +65,6 @@ def format_resource_name(obj, name):
     Pluralize the resource name if more than one object in results.
     """
     if getattr(settings, 'REST_EMBER_PLURALIZE_KEYS', False) and isinstance(obj, list):
-        return inflection.pluralize(name) if len(obj) > 1 else name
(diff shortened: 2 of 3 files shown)


ForSpareParts:
Addendum: just realized that you're running your tests with ``django-admin.py test``, which is also fine. If you'd prefer a version of this pull request without the test runner, just say the word!

ForSpareParts:
@erichonkanen As far as I can tell, the documentation doesn't specify the expected JSON key in cases where you request a plural resource (e.g. `/users`) but the response contains 0 or 1 records. Might be something in that page I'm missing? The [API documentation](http://emberjs.com/api/data/classes/DS.RESTAdapter.html) for RESTAdapter implies that the pluralization of the key is tied to the route, rather than the number of records:

>Similarly, in response to a `GET` request for `/posts`, the JSON should look like this:
>
>    { "posts: [
>        ...

It seems to me that any request for a plural resource (e.g., `/users`) is a plural request -- if it turns out that the response contains only one user, it's still nested within a list, and I'd expect the key to reflect that it's a "list of one" rather than "one on its own". For what it's worth, I've used that behavior (plural for any list-type request) in another, non-Django app and it also works perfectly.

erichonkanen:
@ForSpareParts I'm going to look into that some more.. I'm going off what I've read in the past and behavior from the apps I have using this..

erichonkanen:
@ForSpareParts I was thinking that control over the pluralization might make sense also for cases where you have an unique pluralization... e.g. /api/deer/

ForSpareParts:
@erichonkanen That does seem like a reasonable thing to have. Maybe, instead of turning pluralization on/off with a settings key, you could specify the plural on the ViewSet? You've got `resource_name` already, and a reasonable extension might be `plural_resource_name` for ViewSets where the plural is weird -- the default behavior could be that where `plural_resource_name` is unset, it just defaults to `inflection.pluralize(resource_name)`, which corresponds pretty well to ED's behavior (autopluralize unless I override it).

jerel:
My understanding of the jsonapi.org spec is that resource names should always be plural regardless if the endpoint returns a list or a single object. Check this section http://jsonapi.org/format/#document-structure-resource-collection-representations and the one right below it.

Then in the RESTSerializer's extractSingle method there's this line that leads me to believe that Ember Data is planning to use singular resource names everywhere https://github.com/emberjs/data/blob/v1.0.0-beta.14.1/packages/ember-data/lib/serializers/rest_serializer.js#L283-L284

Is that the way you see it too? Fwiw I'm running a fairly recent version of ED with our pluralization setting turned on and it works well.

jerel:
@ForSpareParts thanks for this PR, I merged it to `develop`. I had misread the diff the first time... I do agree that we should always pluralize if returning a list and the setting is enabled regardless of the count. In fact it looks like Ember Data will handle pluralized keys for an object also. However they haven't updated documentation to reflect that so we can still leave the `if isinstance(obj, list):` check in.

I'm still open to discussion on how to handle literal resource names that shouldn't be inflected. Perhaps we can open an issue for that. I'm leaning toward "if a literal string `resource_name` is set on a view don't inflect it"

ForSpareParts:
@jerel Thanks for merging!

With regards to non-inflected resource names, I'm still in favor of the solution I proposed above: use a special property `plural_resource_name` to indicate the desired plural (and if that property isn't found, inflect the singular name). That way, it's still easy to have a non-inflected resource (just set the `plural_resource_name` to be the same as the singular name), but there's no "special case" behavior where resource names are treated differently based on whether or not they were derived from a model.

jerel:
@ForSpareParts I can see the reasoning behind that. I'll accept a PR that adds that functionality or I'll get to it myself after a bit

PR #21, opened 2015-01-30
Removed `name` variable to fix unbound issue

erichonkanen:


erichonkanen:
@jerel pushed fix for unbound `name` that I forgot to update

PR #33, opened 2015-06-12
Refactor get resource name

firebirdberlin:


jerel:
Thanks for this PR, it looks good. I left one comment for you inline.

firebirdberlin:
Sorry for the delay. Now, I simply removed the second commit. Within the code I did not find any reason of resource_name being False. I missed that the resource_name could be explicitly set to False by the calling view.  That's a point I'll have to think about in further detail. Probably it is to be discussed at a later time.

jerel:
Now that I review this PR again I realize that it removes functionality that we depend on: replacing `try...catch` with `getattr(view, 'resource_name', None)` means that we no longer make a difference between a missing `resource_name` (we then autogenerate the `resource_name`) and a `resource_name` set to `None` (we turn off formatting and output whatever data the developer sends).

Since we're focused on the JSON API version now and the old behavior is in maintenance mode I'm going to close this. Regardless thanks for the contribution!

PR #35, opened 2015-06-23
Moves package towards json-api implementation...

erichonkanen:
- Refactored renderer to properly handle formatting keys and resource names.
- Added detection of error responses and properly return those per json-api spec


erichonkanen:
@jerel This may not be ready to be merged in yet but I wanted to submit it for feedback... I decided to take a stab at adding support for returning `errors` response but go through it to see if it seems right.. Just did a quick run-through with a live app using it and it seemed to work as expected..

Also went through and cleaned up the renderer some..

lmk what you think and I'll write some tests as soon as I can

cheers,
-eric

erichonkanen:
oh btw not sure why those tests were deleted.. I think I had started out by gutting most of it and building it back up a few months ago when I first thought of working on it.. can add those back but lets also figure out what else to put in to better support json-api

PR #36, opened 2015-06-23
Moves package towards json-api implementation...

erichonkanen:
- Refactored renderer to properly handle formatting keys and resource names.
- Added detection of error responses and properly return those per json-api spec


erichonkanen:
@jerel This may not be ready to be merged in yet but I wanted to submit it for feedback... I decided to take a stab at adding support for returning errors response but go through it to see if it seems right.. Just did a quick run-through with a live app using it and it seemed to work as expected..

Also went through and cleaned up the renderer some..

lmk what you think and I'll write some tests as soon as I can

cheers,
-eric

jerel:
Thanks for this PR. The renaming and all is a step in the right direction so I merged it to the `develop` branch (let's keep `master` for stable code). I'm not sure about the rendering aspect yet... I need to think about it some more as JSON API's nested `attributes` format + top level relationships drastically changes from the way we've been rendering and parsing. It makes it simpler in some ways but we need to think through how to handle it

erichonkanen:
@jerel cool, what about the error keyname handling? I think thats important at least in the short term if we can get it in...

also do you have any reference for that attributes format?

jerel:
@erichonkanen here's a link: http://jsonapi.org/format/#document-resource-objects In the new spec `attributes` contains the data that maps to the Django model. So however we end up doing it we need to allow setting relationships, links, meta, and etc at the top level while allowing normal CRUD to be automatic

PR #52, opened 2015-08-02
Override media type + check relation.source to find relation type

bierdybard:
Fixing two issues I've discovered:

1. Override media_type on renderers.JSONRenderer as required by the JSON API 1.0 spec. Ember Data has started sending the required Accept: header causing DRF to return a 406 error code if this isn't set.

2. Encountered an issue whereby read_only relations with a source different than the field name could not resolve their resource types as the queryset field is not set when read_only is set on the field (enforced by DRF). Updated the logic to check for the 'source' field as well.

### DIFF
--- a/rest_framework_json_api/renderers.py
+++ b/rest_framework_json_api/renderers.py
@@ -29,6 +29,8 @@ class JSONRenderer(renderers.JSONRenderer):
     }
     """
 
+    media_type = 'application/vnd.api+json'
+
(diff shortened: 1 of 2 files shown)


bierdybard:
Updated to remove the relation.field_name check as checking relation.source makes this redundant; relation.source is always more correct in this instance and if it is not explicitly set it is copied from relation.field_name.

PR #59, opened 2015-08-14
For list view, ensure 'results' in data

pattisdr:
<h1> Purpose </h1>

There are instances where a list view is not sending 'results' in data so you get a KeyError in `render`. An example is when you send an `options` request on a list resource in the Django Rest Framework. This results in a 'KeyError'.
 
screen shot 2015-08-14 at 10 28 18 am
screen shot 2015-08-14 at 10 28 30 am

<h1> Changes </h1>

Ensures list view has 'results' in data

### DIFF
--- a/rest_framework_json_api/mixins.py
+++ b/rest_framework_json_api/mixins.py
@@ -10,7 +10,7 @@ class MultipleIDMixin(object):
         """
         Override :meth:``get_queryset``
         """
-        ids = dict(self.request.QUERY_PARAMS).get('ids[]')
+        ids = dict(self.request.query_params).get('ids[]')
         if ids:
             self.queryset = self.queryset.filter(id__in=ids)
         return self.queryset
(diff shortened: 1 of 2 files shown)


jsenecal:
I understand the goal behind this commit but I cant replicate the issue you are having there... Have you tried this against the latest `develop`?

pattisdr:
Yes, I tried this against the latest `develop`. With our application, on a **list** view, requesting `options` throws a KeyError.  I imagine there might be other instances where `results` aren't in data.  This check just ensures results are in data before accessing it.

jsenecal:
Strange because my installation does not throw that exception when requesting `options`

jsenecal:
@jerel I dont mind merging this - but I think that `data` might be malformed if it is missing `results` in a list view... (I never had any issues with any of the functions of DRF myself) - Let us know

PR #66, opened 2015-08-23
automated testing; django18, drf31 and drf32 only

abdulhaq-e:


jsenecal:
Wow! that is really great work and thanks for the kind comments in #67 - I think we'll have to have the tests pass though... Maybe you could rebase to a branch we could create for that ? @jerel what do you think?

abdulhaq-e:
No problem, create the -problem fixing- branch and I will rebase this branch. I can also fix some of the problems, some are quite easy. For example, the failing tests for `python 3.2` are because unicode literal are not accepted, so this is wrong `u'my_string` in `py32`. It can be easily fixed though.

PR #69, opened 2015-08-28
Introduced support for nested relationships urls with drf-nested-routers

jsenecal:
Please wait for PR https://github.com/alanjds/drf-nested-routers/pull/50 to be accepted before merging this.


jsenecal:
I think we better wait for the pypi release dont you think ?

jerel:
Oops, yes. I wasn't thinking. Anyway I'm :+1: when upstream is ready

PR #71, opened 2015-09-04
better handling for unicode, should fix some tests

abdulhaq-e:
The key for all of this is the two functions in ```test.utils``` .

I also disabled the ```test_multiple_id_mixin```. I have an almost complete solution for filtering which is compliant with the specifications.


abdulhaq-e:
So now the failing tests are for `drf24` and `drf30`. Supporting the former is more challenging. However, in my opinion, before attempting to fix those, there needs to be some refactoring. Similar code is repeated in many places and there are some major cases which the renderer can't yet handle. For example, non-paginated lists doesn't work. I fixed it but it needs to be tasted. Supporting `drf{24,30}` will be much easier then.

PR #73, opened 2015-09-08
Modify exception_handler for a string error like AuthenticationError

hnakamur:
When I specified `permission_classes = (permissions.IsAuthenticated,)` and called an API, I got a string error.
This pull request modify `exception_handler` for that case.

Could you take a look and merge this?
Thanks!


hnakamur:
Thanks for merging!

PR #75, opened 2015-09-10
Prevent django from throwing errors if url is None

jsenecal:
Also renders proper json `null` when id is None


jsenecal:
The thing is that `encoding.force_text` on None returns a string: 'None' while we want a json `null`

Also, calling `extract_id_from_url(resource.get(field_name))` would dramatically fail if `resource.get(field_name)` would happen to be None

jerel:
@jsenecal you are correct that `force_text()` returns `'None'`. Thanks for clarifying

PR #76, opened 2015-09-14
Feature to format the relation names

schtibe:
Implemented a new option to be able to format the key names in
relations.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -55,7 +55,7 @@ If you set the `resource_name` property on the object to `False` the data
 will be returned without modification.
 
 
-### Inflecting object keys
+### Inflecting object and relation keys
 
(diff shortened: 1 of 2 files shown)


jsenecal:
Why was this closed?

schtibe:
I messed up my repo. Here it is again.

jerel:
I just have two points of feedback:

1. Would it make sense to move `inflector.pluralize()` inside of `format_relation_name` so it's not repeated everywhere?
2. Add a note to [this file in the docs](/django-json-api/django-rest-framework-json-api/blob/develop/docs/usage.md) that the option exists

schtibe:
@jerel I guess that would make sense, I changed it.

jerel:
Good work, thanks for this!

PR #78, opened 2015-09-14
Fix for some error messages that were split into several messages

schtibe:
The exception handler expects the error to be a list on line 33. In my
case they were a string, which lead to the split of the string into
multiple errors containing one character


jsenecal:
Seems good to me, had similar fix in a local branch - Thanks for the PR

@jerel - Do you have any concerns about merging this ?

jerel:
This looks good to me. I'm :+1:

Issue #77, opened 2015-09-14
get_related_resource_type returns parent model name


martinmaillard:
This is a simplified version of my API:

    class Survey(models.Model):
        pass

    class Question(models.Model):
        survey = models.ForeignKey(Survey, related_name='questions')

    class SurveySerializer(serializers.ModelSerializer):
        class Meta:
            model = Survey
            fields = ('id', 'questions')
            read_only_fields = ('questions',)

    class SurveyViewSet(viewsets.ModelViewSet):
         queryset = Survey.objects.all()
         serializer_class = SurveySerializer

Now if I request a survey, the type of the questions is not correct :

    >>> GET '/surveys/1'
    {
        'data': {
            'id': '1',
            'type': 'surveys',
            'attributes': {},
            'relationships': {
                'questions': {
                    'data': [
                        {
                            'id': '1',
                            'type': 'surveys',
                            'attributes': {},
                        },
                    ],
                },
            },
        },
    }

The issue seems to be in `utils.get_related_resource_type`. I would submit a PR, but I'm not entirely sure how to fix it without breaking something else...

Also, I'm guessing that it will be necessary to be able to define a `resource_name` for related models and the relationships handling code will change, but this is still a big issue.

jsenecal:
Pull latest version from github, I fixed that I think in some commit.

On Sep 14, 2015 10:11 AM, Martin Maillard <notifications@github.com> wrote:

This is a simplified version of my API:

class Survey(models.Model):
    pass

class Question(models.Model):
    survey = models.ForeignKey(Survey, related_name='questions')

class SurveySerializer(serializers.ModelSerializer):
    class Meta:
        model = Survey
        fields = ('id', 'questions')
        read_only_fields = ('questions',)

class SurveyViewSet(viewsets.ModelViewSet):
     queryset = Survey.objects.all()
     serializer_class = SurveySerializer


Now if I request a survey, the type of the questions is not correct :

>>> GET '/surveys/1'
{
    'data': {
        'id': '1',
        'type': 'survey',
        'attributes': {},
        'relationships': {
            'questions': {
                'data': [
                    {
                        'id': '1',
                        'type': 'survey',
                        'attributes': {},
                    },
                ],
            },
        },
    },
}


The issue seems to be in utils.get_related_resource_type. I would submit a PR, but I'm not entirely sure how to fix it without breaking something else...

[...]

PR #92, opened 2015-09-17
Fix camelcase -> dasherise issue

schtibe:
Work around the fact that inflection.dasherise does only work from the
underscored state

Referring to
https://github.com/django-json-api/django-rest-framework-json-api/commit/5b6456763ed7cdd51660d5293b66cd899bee7a91#commitcomment-13266779


jsenecal:
Could that fix the behavior described in.https://github.com/django-json-api/django-rest-framework-json-api/issues/77#issuecomment-140990789 ?

PR #94, opened 2015-09-17
Extract type for each related instance

martinmaillard:
Instead of extracting the resource type for the whole related collection, extract the type for each instance individually. This way, polymorphic relationships (using something like django_polymorphic) are
supported.

This PR addresses an issue raised in https://github.com/django-json-api/django-rest-framework-json-api/issues/77

The next step will be to allow customizing the way the type is extracted for each instance, but I'm not sure how to go about that...

### DIFF
--- a/rest_framework_json_api/utils.py
+++ b/rest_framework_json_api/utils.py
@@ -164,7 +164,9 @@ def build_json_resource_obj(fields, resource, resource_instance, resource_name):
 
 
 def get_related_resource_type(relation):
-    if hasattr(relation, 'get_queryset') and relation.get_queryset() is not None:
+    if hasattr(relation, '_meta'):
+        relation_model = relation._meta.model
+    elif hasattr(relation, 'get_queryset') and relation.get_queryset() is not None:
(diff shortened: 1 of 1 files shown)


jsenecal:
Could you provide us with a way to replicate your issue and test your solution uppon it? I don't see any problem about it but I would like to make sure. 

Also, what/how exactly does that address #77 ?

martinmaillard:
It addresses this comment https://github.com/django-json-api/django-rest-framework-json-api/issues/77#issuecomment-141083344

I'll try to write a test.

jsenecal:
ha! okay I missed that comment - I have nothing against merging that... @jerel any objections ?

jerel:
Once we get the `example` app set up properly we can require tests an a PR

martinmaillard:
Was that reverted ?

jerel:
I don't believe so:
https://github.com/django-json-api/django-rest-framework-json-api/blob/develop/rest_framework_json_api/utils.py#L370

PR #105, opened 2015-09-23
Support List views as well for HyperlinkedRelatedField

jsenecal:


jsenecal:
Ooups.. Lets close this - That fix was already committed to develop.

PR #114, opened 2015-10-02
fixed rendering of non paginated responses

abdulhaq-e:


example/tests/test_generic_viewset.py
```diff
@@ -11,6 +11,16 @@ class GenericViewSet(TestBase):
     """
     Test expected responses coming from a Generic ViewSet
     """
+
+    def setUp(self):
+        super(GenericViewSet, self).setUp()
+
+        setattr(settings, 'JSON_API_FORMAT_KEYS', 'dasherize')
```
abdulhaq-e: The attributes used in the JSON were dasherized so this is needed. Previously, the tests passed by coincidence due to another test changing the format to 'dasherize'.

rest_framework_json_api/pagination.py
```diff
@@ -109,6 +109,9 @@ class PageNumberPagination(BasePagination):
     A json-api compatible pagination format
     """
 
+    page_size_query_param = 'page_size'
+    max_page_size = 100
+
```
jsenecal: Why is this there ?
> ```python
# Default is 'None'. Set to eg 'page_size' to enable usage.
page_size_query_param = None
```
> ``` python 
# Set to an integer to limit the maximum page size the client may request.
# Only relevant if 'page_size_query_param' has also been set.
max_page_size = None
```

From `BasePagination` class....

example/tests/test_pagination.py
```diff
@@ -6,19 +6,6 @@
...
-    author = author_factory(name="Joel Spolsky")
-    entry = entry_factory(
-        headline=("The Absolute Minimum Every Software Developer"
-                  "Absolutely, Positively Must Know About Unicode "
-                  "and Character Sets (No Excuses!)"),
-        blog__name='Joel on Software',
-        authors=(author, )
-    )
-
-
```
jsenecal: Why removing this test ?

abdulhaq-e: This fixtures and the new fixture I used were moved to `conftest.py`. It's better than polluting the test files.

rest_framework_json_api/pagination.py
abdulhaq-e: I moved those from the settings. Pagination configuration from Drf >= 3.1 should be set in the pagination class. Putting them in the view or settings as per Drf < 3.1 now produces a 'deprecation warning' and I think it will produce an error from Drf 3.3.

jsenecal: HA! good catch ! Thanks for that!

example/tests/test_pagination.py
jsenecal: oops - my bad :+1:

jsenecal:
I'm :+1: on this one @jerel, all yours :)

Issue #126, opened 2015-10-09
Included compound documents are being included recursively


jsenecal:
On a model having a relation with itself, including the relation will have the consequence of including the fields in the relation itself.

http://localhost:8000/api/contacts/organizations/534?format=vnd.api%2Bjson&include=resellers,people&fields[organization]=resellers,people&fields[person]=full_name
```json
{
    "data": {
        "type": "organization",
        "id": "534",
        "attributes": {},
        "relationships": {
            "resellers": {
                "data": [
                    {
                        "type": "organization",
                        "id": "4"
                    }
                ],
                "links": {
                    "self": "http://api-dev.zerofail.com:8080/api/contacts/organizations/534/relationships/resellers?format=vnd.api%2Bjson",
                    "related": "http://api-dev.zerofail.com:8080/api/contacts/organizations/534/resellers?format=vnd.api%2Bjson"
                },
                "meta": {
                    "count": 1
                }
            },
            "people": {
                "data": [
                    {
                        "type": "person",
                        "id": "628"
                    },
                    {
                        "type": "person",
                        "id": "1022"
                    }
                ],
                "links": {
                    "self": "http://api-dev.zerofail.com:8080/api/contacts/organizations/534/relationships/people?format=vnd.api%2Bjson",
                    "related": "http://api-dev.zerofail.com:8080/api/contacts/organizations/534/people?format=vnd.api%2Bjson"
                },
                "meta": {
                    "count": 2
                }
            }
        },
        "links": {
            "self": "http://api-dev.zerofail.com:8080/api/contacts/organizations/534?format=vnd.api%2Bjson"
        }
    },
    "included": [
        {
            "type": "organization",
            "id": "4",
            "attributes": {},
            "relationships": {
                "resellers": {
                    "data": [
                        {
                            "type": "organization",
                            "id": "504"
                        }
                    ],
                    "links": {
                        "self": "http://api-dev.zerofail.com:8080/api/contacts/organizations/4/relationships/resellers?format=vnd.api%2Bjson",
                        "related": "http://api-dev.zerofail.com:8080/api/contacts/organizations/4/resellers?format=vnd.api%2Bjson"
                    },
                    "meta": {
                        "count": 1
                    }
                },
                "people": {
                    "data": [
                        {
                            "type": "person",
                            "id": "111"
                        },
                        {
                            "type": "person",
                            "id": "1783"
                        },
                        {
                            "type": "person",
                            "id": "1985"
                        },
[...]

PR #127, opened 2015-10-12
Fix the inclusion of relations

schtibe:
When getting a list and including data from the models in the list, only
the related data of the first model in the list was included

Explanation:
Let's assume a Model `M` that has a foreign key `type` referencing `Type`. Now we get a list of `M` which have different types and include the types (?include=type), only the `Type` of the first reference will be included, because of line 417.

### DIFF
--- a/rest_framework_json_api/utils.py
+++ b/rest_framework_json_api/utils.py
@@ -413,14 +413,13 @@ def extract_included(fields, resource, resource_instance, included_resources):
         if not isinstance(field, (RelatedField, ManyRelatedField, BaseSerializer)):
             continue
 
-        try:
-            included_resources.remove(field_name)
-            new_included_resources = [key.replace('%s.' % field_name, '', 1) for key in included_resources]
-            relation_instance_or_manager = getattr(resource_instance, field_name)
-            serializer_data = resource.get(field_name)
-        except ValueError:
+        if field_name not in included_resources:
             # Skip fields not in requested included resources
             continue
+        
+        new_included_resources = [key.replace('%s.' % field_name, '', 1) for key in included_resources]
+        relation_instance_or_manager = getattr(resource_instance, field_name)
+        serializer_data = resource.get(field_name)
 
         if isinstance(field, ManyRelatedField):
             serializer_class = included_serializers.get(field_name)


jerel:
@jsenecal I'm :+1: on this if you want to review and merge

jsenecal:
Have you tried setting a path in the include parameter? Like so: include=blog.author otherwise I don't understand your issue

schtibe:
Table `Person`:


pk | fruit (fk to `Fruit`) | Person
----|-------------------------------| ------------
1 | 1 | John
2 | 1 | Jack
3 | 2 | Jill



Table `Fruit`:


pk | name
-----|---------
1 | apple
2 | banana

Now we do
`GET /api/v1/apple?include=type`
which is programmed to return all persons

Gives roughly following result (I wrote this from memory ):

```json
	{
		"data": [{
				"type": "persons",
				"id": 1,
				"attributes": { "name": "John" },
				"relationships": {
					"fruit": {
						"data": { "type": "fruit", "id": "1" }
					}
				}
			},
			{
				"type": "persons",
				"id": 2,
				"attributes": { "name": "Jack" },
				"relationships": {
					"fruit": {
						"data": { "type": "fruit", "id": "1" }
					}
				}
			},
			{
				"type": "persons",
				"id": 3,
				"attributes": { "name": "Jill" },
				"relationships": {
					"fruit": {
						"data": { "type": "fruit", "id": "2" }
					}
				}
			}
		# so far so good, now here it gets interesting
		],
		"included": [{
			"type": "fruit",
			"id": 1,
			"attributes": {
				"name": "apple"
			}
		}]
		# banana is missing, even though the person 3 references it
	},
```

jsenecal:
Ok, so I'm trying to understand your last comment @schtibe - why would `/api/v1/apple` return person objects ?

Anyways - I'll merge this as the change makes sense and somewhat seem to work for me too but I'll try to make tests to validate the behavior when I fix #126

schtibe:
@jsenecal oh, sorry this is an error. It would be /api/v1/person of course.

jerel:
@jsenecal I also ran into a strange bug with included records today. Could you take a look [at my PR](https://github.com/django-json-api/django-rest-framework-json-api/pull/129) and see if it's related to the behavior you were seeing?

jerel:
@schtibe I also ran into a strange bug with included records today. Could you take a look [at my PR](https://github.com/django-json-api/django-rest-framework-json-api/pull/129) and see if it's related to the behavior you were seeing?

Edit: mentioned the wrong person by accident.

PR #128, opened 2015-10-13
Revert "Fix the inclusion of relations"

jsenecal:
Reverts django-json-api/django-rest-framework-json-api#127

This broke many things using the include parameter.

@schtibe could you please document your issue in a proper 'issue' so that we can work on this ?

### DIFF
--- a/rest_framework_json_api/utils.py
+++ b/rest_framework_json_api/utils.py
@@ -413,13 +413,14 @@ def extract_included(fields, resource, resource_instance, included_resources):
         if not isinstance(field, (RelatedField, ManyRelatedField, BaseSerializer)):
             continue
 
-        if field_name not in included_resources:
+        try:
+            included_resources.remove(field_name)
+            new_included_resources = [key.replace('%s.' % field_name, '', 1) for key in included_resources]
+            relation_instance_or_manager = getattr(resource_instance, field_name)
+            serializer_data = resource.get(field_name)
+        except ValueError:
             # Skip fields not in requested included resources
             continue
-        
-        new_included_resources = [key.replace('%s.' % field_name, '', 1) for key in included_resources]
-        relation_instance_or_manager = getattr(resource_instance, field_name)
-        serializer_data = resource.get(field_name)
 
         if isinstance(field, ManyRelatedField):
             serializer_class = included_serializers.get(field_name)


schtibe:
See issue #131

schtibe:
@jsenecal Can you tell me what broke because of this commit?

jsenecal:
Infinite recursion on models with self relations. It also broke something for @jerel (please help me there @jerel)

schtibe:
Ok... tried to reproduce with a model with self relation that I have, but couldn't. Am I doing it unlike you?
My Code:

```python
class Section(models.Model):
    parent_section = models.ForeignKey('Section', null=True, blank=True)
    name           = models.CharField(max_length=200, unique=True)
    code           = models.CharField(max_length=2, unique=True)
    slot_count     = models.IntegerField(default=6)

    def __str__(self):
        return self.name


class SectionSerializer(serializers.ModelSerializer):
    class Meta:
        model  = models.Section
        fields = ('id', 'parent_section', 'name', 'code')

SectionSerializer.included_serializers = {
        'parent_section': SectionSerializer
    }
```

Fixtures:
```json
  {
    "pk": 1,
    "model": "masterdata.Section",
    "fields": {
      "parent_section": null,
      "name": "foo",
      "name_de": "foo",
      "name_fr": "fuu",
      "code": "fo"
    }
  },
  {
    "pk": 2,
    "model": "masterdata.Section",
    "fields": {
      "parent_section": 1,
      "name": "bar",
      "name_de": "bar",
      "name_fr": "brr",
      "code": "ba"
    }
  },
```

The result:
```json
{"data": {"attributes": {"code": "ba", "name": "bar"},
          "id": "2",
          "relationships": {"parent-section": {"data": {"id": "1",
                                                        "type": "sections"}}},
          "type": "sections"},
 "included": [{"attributes": {"code": "fo", "name": "foo"},
               "id": "1",
               "relationships": {"parent-section": {"data": None}},
               "type": "sections"}]}

```

jsenecal:
Well, my models are a bit more complicated than that and the actual issue might be related to issue #126 - To be quite honest I the issue appeared after merging your PR - I reverted it to test and the problem was gone - I did not investigate further - sry

I'm a bit swamped in a work-related project but I would like to write tests to help with the compound-documents feature. Let me know if you would like to help with those (it would be greatly appreciated)

jerel:
The issue I was having wasn't a `self` relationship. The bug showed up as an undefined serializer class. I will likely have time today to write tests for this issue so we can resolve it properly

jsenecal:
Oh yeah ! I remember now - yeah - the extract_included was stripping off included_serializers that shouldn't have been removed.

PR #129, opened 2015-10-13
Accessing by index on Model.objects.all() produces weird behavior

jerel:
Namely what should be a list of records like [1, 2, 3] becomes
[1, 1, 2]. The solution is to cast to a list first.

I've verified the fix with tests in my private project and I plan to write a test for this before it's merged. If someone needs it in the meantime here it is.


schtibe:
@jerel I think it's not related, since it does not change the behaviour of my project.

jerel:
Oops, commented on your issue asking you to retest before I saw this comment.

Issue #133, opened 2015-10-15
tests fail on Django 1.9


jamesturk:
seems mostly related to some internal refactoring of related descriptors, PR incoming

PR #134, opened 2015-10-15
start testing on Django 1.9a1 and get tests passing

jamesturk:
* added appropriate lines to travis/tox for Django 1.9 on Py2.7 and 3.4 (3.2/3.3 not supported by Django 1.9)
* fixed two issues caused by https://github.com/django/django/commits/stable/1.9.x/django/db/models/fields/related_descriptors.py changes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,3 +25,7 @@ env:
   - TOXENV=py33-django18-drf32
   - TOXENV=py34-django18-drf31
   - TOXENV=py34-django18-drf32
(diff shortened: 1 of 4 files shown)


jamesturk:
fixes #133

jsenecal:
wow! that is great work @jamesturk! I'm all :+1: on this @jerel  if you want to merge ...

jsenecal:
I think that contains code from #134 right ?

PR #135, opened 2015-10-15
test on Python 3.5

jamesturk:
* required bumping the pytest version, doesn't seem to cause any issues
* tox base python set to 3.5 so it's available for tox

(tests are getting long, probably OK to drop Python 3.2 support, maybe 3.3 as well?)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: python
+python: 3.5
 sudo: false
 install: pip install tox
 script: tox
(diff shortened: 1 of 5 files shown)


jamesturk:
(also realized it might be OK to drop Django 1.6 support as security updates ended in April)

jsenecal:
I think that contains code from #134 right ?

jamesturk:
yep, this is a superset of #134

jerel:
Thanks for this! Excellent work.

PR #150, opened 2015-10-29
django-rest-framework 3.3 support/tests

jamesturk:
from what I can tell this was all that was needed for django-rest-framework 3.3 support


jamesturk:
actually may have spoken too soon... looking into an issue w/ pagination that I'm getting

jamesturk:
nope, my mistake (had some stale settings from DRF 3.1 style pagination), all should be good

PR #151, opened 2015-11-04
ResourceRelatedField now accepts serializer methods which return querysets.

rollokb:
Specifying the source of a field to be a method on a serializer, only works if `many=False`.  This pull request allows you to use serializer methods for `ResourceRelatedField` which return multiple objects.

Which allows you to do things like this:

    class EntrySerializer(serializers.ModelSerializer):
        included_serializers = {
            'approved_comments': 'example.serializers.EntrySerializer',
        }
        approved_comments = relations.ResourceRelatedField(source='get_approved_comments',
                many=True, model=Comment, read_only=True)

        def get_approved_comments(self, obj):
            return self.comments.approved()

### DIFF
--- a/example/serializers.py
+++ b/example/serializers.py
@@ -29,13 +29,19 @@ class EntrySerializer(serializers.ModelSerializer):
     suggested = relations.ResourceRelatedField(
             source='get_suggested', model=Entry, read_only=True)
 
+    all_comments = relations.ResourceRelatedField(source='get_all_comments',
+            many=True, model=Comment, read_only=True)
+
+    def get_all_comments(self, obj):
+        return Comment.objects.all()
+
     def get_suggested(self, obj):
         return Entry.objects.exclude(pk=obj.pk).first()
 
     class Meta:
         model = Entry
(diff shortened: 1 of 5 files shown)


jsenecal:
@rollokb Please answer to @jerel 's questions/comments or we will otherwise have to close this PR.

Thanks :)

rollokb:
@jerel sorry I just saw these in my email. I'll get back to you by the evening (UTC).

jsenecal:
@rollokb any news on this ?

cpa:
If the original author is fine with it, I'd like to take over this PR.

@jerel, which part of the PR are you referencing in this comment?
> The way our ResourceRelatedField was implemented before wasn't ideal (source was ambiguous between serializer and model) so this functionality has been moved to the more explicitly named SerializerMethodResourceRelatedField since you opened this PR.

Thanks!

jerel:
@cpa the functionality should ideally be moved into `SerializerMethodResourceRelatedField` with `many=True` instead of adding `ManyResourceRelatedField` like this PR does.

Issue #74, opened 2015-09-09
Serializer `resource_name` for nested objects is not used


mkohram:
Nested serializers don't use `resource_name` for the `type`. In the example below the `type` for the child would show something like `nestedmodels`.
```python
class NestedSerializer(serializers.ModelSerializer):
    resource_name = "pretty_nested"
    class Meta:
        model = NestedModel

class ParentSerializer(serializers.ModelSerializer):
    nested = NestedSerializer(source='some_field', many=True)
    class Meta:
        model = ParentModel
```

Looking at the code it looks like in `utils.py` the only function that performs the correct checks to get the `type` is `get_resource_name()`. All the other functions simply do something like `relation_type = inflection.pluralize(relation_model.__name__).lower()` without checking for a `resource_name` on the serializer as the docs suggest. I could submit a PR if it makes sense to put these checks in place everywhere. Ideally a function like `get_related_resource_type()` should act as a getter for `type` so that we don't see these inconsistencies.

topaxi:
We just stumbled upon the `relation_type` too, we use Ember.js for our frontend and create models based on the `type` property.

It would be very beneficial to us, if we could configure how the `relation_type` is formatted, very similar to `JSON_API_FORMAT_KEYS`. Just lowercasing the model name seems to remove information on the actual model name, a dasherized, underscored or camelcased `type` would be much better.

Otherwise we're forced to use "single worded" models.

jsenecal:
Thanks for the suggestion @topaxi - I'm quite swamped however lately so unless @jerel has some time to spare, we are open to PRs ;)

jerel:
This should be fixed by #76 when it is merged.

mkohram:
#76 does centralize the formatting but still `relation_model.__name__` is being used everywhere disregarding `resource_name` on serializers. The docs say 
> You may manually set the `resource_name` property on views or **serializers** to specify the `type` key in the json output.

For example:
https://github.com/django-json-api/django-rest-framework-json-api/blob/3498201af79dad06c497ffbc3f128985c21cade7/rest_framework_json_api/utils.py#l298

The serializer `resource_name` is not looked for when creating the `type` string.

jerel:
Thanks for the comment on `relation_model.__name__`. I'll close this and we'll add `resource_name` capabilities on the serializer as part of fixing #77

mkohram:
Sweet. Let me know if I can help!

jsenecal:
[...]

PR #152, opened 2015-11-05
Added the ability to specify a resource_name on Models

scottfisk:
I mentioned this yesterday on issue #74  and figured I would just open a PR to get opinions on this change. 

This would allow a `resource_name` to be specified on the model. This was required in certain cases (relations: [this line](https://github.com/django-json-api/django-rest-framework-json-api/blob/73bc6d57a1321ec4d831e2a78a1b09d1d6573308/rest_framework_json_api/relations.py#L121)) where no serializer or view could have overridden the resource_name. This would allow the `resource_name` of a model to be overridden globally for that model. 

I haven't written any tests yet as I uncovered a lot of places the model `__name__` property is used, hence WIP. 

EDIT:
Test  TODOs:
- [x] Model resource_name used on a relation
- [x] Model resource_name in included document
- [x] Serializer resource_name override over Model resource_name
- [x] View resource_name override over Model resource_name

### DIFF
--- a/example/models.py
+++ b/example/models.py
@@ -35,18 +35,6 @@ class Author(BaseModel):
 
 
 @python_2_unicode_compatible
-class RenamedAuthor(Author):
-    class Meta:
-        proxy = True
-
-    class JSONAPIMeta:
-        resource_name = "super-author"
-
-    def __str__(self):
-        return self.name
-
-
-@python_2_unicode_compatible
 class Entry(BaseModel):
     blog = models.ForeignKey(Blog)
     headline = models.CharField(max_length=255)
--- a/example/models.py
+++ b/example/models.py
@@ -35,6 +35,18 @@ class Author(BaseModel):
 
 
 @python_2_unicode_compatible
+class RenamedAuthor(Author):
+    class Meta:
+        proxy = True
+
+    class JSONAPIMeta:
+        resource_name = "super-author"
+
+    def __str__(self):
+        return self.name
+
+
+@python_2_unicode_compatible
 class Entry(BaseModel):
     blog = models.ForeignKey(Blog)
     headline = models.CharField(max_length=255)
--- a/example/serializers.py
+++ b/example/serializers.py
@@ -1,5 +1,5 @@
 from rest_framework_json_api import serializers, relations
-from example.models import Blog, Entry, Author, Comment
+from example.models import Blog, Entry, Author, Comment, RenamedAuthor
 
 
 class BlogSerializer(serializers.ModelSerializer):
@@ -42,6 +42,13 @@ class AuthorSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = Author
+        fields = ('name', 'email',)
+
+
+class RenamedAuthorSerializer(serializers.ModelSerializer):
+
+    class Meta:
+        model = RenamedAuthor
         fields = ('name', 'email',)
 
 
--- a/example/serializers.py
+++ b/example/serializers.py
@@ -1,5 +1,5 @@
 from rest_framework_json_api import serializers, relations
-from example.models import Blog, Entry, Author, Comment, RenamedAuthor
+from example.models import Blog, Entry, Author, Comment
 
 
 class BlogSerializer(serializers.ModelSerializer):
@@ -35,20 +35,13 @@ class EntrySerializer(serializers.ModelSerializer):
     class Meta:
         model = Entry
         fields = ('blog', 'headline', 'body_text', 'pub_date', 'mod_date',
-                'authors', 'comments', 'suggested',)
+                  'authors', 'comments', 'suggested',)
 
 
 class AuthorSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = Author
-        fields = ('name', 'email',)
-
-
-class RenamedAuthorSerializer(serializers.ModelSerializer):
-
-    class Meta:
-        model = RenamedAuthor
         fields = ('name', 'email',)
 
 
--- a/example/tests/integration/test_model_resource_name.py
+++ b/example/tests/integration/test_model_resource_name.py
@@ -1,0 +1,13 @@
+import pytest
+from django.core.urlresolvers import reverse
+
+from example.tests.utils import load_json
+
+pytestmark = pytest.mark.django_db
+
+
+def test_model_resource_name_on_list(single_entry, client):
+    response = client.get(reverse("renamed-authors-list"))
+    data = load_json(response.content)['data']
+    # name should be super-author instead of model name RenamedAuthor
+    assert [x.get('type') for x in data] == ['super-author'], 'List included types are incorrect'
--- a/example/tests/integration/test_model_resource_name.py
+++ b/example/tests/integration/test_model_resource_name.py
@@ -8,20 +8,27 @@ from example import models, serializers
 pytestmark = pytest.mark.django_db
 
 
-def test_model_resource_name_on_list(single_entry, client):
-    response = client.get(reverse("renamed-authors-list"))
+class _PatchedModel:
+    class JSONAPIMeta:
+        resource_name = "resource_name_from_JSONAPIMeta"
+
+
+def test_match_model_resource_name_on_list(single_entry, client):
+    models.Comment.__bases__ += (_PatchedModel,)
+    response = client.get(reverse("comment-list"))
     data = load_json(response.content)['data']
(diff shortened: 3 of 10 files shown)


scottfisk:
Turns out that it is fairly hacky to add a custom property to the `Meta` class of a Django Model. So this approach is not as clean as I thought.
Alternatives I can see right now is hooking into the existing `verbose_name` on the Meta class of a django model as the default instead of using `__name__` as is done now. (This is actually how the other python drf-jsonapi adapter determines resource name of a model) The inflection translation will have to be tweaked a bit to enable the verbose name returns names with spaces: `model type` instead of `ModelType`.

jsenecal:
@scottfisk I'm wondering if this realy is a good way to deal with the resource name.
using verbose_name seems a bit too /verbose/ for me...

scottfisk:
@jsenecal Yeah I don't really like the state this got to either. Unfortunately you can't add something like `resource_name` on `model.Meta` cleanly in Django.
Perhaps for now it is best to just make the call that we cannot rename resources at the model level. I think the implications of that is that there are quite a few places, like [relations](https://github.com/django-json-api/django-rest-framework-json-api/blob/73bc6d57a1321ec4d831e2a78a1b09d1d6573308/rest_framework_json_api/relations.py#L121), where the model name will always be used as the resource name

leifurhauks:
@scottfisk, since adding custom properties to Meta is impractical, what
about either 1) putting it in a class attribute on the model class itself
or 2) putting it in a different nested class on the model class? I've seen
django packages use both techniques as an alternative to hacking the Meta
class.

An example of 2) would be

```python
class MyModel(models.Model):
    class Meta:
        # my meta options

    class JSONAPIMeta:
        resource_name = 'custom_resource_name'
```

jsenecal:
I personally like option 2 a lot :+1:

scottfisk:
@leifurhauks Im open to updating this PR to approach 2 if the maintainers think that would be preferable.

jsenecal:
Thanks to both of you @scottfisk @leifurhauks

bierdybard:
A thought -- I see in the current PR all references to `_meta.model.__name__` have been replaced by calls to the new mechanism (I know it hasn't been updated for recent discussion).

It may be appropriate to keep the original _meta.model.__name__ as a fallback in case the `JSONAPIMeta` doesn't exist. Thoughts?

scottfisk:
@bierdybard Yeah that makes sense to me. I plan to do that when I update this PR (early this week). Keeps the fallback path basically the same as on master.

mkohram:
How about `JSONAPIMeta` taking precedence over serializer and view `resource_name`. I think I have a test that justifies this.

jerel:
Making `JSONAPIMeta` take precedence over the serializer's `resource_name` would prevent a serializer from setting its own `resource_name`, right? I use different serializers for different API authorization levels and I could imagine setting a `resource_name` of `identities` for superadmins and `users` for normal users.

jsenecal:
@jerel One could use proxy models and then have different serializers to point to them...

jerel:
@jsenecal right, proxy models would work fine if we go that route. I tend to think of models as the core with serializers next and views at the outer layer. So the natural priority seems like view > serializer > model

jsenecal:
So the lookup order shall be what exactly ?
1- Look for JSONAPIMeta in model
2- Look for resource_name in serializer
3- Look for the view name

I think it is important we discuss that in detail here.

mkohram:
Here is where the conflict occurs: A `ResourceRelatedField` gets its `type` from the model (no access to the model serializer) but when it gets `included` the `type` would come from the serializer. So now you have the same resource in *one* payload with *two* different `type`s. right? I could give an example if that just sounded like nonsense!

mkohram:
BTW, this is the case where `JSONAPIMeta` is not defined and `resource_name` is defined on the serializer.

jsenecal:
So, maybe we should drop ressource_name on serializers altother, for consistency ?

mkohram:
what about a serializer that is not tied to a model?

jsenecal:
@mkohram How would you use `ResourceRelatedField` then ?

jerel:
@jsenecal I was thinking the reverse of that:
1. Look at view
2. Look at serializer
3. Look at model

Typically it would get derived from the model unless there was a need to set it at the serializer or view level.

@mkohram that makes sense. I hadn't thought of that. I guess we could either add `resource_name` as a kwarg on `ResourceRelatedField` or always do the config on the model for simplicity

jsenecal:
@jerel That would work for me

mkohram:
@jerel Just to be clear we are still talking about keeping `JSONAPIMeta` right?

jsenecal:
@mkohram yes, I am

jerel:
@mkohram yes, I was also

mkohram:
This makes sense to me too. Defining `resource_name` on serializers will definitely generally cause other problems so if there was a way to not allow it for `ModelSerializer`, I know that's out there though!

scottfisk:
So I switched this over to using `JSONAPIMeta.resource_name` (easy change). 

I don't know what all you want me to include in this PR but I think:
The precedence of `resource_name` still needs to be figured out. Sounds like we are going to rip them out completely in favor of proxy models?
More tests for all this renamed model business
Some tox versions are failing with a `cannot import name OrderedDict` error. I don't think that was my changes and I recall something getting PRed to master a while back for that. I ll see if its as simple as a rebase or merge from `develop`

jerel:
@scottfisk I updated the `compound_documents` branch that your PR was based on so you should be able to carry on now (the merge fixed the tests). Looks like you'll have one small merge conflict to resolve in utils.py.

scottfisk:
rebased which fixed tests, fixed conflict and squashed some of the iterative tweak commits. Only thing left would be to get more coverage on edge cases w/ tests. Only thing in there now is a very basic test.

mkohram:
I think I have a good [start](https://github.com/mkohram/django-rest-framework-json-api/tree/feature/compound_documents).

8 tests up to now:

Included document and relation type match for all combinations (4)
Document type and relation type match for all combinations (4)

The other cases should be simpler to write.

scottfisk:
Alright, pretty big update on the tests (thanks @mkohram) and a couple fixes. 

I would like some feedback on my solution to the serializer overriding with includes thing. At the moment I did https://github.com/django-json-api/django-rest-framework-json-api/commit/fe0056d86321255d3512fd251089b5c579633e69. I think this covers the weird edge case by always using a serializer's resource_type even if it is not being requested as part of the `?include=`.

mkohram:
Thanks a lot for the effort. This is looking good to me. @jerel, @jsenecal need to comment. BTW, do you know why the build is failing?

jerel:
@scottfisk @mkohram thanks to both of you for your significant effort on this. It looks like the only trouble with the build is due to Django 1.7. I don't see the error right away but it looks like it's related to pagination. If you do `pip install django==1.7.11` and then `py.test -vv` on your local it should be easy to spot the diff and hopefully track it down

scottfisk:
@jerel @mkohram Ughh, fixing that test in django 1.7 was incredibly annoying to track down. I ended up just refactoring the tests we added to use `pytest-django`'s client fixture which seemed to work. It seemed something was not cleaning up properly with the DRF test client in 1.7. 

Also, is there any testing we should do on POST/PUTs. Is the `resource_name` ever used to infer the type of model being created? If so that may not be working properly.

jerel:
The parser uses `utils.get_resource_name` so it should work the same on input as it does on output.

Thanks a lot for tracking that failure down and for the rest of the work here. @jsenecal any last words? If there's no objections in the meantime I'll merge after I get back from Christmas break

jsenecal:
Great work on this everyone and Merry Christmas :) @jerel this looks great to me so I'll merge right away!

mkohram:
Thanks everyone especially @scottfisk. This wasn't straightforward. Still missing a couple of tests I believe - combinations of `resource_name` on the view/serializer/model, we only have serializer/model combinations now - I'll add them at some point. We also need some documentation for `JSONAPIMeta` but that's not news :smiley:. Happy Holidays!

mkohram:
I don't see being merged into master ...

jerel:
@mkohram it looks like the original PR targeted the `feature/compound_documents` branch and we accidentally merged to that without subsequently merging to develop. I opened a replacement PR but it has two small merge conflicts that we'll need to resolve before merging

jsenecal:
@mkohram @jerel - Woops, indeed - Thanks for #197

Issue #157, opened 2015-11-22
support nested objects


patroqueeet:
Dear,

thanks for this great module here!

Question while using it for my project:

emberjs sends JSON-API complient nested `venue` data:
```
{"data":{
 "attributes":{
  "bought-at":null,
  "sold-at":null
 },
 "venue":{
   "data":{
    "attributes":{
     "title":"Brackenweg 13, 01936 Schwepnitz, Deutschland",
     "place-id":"ChIJwzCB6gs3CEcRDw0G3kCOU34",
     "lat":51.33011,
     "lon":13.965159999999969},
    "type":"venues"
   }
  },
 "type":"landlords"}}
```

according to drf frontend it needs:
```
{"data":{
 "attributes":{
  "bought-at":null,
  "sold-at":null,
  "venue":{
   "title":"Brackenweg 13, 01936 Schwepnitz, Deutschland",
   "place-id":"ChIJwzCB6gs3CEcRDw0G3kCOU34",
   "lat":51.33011,
   "lon":13.965159999999969
  },
  "type":"venues"},
"type":"landlords"}
}
```

for my serializer:
```
from rating.models import Rating
from location.models import Venue
from utils.images import get_location_images
from landlord.models import Landlord

from rest_framework import serializers

class LandlordSerializer(serializers.HyperlinkedModelSerializer):
    venue = VenueSerializer()

    class Meta:
        model = Landlord
        fields = ('venue', 'sold_at', 'bought_at', 'id')
```

I'm running your latest 2.0.0.beta.2

I read your docu, read the code, read the example. still I'm too stupid to make it work.

I'd be extremely glad for a quick hint or a docu update.

jsenecal:
Nested objects are described in the JSON:API format [here](http://jsonapi.org/format/#document-compound-documents) but nothing stops you from using some sort of SerializerMethodField and do the rendering here.

Would that explanation help your use case ?

patroqueeet:
dear, thanks for the link. according to that, either one - what ember sends and what drf needs is neither json api compatible. so 'm not smarter now :)

but I believe the pull request I sent you will resolve that hopefully. though it does not meet json api requirements. should we drop the pull request accordingly?

best

j

jsenecal:
Sorry for the lack of follow-up. Did you resolve your issue?

patroqueeet:
[...]

PR #159, opened 2015-11-22
supporting nested objects based on serializer definition

patroqueeet:
Dear, please feel free to see and review my draft for nested data sets. Its scanning the serializer first, then checks the json api data if there is matching data. then converts and sends to drf for further handling.

rough inline comments done, no tests yet. pls provide feedback.


jsenecal:
Please fix you code so the tests pass and we'll review.

patroqueeet:
dear, will do accordingly. just needed some initial feedback that you would consider this improvement valuable. gimme a view days for that pls.

krgds

j

jerel:
Also closes #157 as this use case should already be covered by Compound Documents

Issue #155, opened 2015-11-19
Support custom object keys


orf:
Hey, thanks for your work on this project. I was investigating integrating it with our EmberJS app but I ran into an issue. We have an `Employee` model that has a normal integer primary key, but also a `username` attribute. I'd like to only expose the `username` as the primary key, using DRF this is simple:

```python
class EmployeeSerializer(serializers.ModelSerializer):
    id = serializers.ReadOnlyField(source="user.username")
    ...
```

However in `build_json_resource_obj` it just uses instance.pk:

```python
def build_json_resource_obj(fields, resource, resource_instance, resource_name):
    resource_data = [
        ('type', resource_name),
        ('id', encoding.force_text(resource_instance.pk) if resource_instance else None),
        ...
```

Can you use the serializers ID field if available, and fallback to the instances PK?

jerel:
So your `Employee` model has an integer primary key but you're overwriting it on output to use another field's (`username`) value? That sounds like a rocky road to go down as I'm pretty sure relationships will fail since you're masquerading a non primary key value as a primary key.

Would it be possible for you to change your model's `username` field to `primary_key=True` and get rid of `id` or better yet change your `id` field to a `uuid`? (the latter is what I do, Django 1.8 introduced core support for UUID pks).

I'm a little hesitant to change `build_json_resource_obj` right off because, at least in my mind, the `id` should always be a primary key. I'm open to a discussion on the matter though

orf:
I agree it sounds rocky, but in the context of the app it makes sense. We have an app with a django user model that is backed by LDAP, so the username field is the 'true' primary key and the integer id is an implementation detail for that application. The Emberjs app pulls in data from this system and two others, so exposing the primary key seems risky as its just the ID for this particular row, the username is much safer to share between the systems.

Also we view exposing the user id as a security risk, it opens the system up to easy enumeration attacks whereas the username is not susceptible without a large brute forcing effort. Upgrading the central app would be a large undertaking, and we've been using this method for a while without any issues with the DRF ember package.

Surely it's not too much to just use the `id` field specified in the serializer, it seems a weird limitation to have.

jsenecal:
How would you deal with the relations then? What is the ID used in other objects ForeignKeys?

On Nov 20, 2015 4:23 AM, orf <notifications@github.com> wrote:

[...]

PR #165, opened 2015-11-27
WIP: Issue #155 - support serializers with custom primary keys

orf:
This is a first attempt at adding support for alternative primary keys. This is done by removing any hard-coded 'instance.pk' calls, instead first checking if the PK attributes name is in any serialized output. If it is then we return that instead of the `pk` attribute. This works for the following serializers/viewsets:

``` python
class AuthorSerializer(serializers.ModelSerializer):
    id = serializers.CharField(source='email')

    class Meta:
        model = Author  # has (id, email) fields
        fields = ('id',)

class AuthorViewSet(viewsets.ModelViewSet):
    queryset = Author.objects.all()
    serializer_class = AuthorSerializer

    lookup_field = 'email'
```

This produces the following output:

``` json
 {
      "type": "Author",
      "id": "someemail@gmail.com",
      "attributes": {},
}
```

Works for foreign keys and m2m relationships as well as creating new instances (as far as I can tell). I started to write some tests but quickly got a bit lost, so I thought I would submit this for comments before the weekend and maybe tidy it up on Monday if you think it looks promising.

### DIFF
--- a/example/tests/test_serializers.py
+++ b/example/tests/test_serializers.py
@@ -70,4 +70,3 @@ class TestResourceIdentifierObjectSerializer(TestCase):
         self.assertTrue(serializer.is_valid(), msg=serializer.errors)
 
         print(serializer.data)
-
--- a/example/views.py
+++ b/example/views.py
@@ -41,4 +41,3 @@ class CommentRelationshipView(RelationshipView):
 class AuthorRelationshipView(RelationshipView):
     queryset = Author.objects.all()
     self_link_view_name = 'author-relationships'
-
--- a/rest_framework_json_api/renderers.py
+++ b/rest_framework_json_api/renderers.py
@@ -105,8 +105,8 @@ class JSONRenderer(renderers.JSONRenderer):
             for position in range(len(serializer_data)):
                 resource = serializer_data[position]  # Get current resource
                 resource_instance = resource_serializer.instance[position]  # Get current instance
-                json_api_data.append(
-                    utils.build_json_resource_obj(fields, resource, resource_instance, resource_name))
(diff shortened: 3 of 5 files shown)


grjones:
+1 For this.  There are edge cases where this is needed and would have parity with DRF serializer support for custom primary keys.

orf:
I tried to explain the issue and premise of this PR in #155 but the maintainers didn't seem to get it. I don't think this will ever be added which is a great shame as it seems pretty arbitrary. DRF supports it, but this library doesn't, because reasons.

auvipy:
@orf how about adding to https://github.com/chibisov/drf-extensions

orf:
It's not an extension to DRF, DRF already supports this. It's a bugfix for this library that seems to always assume you want to expose your `pk`

auvipy:
then try https://github.com/AltSchool/dynamic-rest this

mblayman:
Hi, @orf. Thanks for the contribution! This branch is now in conflict in the main branch and rather old. Are you interested in fixing this up or should we close out this PR? I'm stepping up to maintain the PR queue and get DRF JA changes flowing again, but I can't merge conflicted stuff. Also, this branch is marked as a work in progress so I'm not sure if it was ever finished.

If this is still valuable, resolving the merge conflicts will give me a chance to review it well. If this is no longer important to you, closing the PR is right next step.

Please let me know what you'd like to do. Thanks!

orf:
Hey @mblayman, I believe a more up-to-date patch is here: https://github.com/rpatterson/django-rest-framework-json-api/commit/6334ceb84f3dc7fe9cec9f6db89b5170315a5317

Feel free to close this MR, if @rpatterson is willing to make a MR with that change it would be great. I could contribute there rather than resurrect this branch.

PR #167, opened 2015-11-30
ResourceRelatedField edited to work with plural names

arthur-s:
example to use:
```python
members = ResourceRelatedField(
        queryset=Internal.objects.all(),
        many=True,
        related_resource_name='internals'
    )
```
When ManyToMany relation, ResourceRelatedField expects the relation type named `internal`, while my app sends to server data like 
```javascript
members: {data: [{type: "internals", id: "1"}, {type: "internals", id: "4"}]}
```
Also simplified `choices` property


jerel:
Thanks for this. I'm going to close it since #152 or http://django-rest-framework-json-api.readthedocs.org/en/latest/usage.html#inflecting-object-and-relation-keys should solve any pluralization issues with relationships.

PR #168, opened 2015-12-10
Update example for usage of relation options

ralmidani:
From my experience in my app, the pluralization of the relation key happens inside the "data" array, on the value for "type", and not on the outer "home_town" key inside "relationships".

Edit: The behavior I describe above is what the Ember-Data JSONAPIAdapter expects, so the correction is simply related to documentation, not implementation.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -192,7 +192,7 @@ Example without pluralization:
                 ...
         },
         "relationships": {
-            "home_towns": {
+            "home_town": {
                 "data": [{
                     "type": "home_town",
                     "id": 3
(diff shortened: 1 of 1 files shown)


jsenecal:
@ralmidani , the name of the relationship itself has nothing to do with pluralization. It is defined by the model fields names. In a sense, you are correct to understand that the pluralization of the relation key happens inside the "data" array, on the "type" key but this documentation fix has nothing to do with that as the actual demonstration is properly depicted in the example code ("type": "home_town" vs "type": "home_towns")

Maybe adding some other pointers could help, but not that one precisely.

Issue #169, opened 2015-12-11
Delete Relationships


Sapena:
Hey thanks for the nice work. But I've got an issue when I try to delete an relationship over the "relationships" JSON attribute. For example the following request won't delete a relationship:
```json
{
  "data": {
     "id": 1,
     "type": "model",
     "attributes": {
       "name": "test-model",
       "number": 1
     },
     "relationships": {
       "other-model": {"data": null}
     }
  }
}
```
However when I move the relationship to the attributes field, it is being deleted successfully:
```json
{
  "data": {
     "id": 1,
     "type": "model",
     "attributes": {
       "name": "test-model",
       "number": 1,
       "other-model": null
     }
  }
}
```

I've debugged the code and made some tests. The file [rest_framework_json_api/parsers.py:41](https://github.com/django-json-api/django-rest-framework-json-api/blob/develop/rest_framework_json_api/parsers.py#L41) does not handle this specific case when the `data` attribute is `None`. But as stated in the JSON API specification, the relation should be deleted when this case is given, as [here](http://jsonapi.org/format/#crud-updating-to-one-relationships) explained.

Changing the mentioned code to the following example seems to fix this issue:
```python
    @staticmethod
    def parse_relationships(data):
        relationships = (utils.format_keys(data.get('relationships'), 'underscore')
                         if data.get('relationships') else dict())

        # Parse the relationships
        parsed_relationships = dict()
        for field_name, field_data in relationships.items():
            field_data = field_data.get('data')
            if isinstance(field_data, list):
                parsed_relationships[field_name] = list(relation for relation in field_data)
            else:
                parsed_relationships[field_name] = field_data
        return parsed_relationships
```

Is this method of deleting relationships going to be supported by this framework or should an alternative method being used to delete relationships?

jsenecal:
Hey there - thanks for reporting this :)

At first I was skeptical at your explanation as you seemed to have mixed up few concepts as dealing with relationships at the resource level is different from dealing with relationships from a relationship link. [This](http://jsonapi.org/format/#crud-updating-to-one-relationships) link is about relationship links (here implemented by `RelationshipView`s)

However, since you seem to have figured a way to handle your specific issue, I would suggest you fork the project and submit a PR for reviewal. Adding some failing tests prior to this fix would also help.

Thanks :)

abdulhaq-e:
But according to the link from the specs, updating a relationship should be done by a PATCH/POST/DELETE to the relationship link. For example:

[...]

PR #171, opened 2015-12-14
Relationship deletion

Sapena:
Allow the deletion of relationships using the `relationships` values of a request, eg:

```JSON
{
  "data": {
     "id": 1,
     "type": "model",
     "attributes": {
       "name": "test-model",
       "number": 1
     },
     "relationships": {
       "other-model": {"data": null}
     }
  }
}
```

See [JSON API specification](http://jsonapi.org/format/#crud-updating-to-one-relationships)

References  #169

### DIFF
--- a/example/tests/test_views.py
+++ b/example/tests/test_views.py
@@ -146,20 +146,20 @@ class TestRelationshipView(APITestCase):
         assert response.status_code == 405, response.content.decode()
 
     def test_delete_relationship_overriding_with_none(self):
-        url = '/entries/{}'.format(self.first_entry.id)
+        url = '/comments/{}'.format(self.second_comment.id)
         request_data = {
             'data': {
-                'type': 'posts',
+                'type': 'comments',
                 'relationships': {
-                    'blog': {
+                    'author': {
                         'data': None
                     }
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


jerel:
This looks good, thanks! Would you mind adding a test to this file to demonstrate the fix? https://github.com/django-json-api/django-rest-framework-json-api/blob/develop/example/tests/test_views.py#L140-L146 You can use the highlighted test as an example.

Sapena:
Added a unit test for relationship deletion using the `relationships` attribute. I've used the `blog` field of an `Entry` instance. But as this field cannot be `Null`, the exception message is being tested and compared to "This field may not be null."

jsenecal:
This does not test the added functionality unfortunately... Any idea on what to add to the example code to have a nullable to-one relationship?

abdulhaq-e:
I think the model can be redefined inside the test function.

Sapena:
After a closer look at the `Model` classes I've found a working use case, so I changed the deletion test to use the `author` relation field from the second comment.

jerel:
Thank you!

jsenecal:
Yes, thank you :) can we close #169?

PR #172, opened 2015-12-14
Included field might not be set

lukaslundgren:
I have three models:
 * Account
 * AccountTypeA
 * AccountTypeB

AccountTypeA/AccountTypeB each have a OneToOneField to Account. Account can only be of type A or B, not both. In my AccountSerializer I've included the attributes for both the reversed relationships. I've also included the AccountTypeA/AccountTypeB serializers in included_serializers.

extract_included expects both fields to be present on every account instance when in reality each account will have either the A or B relation. This pull request makes extract_included to ignore missing fields.

### DIFF
--- a/example/factories/__init__.py
+++ b/example/factories/__init__.py
@@ -3,7 +3,7 @@ from __future__ import unicode_literals
 
 import factory
 from faker import Factory as FakerFactory
-from example.models import Blog, Author, Entry, Comment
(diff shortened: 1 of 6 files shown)


jsenecal:
Could you please update the example for that specific use case and add appropriate tests ?

lukaslundgren:
Sure, I've added my first attempt at this.

jerel:
This looks good to me now. Thanks for the PR!

PR #174, opened 2015-12-17
Make read_only=True as default on ResourceRelatedField

Anton-Shutik:
I assume that ResourceRelatedField should be in read_only mode in most cases (or even in all cases), so there no need to add read_only=True on every field entrance.

What do you think ?


Anton-Shutik:
@jerel  does this PR makes a sense for you? Lets close this if not.

PR #177, opened 2015-12-21
Use related obj for related link

Anton-Shutik:


jerel:
@Anton-Shutik thanks for this, looks good! Would you mind adding a test to prevent this from regressing in the future?

Anton-Shutik:
@jsenecal , @leifurhauks can you do a review of this PR and merge if it is ok ?

Thanks in advance

jsenecal:
I'll try to review this today. All this looks good to me but I'll need to test it on my implementation to make sure the tests are correct.

Anton-Shutik:
Yep, for sure. Actually it should work for all cases that follow json api :)

jsenecal:
Ahah - I actually meant that I need to make sure the implementation is flexible enought to allow my use-case :p

jerel:
@jsenecal do you have further comments on this PR?

jsenecal:
@jerel I lack the time to test this as much as I would like to, feel free to merge after resolving the conflicts. (Sorry)

leifurhauks:
I'm taking a look at it now

leifurhauks:
Hi @Anton-Shutik, thanks for contributing.
After reviewing your PR, I realize that there is potentially a conflict between two different use cases. Your PR makes the ResourceRelatedField generate valid links that point to the non-nested url for the related object in to-one relationships. But it introduces a problem with using nested routes.

Nested routes are used for to-many relationships, for example `/entries/<entry_pk>/comments` would be the route for all comments that have a foreign key to the entry with pk `<entry_pk>`. That behavior appears to be unchanged by your PR. 

However, examples in the JSON API spec use nested routes for to-one relationship links as well; note that in [this example](http://jsonapi.org/format/#document-resource-object-linkage) that in the author link, `"http://example.com/articles/1/author"`, the id 1 in that url is the id of the article, not of the related person, whose id is 9.

I think flat routes in relationship links may actually violate the JSON API spec:
>Additionally, a related resource link MUST NOT change because its relationship's content changes.

(See [Related Resource Links](http://jsonapi.org/format/#document-resource-object-related-resource-links))

With the flat route, the relationship link will change if the relationship is changed, i.e. if the foreign key is updated to point at a different object. With a nested route, `/articles/1/author` will always point to the author of article 1, even if the relationship is changed to point at a different person object.

In the project @jsenecal and I are working on, we are using a lot of nested routes, with the help of [drf-nested-routers](https://github.com/alanjds/drf-nested-routers). For example, an Organization may have a parent organization, which is represented by a foreign key. We have a route that maps the url `organizations/<pk>/parent` to the detail view for the parent of organization <pk>.

With the develop branch, the `parent` relationship object on `organizations/2` looks like this:
```json
      "parent": {
        "data": {
          "type": "organization",
          "id": "1"
        },
        "links": {
          "self": "http://localhost:8000/api/contacts/organizations/2/relationships/parent",
          "related": "http://localhost:8000/api/contacts/organizations/2/parent"
        }
      }
```
Even if this organization is sold to a different parent organization, the related link will be the same: `organizations/2/parent`.

With this PR it looks like this:
```json
      "parent": {
        "data": {
          "type": "organization",
          "id": "1"
        },
        "links": {
          "self": "http://localhost:8000/api/contacts/organizations/2/relationships/parent",
          "related": "http://localhost:8000/api/contacts/organizations/1/parent"
        }
      }
```

As you can see, the related link in the second version is now pointing to the parent of organiation 1, rather than the parent of organization 2.

noazark:
@leifurhauks, you are correct. It looks like there is just some ambiguity in the added test coverage. [This line](https://github.com/django-json-api/django-rest-framework-json-api/pull/177/files#diff-42a9fa628a7b3133b75278c81bb488cdR38) shows the the wrong version, while [this line](https://github.com/django-json-api/django-rest-framework-json-api/pull/177/files#diff-42a9fa628a7b3133b75278c81bb488cdR50), just below, shows the correct version. It sounds like @Anton-Shutik (I just talked to him) will be cleaning this up. Thank you for taking the time to write such a thorough explanation!

Anton-Shutik:
@leifurhauks I've made updates, so now it generates right links according to json api spec. Would you like to do one more review ?

Thanks in advance

jsenecal:
@Anton-Shutik looking at you tests I can tell that I feel confident that this should work as expected :+1: 

@leifurhauks let me know if you feel that those fixes are addressing the issues you raised properly

Thank you all !!

leifurhauks:
Hi @Anton-Shutik , sorry for the long delay.

This way of handling related links works, but changes the way the views must be organized. 

In my project we have a model `Departments` that has a to-many relationship to `Person`. The related link, for a department with pk 2, would be `/departments/2/people`, and the self link (for the relationship view) is `/departments/2/relationships/people`. The related link routes to the same `PersonViewSet` we use for the `/people` endpoint, where we have overwritten the `get_queryset` method to filter by department if the `department_pk` kwarg is passed to the view.

We find that this is convenient because if there is any view(set)-level logic that is specific to the person model, we only need to put it in one place. An example would be customizing the way the serializer gets instantiated.

Your way of handling the related link suggests an arrangement that is sort of the inverse of this -- to have one view(set) that handles all related links for a given model. You would probably need to write code (for example in `get_serializer` if it's a viewset) to pick the right serializer based on the name of the related field.

The reason we wrote RelationshipView under the assumption that there's one RelationshipView for a given object is mainly because the relationship endpoint only needs to serialize type names and IDs for the related objects, not the entire objects themselves; and it makes sense for the RelationshipView to "know" about the parent object, since it will be changing that object (for a to-one relationship) or its many-to-many relation(s) (for to-many relationships). The related link view, on the other hand, is more like a filter.

I think your way of doing it is okay, but it will break projects that are already set up the other way.

@jsenecal , @jerel , any thoughts?

jerel:
@jsenecal @leifurhauks if we're going to change the API we need to do it before 1.0.0 final. If there's a clear advantage to the new API we can do the change and document in the CHANGELOG.md (that we need to create). I don't feel like I've used relationships enough to give a good opinion on which is better in this PR.

leifurhauks:
Personally, I prefer the ResourceRelatedField API as it is, I find it a bit simpler in terms of the viewsets, even if it involves more configuration with respect to the routes themselves.

If we were to adopt this version I would like for it to be optional somehow, so it doesn't break existing projects (like ours).

jerel:
@Anton-Shutik @leifurhauks @jsenecal which way are we resolving this? Are you :+1: on a merge?

Anton-Shutik:
@jerel Actually I agree with @leifurhauks . His approach much easier to implement and use, while my approach will work fine just for generating urls only, and it will be tricky to create a view that should handle all related objects/entities.

Issue #154, opened 2015-11-13
Dynamic object-dependent metadata


leifurhauks:
It would be useful to be able to specify that a field or fields on a serializer should appear in the `meta` object rather than in `data['attributes']`. Perhaps a field could be marked (with a kwarg or attribute on the serializer) as metadata so that the renderer knows to put it in the meta section?

jerel:
Good question, I'm likely going to need this feature myself next week. Currently the `meta` is set by pagination and I implemented it in sort of a stop gap manner.

So there is `fields`, `read_only_fields`, and `write_only_fields` now. I could imagine a `meta_fields` property fitting in nicely. However we should think about customization: is `meta_fields` plus `SerializerMethodField` the best way to add arbitrarily deep meta structures? Should the renderer call a hook on each serializer to allow it to add/modify meta info?

jsenecal:
Actually, IMO, Using the serializer's fields in meta makes sense as the behavior of getting instance related data is what the serializer itself is made for... 

A usage I have myself is a custom `PermissionField` that inherits from the base `Field` class.

I don't think it should be too hard to write an `extract_meta` method and use the fields declared as an attribute like we did for included_serializers in that method. The `extract_attributes` would then avoid going through those.

riklaunim:
What about adding meta to object list endpoints? The pull request handles detail requests with get_root_meta but not lists.

jerel:
@riklaunim they were *supposed* to work the same. Looks like this test needs to be duplicated and modified to fetch a list instead of a single post so we can test that case. https://github.com/django-json-api/django-rest-framework-json-api/blob/develop/example/tests/integration/test_meta.py#L10

riklaunim:
I've added the test for list view. 

Also the most handy metadata for typical list/get/create/update frontend apps is under OPTIONS request in DRF - like options for choices fields or max lengths, is required etc. That meta can be fetched in code http://www.django-rest-framework.org/api-guide/metadata/#creating-schema-endpoints : 
```
meta = self.metadata_class()
data = meta.determine_metadata(request, self)
```
But it expects to be called in a view not in the serializer. So either second request (OPTIONS) is required or some way to add the most interesting part of that metadata to default metadata of normal requests.

PR #179, opened 2015-12-29
[#154] Add integration test for list view metadata.

riklaunim:
For list view the 'apiDocs': '/docs/api/blogs' part is missing in the response while as noted in #154 it supposed to be there.


jerel:
Fixed by #216 (which uses this test) thanks!

Issue #181, opened 2015-12-30
meta_fields implementation assumes that Meta.fields is explicitly defined


leifurhauks:
The implementation of the `meta_fields` functionality breaks if a serializer doesn't have an explicitly defined `fields` attribute on its `Meta`. That attribute is not required by DRF; if omitted on a `ModelSerializer` all the model fields are used, minus any fields listed in `Meta.exclude`. While it is arguably better practice to explicitly define `Meta.fields`, the implicit form should be supported for compatibility with upstream. 

The crash occurs in the `__init__` of `rest_framework_json_api.serializers.ModelSerializer`: https://github.com/django-json-api/django-rest-framework-json-api/commit/a763ace0f31217f2bd8e18f472dbb5f274d92e08#diff-8f2fdb5cb1d819dce410d6b74d0b77b6R143

I'm working on a possible fix.

leifurhauks:
I realized that creating a `ModelSerializer` without either the `fields` or `exclude` attributes is being [deprecated upstream](http://www.django-rest-framework.org/topics/3.3-announcement/#deprecations). This issue still needs to be fixed because the current implementation still breaks with serializers that use `exclude` instead of `fields`, and serializers that use the `fields = '__all__'` shortcut.

jerel:
Would it make sense to just throw an exception if `fields` is missing?

leifurhauks:
That would make it impossible to use `exclude` or `fields = '__all__'`.

I think it might be necessary to override the `get_fields` method on `ModelSerializer`. That would at least make it possible to use `fields = '_all__'`. I'm not sure about the use of `exclude`.

PR #182, opened 2015-12-30
Fix crash related to Meta.fields / Meta.exlcude

leifurhauks:
refs django-json-api/django-rest-framework-json-api#181


jerel:
@leifurhauks I like this fix. Before I merge could you squash the commits to remove the first attempt from history?

leifurhauks:
#187 is the squashed version of this PR

PR #187, opened 2016-01-04
Fix crash related to Meta.fields / Meta.exlcude

leifurhauks:
refs django-json-api/django-rest-framework-json-api#181


leifurhauks:
replaces #182

PR #194, opened 2016-01-22
Document usage of related fields and RelationshipView

leifurhauks:


leifurhauks:
what happened with tox?

jerel:
@leifurhauks I'm guessing some sort of an outage. I restarted the build. Fwiw my recent README edit failed too

PR #200, opened 2016-02-04
Nested includes did not work

adaptivdesign:
Unless I've understood the spec incorrectly, the current implementation did not handle nested includes correctly. This fixes it for me, might want to have a look at this.


jerel:
@adaptivdesign thanks for this. Would you be able to add a test for this change? If so [this](https://github.com/django-json-api/django-rest-framework-json-api/blob/develop/example/tests/integration/test_includes.py) seems like a good place for it.

adaptivdesign:
I should have time for this next week, will do!

PR #204, opened 2016-02-10
Reverse the included relation name formatting

schtibe:
See https://github.com/django-json-api/django-rest-framework-json-api/issues/203 for more information

### DIFF
--- a/example/tests/integration/test_includes.py
+++ b/example/tests/integration/test_includes.py
@@ -50,3 +50,21 @@ def test_missing_field_not_included(author_bio_factory, author_factory, client):
     assert 'included' in data
     assert len(data['included']) == 1
     assert data['included'][0]['attributes']['body'] == bio.body
+
+def test_reverse_included(single_entry, client):
+    """Test the parsing of included names"""
+    from django.conf import settings
+
+    parse_relation       = getattr(settings, 'JSON_API_PARSE_RELATION_KEYS', None)
+    singularize_included = getattr(settings, 'JSON_API_SINGULARIZE_INCLUDE_TYPE', None)
+
+    settings.JSON_API_PARSE_RELATION_KEYS = 'underscore'
+    settings.JSON_API_SINGULARIZE_INCLUDE_TYPE = True
+
+    response = client.get(reverse('entry-list') + '?include=blogs')
+    included = load_json(response.content).get('included')
+
+    assert [x.get('type') for x in included] == ['blogs'], 'Related Blogs are incorrect'
+
+    settings.JSON_API_PARSE_RELATION_KEYS      = parse_relation
+    settings.JSON_API_SINGULARIZE_INCLUDE_TYPE = singularize_included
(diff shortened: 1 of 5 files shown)


schtibe:
I have yet to figure out how the unit testing works...

jerel:
As far as the tests go `python runtests.py` should get you up and running. Is there something specific you're wondering about? You can add a function to this file https://github.com/django-json-api/django-rest-framework-json-api/blob/develop/example/tests/integration/test_includes.py and it should just be ran with the tests. Name it similar to the ones in there and if you add `single_entry` as function parameters pytest will know to create mocks for you.

schtibe:
Right now I have problems like this http://paste.codedump.ch/view/837a4192 (regardless of the repos)

schtibe:
I think for it to test the existing code I'd have to rename all the 'comments' and 'authors' to singular.. which has an effect on the "relationships" too.. not sure if it's OK to do that?

example/tests/integration/test_includes.py
```diff
@@ -50,3 +50,21 @@ def test_missing_field_not_included(author_bio_factory, author_factory, client):
...
     assert data['included'][0]['attributes']['body'] == bio.body
+
+def test_reverse_included(single_entry, client):
+    """Test the parsing of included names"""
+    from django.conf import settings
+
+    parse_relation       = getattr(settings, 'JSON_API_PARSE_INCLUDE_KEYS', None)
+    singularize_included = getattr(settings, 'JSON_API_SINGULARIZE_INCLUDE_TYPE', None)
+
+    settings.JSON_API_PARSE_INCLUDE_KEYS = 'underscore'
```
jesushernandez: pytest-django provides a good way of doing this: http://pytest-django.readthedocs.org/en/latest/helpers.html#settings

birkholz:
Quote from #257:
> It seems that #204 is taking things further by also reversing pluralization and adding a setting to specify which format the keys should be reverted to. In my opinion, the effort of that PR should be expanded to the other areas that perform this behavior, namely parsing. This could then be considered a stopgap until that larger development is ready.

I suggest this PR be expanded to also apply the reversal to attribute/relationship parsing [here](https://github.com/django-json-api/django-rest-framework-json-api/blob/develop/rest_framework_json_api/parsers.py#L32).

mblayman:
Hi, @schtibe. Thanks for the contribution! This branch is now in conflict in the main branch. Are you interested in fixing this up or should we close out this PR? I'm stepping up to maintain the PR queue and get DRF JA changes flowing again, but I can't merge conflicted stuff.

If this is still valuable, resolving the merge conflicts will give me a chance to review it well. If this is no longer important to you, closing the PR is right next step.

Please let me know what you'd like to do. Thanks!

schtibe:
Hi @mblayman. I'd like to see if I can fix the the conflict, but I'm currently not even sure anymore if the issue adressed in this PR really complies with the JSON API standard. Please give me some time as I cannot check it right away.

schtibe:
Ok I don't see the possibility any time soon to investigate further.

auvipy:
will try to investigate this in near future

PR #205, opened 2016-02-12
Add LimitOffsetPagination

sarendsen:
While its not an official specification or requirement of the JSON API, DRF does provide support for limit/offset pagination and is widely used. So i think this might be an useful addition to this package, and can be used as optional paginator.

Limit and offset arguments are based on the pagination notes here: http://jsonapi.org/format/#fetching-pagination.

Examples:

    http://api.example.org/accounts/?page[limit]=100
    http://api.example.org/accounts/?page[offset]=400&page[limit]=100

 I will write some unit test once one of the maintainers is oke with this going in.

### DIFF
--- a/example/tests/unit/test_pagination.py
+++ b/example/tests/unit/test_pagination.py
@@ -1,0 +1,79 @@
+from collections import OrderedDict
+
+from rest_framework.request import Request
+from rest_framework.test import APIRequestFactory
+from rest_framework.utils.urls import replace_query_param
+
+from rest_framework_json_api.pagination import LimitOffsetPagination
+
+
+factory = APIRequestFactory()
+
+
+class TestLimitOffset:
+    """
+    Unit tests for `pagination.LimitOffsetPagination`.
+    """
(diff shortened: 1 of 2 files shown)


jerel:
@sarendsen nice work! I'd be happy to merge this in so that users have the option to enable it. We've been slowly switching to pytest style tests so this would be the proper place to add new tests: https://github.com/django-json-api/django-rest-framework-json-api/tree/develop/example/tests/integration or https://github.com/django-json-api/django-rest-framework-json-api/tree/develop/example/tests/unit

jsenecal:
hmmm, what's wrong with Travis ?

jerel:
Dunno, it's passing now though so they must have had an outage.

sarendsen:
Added a very basic unit test. The paginator has a lot of shared code with DRF. So DRF tests applies as well. If more needed let me know.

jerel:
@sarendsen thanks for the contribution!

PR #210, opened 2016-02-22
Support deeply nested includes

leo-naeka:
As reported in #200, include paths can be compound of many nodes.
When requesting a deeply nested node, all intermediate nodes must be included as well.

This PR is similar to #200 but involves less code base modifications and includes tests.

### DIFF
--- a/example/factories/__init__.py
+++ b/example/factories/__init__.py
@@ -22,6 +22,8 @@ class AuthorFactory(factory.django.DjangoModelFactory):
     name = factory.LazyAttribute(lambda x: faker.name())
     email = factory.LazyAttribute(lambda x: faker.email())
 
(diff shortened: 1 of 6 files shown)


adaptivdesign:
Looks great if you ask me, skips me the hassle of writing tests ;) I will cancel mine!

leo-naeka:
@adaptivdesign Of course you're part of the discussion. I'm pleased you found this great as well.

adaptivdesign:
Credit where credit is due! I see you've implemented polymorphic support as well, I'll have to check that out. Although i'm using an own implementation (django-polymorphism). That is why I needed nested inclusions in the first place, so I could query polymorphic endpoints/relations like this:

/projects?include=artproject&include=researchproject
/author?include=projects.artproject&include=projects.researchproject

jerel:
@leo-naeka @adaptivdesign thank you both for your work on this.

PR #211, opened 2016-02-22
Support polymorphic models

leo-naeka:
Handles polymorphic resources with the correct models types.
This is transparent: no extra configuration is needed.

Should be tested a little bit more (typically on inverse relations, coming in the few next days), but I opened the PR to let anybody play with this and eventually get some feedback.


jerel:
@leo-naeka your code looks nice, I don't currently have a use for this in my own apps so I'd welcome usability feedback from the community.

Would it be better to set the polymorphic model in SETTINGS rather than detecting it in `utils.POLYMORPHIC_ANCESTORS`? Autodetecting it feels a little like magic that might trip someone up. "Why isn't polymorphism working?" followed by "Oh, I forgot to install that package". It also may be possible for someone to be using both packages.

leo-naeka:
Well, this is a pertinent question.

The arguments you raised seems not legitimate since `utils.POLYMORPHIC_ANCESTORS` checks for modules present in the environment, so if you forgot to install one of the modules, your models are not polymorphic in that sense. This is also a dynamically constructed tuple, so if both modules are available, anybody can use both seamlessly.

The main advantage of the auto-detection approach is to support these polymorphic modules without requiring extra configuration.
But this is definitely less flexible (other polymorphic modules like @adaptivdesign's one are simply ignored) and the whole thing would require extra maintenance and in some way, pollute the code.
So I'm totally open to declare these in a `JSON_API_POLYMORPHIC_ANCESTORS` setting and use this instead. That would require a little paragraph in the documentation.

adaptivdesign:
If I see this correctly the whole 'POLYMORPHIC_ANCESTORS' is for autodetection right? If one would **explictly** define a ResourceRelatedField relation, with is_polymorphic=True this would work too right?

I believe this is a classic case of implicit vs explicit behaviour. Since we explicitly have to declare serializers for each model, I think it's natural practice to explicitly declare a relation if it has different behaviour.

This also enables different behaviour when dealing with list vs detail api calls. I personally can imagine situations where you do not want to access entities in a polymorphic manner when performing list calls, only when retrieving the details for an entity.

The is_polymorphic flag can even be renamed to something like resource_type_from_instance.

adaptivdesign:
Ok, ofcourse what i mentioned only applies to polymorphic relations, I see that 'POLYMORPHIC_ANCESTORS' is ofcourse also used by the renderer. I believe this can be made explicit as well. As the method build_json_resource_obj is called at line:

294, 312, 443, 458 and all those lines have access to a serializer instance. Thus it should be possible to access a 'resource_type_from_instance' attribute defined in the serializer.

I'm willing to make a pull request if you guys want.

stianpr:
Any progress on this?

We need to use use relations with polymorphism our selves, but after a bit of testing this branch we noticed that it does not handle incoming data? It seems that the method `ResourceRelatedField.to_internal_value` still determines the expected resource name from the queryset. Which is the base model in many polymorphism cases. This needs to be sorted out based on the instances in some way. I think you should add some tests for receiving data with relations and make sure that it works.

I'm willing to put some effort into getting this landed.

scottfisk:
:+1: on tests for the create(POST) cases. Definitely want those to get those in.

In our app we use `django-polymorphic` in python land but not the api/frontend level. We just expose each sybtype as a separate resource. We may rework our client to use polymorphic behavior if this makes it in. 

I do think the autodetection would break our current API, because some relations would be serialized to their real type and not their base type. That may be our issue, not an issue with this feature. It just really depends how people are building their API.

stianpr:
We recently made a simple field for our code to make it work with django-polymorphic. I'm a big fan of a more explicit approach with this field than doing lots of magic in `ResourceRelatedField`. 

The fundamental reason for this to work is that we need to get the model name based on the object ID with the queryset. That will make the type verification work.

```Python
from rest_framework_json_api.exceptions import Conflict
from rest_framework_json_api.utils import get_resource_type_from_instance


class PolymorphicRelatedField(ResourceRelatedField):
    """Polymorphic Related Field.

    A field to ignore failed type verification in
    `ResourceRelatedField.to_internal_value`. Rest Framework JSON API does not
    support Django-polymorphic yet. Its type verification is based on
    retrieving resource name from `queryset.model` on the field. In our case
    that will always return the base model name, which will fail for sub models.
    """

    def to_internal_value(self, data):
        """Override to ignore `Conflict` exception.

        Since the resource name is retrieved from `queryset.model` on the
        field, we make sure it's ignored.
        """

        try:
            return super().to_internal_value(data)

        except Conflict:
            obj = self.queryset.filter(id=data['id']).first()
            if data['type'] != get_resource_type_from_instance(obj):
                raise

            # Luckily the Conflict exception is raised at the end so we can
            # explicitly call super on `ResourceRelatedField`, which is the
            # `PrimaryKeyRelatedField`.
            return super(ResourceRelatedField, self).to_internal_value(
                data['id'])
```

`to_representation()` will just work out of the box because django-polymorphic returns the correct sub type of the model. What do you think?

leo-naeka:
@stianpr I'm back on this PR.
I also prefer handling polymorphism with an *opt-in* approach, that's why there's both `PolymorphicResourceRelatedField` and `PolymorphicSerializerMetaclass` and not so much magic put in `ResourceRelatedField` (the strict minimum to be inherited).
Would be much appreciated if anyone involved with polymorphic models plays a bit with this implementation and gives feedback (doc has been updated).

scottfisk:
I use polymorphic and stated in my precious comment I like _opt-in_.

stianpr:
Hmm, I'm still not convinced of why we need the `JSON_API_POLYMORPHIC_ANCESTORS` setting? Is it absolutely necessary to have the code in `ResourceRelatedField.__init__`? Can't we have all the logic in the `PolymorphicResourceRelatedField`?

rest_framework_json_api/relations.py
```diff
@@ -47,6 +48,12 @@ def __init__(self, self_link_view_name=None, related_link_view_name=None, **kwar
 
         super(ResourceRelatedField, self).__init__(**kwargs)
 
+        # Determine if relation is polymorphic
+        self.is_polymorphic = False
+        model = model or getattr(self.get_queryset(), 'model', None)
+        if model and issubclass(model, POLYMORPHIC_ANCESTORS):
+            self.is_polymorphic = True
+
```
leo-naeka: Actually that block is useless since fddb06b

rest_framework_json_api/renderers.py
```diff
@@ -382,6 +381,9 @@ def extract_root_meta(serializer, resource):
 
     @staticmethod
     def build_json_resource_obj(fields, resource, resource_instance, resource_name):
+        # Determine type from the instance if the underlying model is polymorphic
+        if isinstance(resource_instance, utils.POLYMORPHIC_ANCESTORS):
+            resource_name = utils.get_resource_type_from_instance(resource_instance)
```
leo-naeka: Afaik, all the `POLYMORPHIC_ANCESTORS` related code could be dropped if we can detect if the resource is polymorphic here.

stianpr: But doesn't this just work because Django Polymorphic's manager will resolve to the correct type? I don't see why we need code in the renderer when the types are already resolved to the correct model when using that manager...

rest_framework_json_api/relations.py
stianpr: If it is, then maybe update this PR?

ntadej:
I tried rebasing this on develop to see if it is usable in a project I'm working on currently. There are some issues with fields selection for both list and item views.

scottkidder:
Any update on where this is at?

scottkidder:
I was able to [rebase this against develop](https://github.com/scottkidder/django-rest-framework-json-api/tree/pm-rebase) without too much hassle. All drf-json-api tests are passing but I am seeing the same issue as ntadej in my tests.

`File "/home/skidder/.venvs/django_bm_engines/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 217, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/home/skidder/.venvs/django_bm_engines/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 215, in _get_response
    response = response.render()
  File "/home/skidder/.venvs/django_bm_engines/local/lib/python2.7/site-packages/django/template/response.py", line 109, in render
    self.content = self.rendered_content
  File "/home/skidder/.venvs/django_bm_engines/local/lib/python2.7/site-packages/rest_framework/response.py", line 72, in rendered_content
    ret = renderer.render(self.data, accepted_media_type, context)
  File "/home/skidder/.venvs/django_bm_engines/src/djangorestframework-jsonapi/rest_framework_json_api/renderers.py", line 459, in render
    fields = utils.get_serializer_fields(serializer)
  File "/home/skidder/.venvs/django_bm_engines/src/djangorestframework-jsonapi/rest_framework_json_api/utils.py", line 96, in get_serializer_fields
    fields = getattr(serializer.child, 'fields')
  File "/home/skidder/.venvs/django_bm_engines/local/lib/python2.7/site-packages/rest_framework/serializers.py", line 363, in fields
    for key, value in self.get_fields().items():
  File "/home/skidder/.venvs/django_bm_engines/src/djangorestframework-jsonapi/rest_framework_json_api/serializers.py", line 211, in get_fields
    serializer_class = self.get_polymorphic_serializer_for_instance(self.instance)
  File "/home/skidder/.venvs/django_bm_engines/src/djangorestframework-jsonapi/rest_framework_json_api/serializers.py", line 225, in get_polymorphic_serializer_for_instance
    return cls._poly_model_serializer_map[instance._meta.model]
AttributeError: 'list' object has no attribute '_met`

mblayman:
The bulk of this conversation seems to have died off back in September of last year. Given that was 8 months ago and there are conflicts on this branch, I'd like to consider closing this PR unless it finds a champion who can bring it to closure.

Is that ok? What needs to be done to bring this to a conclusion?

leo-naeka:
Fair enough.

I'm still running a project which involves and requires polymorphism but haven't got the time to dig into this much more...  
We've got some debate and interesting ideas here, that I'll consider when reopening this (or making a new PR) with a more complete and tested solution.

astronouth7303:
I'm working on a fixed version of this. Should I submit a new PR or PR my change's to @leo-naeka's repo?

mblayman:
@astronouth7303 I appreciate the help. I think it makes the most sense to contribute a new PR to this repository.

leo-naeka:
If you're working with the same codebase, feel free to work on my repo (I've added you as a collaborator), that way we're keeping history and debate here.

If you're re-implementing this with some inspiration from my work, I think the better is to start a new PR.

Thank you for your time spent on this.

astronouth7303:
I'm currently just trying to bring the PR up to date.

astronouth7303:
so i think i got all the merging done, but i'm having a heck of a time sussing out the problems.

At least partly because I'm getting different ones locally.

astronouth7303:
DRF 3.5 and 3.6 are already in the testing matrix and are listed as supported in the README.

How do the reviewers suggest handling this?

astronouth7303:
Is that it? Are we ready for review?

leo-naeka:
This is it @astronouth7303. Ready for review.

mblayman:
Hi, @leo-naeka and @astronouth7303. Thanks for your heroic effort on this so far. I'm going to start reviewing this (slowly, as I have time since this is quite a bit of code), but I wanted to start by noting how much I appreciate all the time that you've put into this.

To provide some context about my review, I'm coming at this as a developer who has zero experience with Django polymorphic models. I hope that will help make my feedback valuable for teaching other future users of *why* they would want to use this feature of DJA.

I look forward to seeing this feature merged as it seems like a powerful addition to DJA.

mblayman:
As an administrative comment to start, could you please add your names to the `AUTHORS` file? I would like to see all contributors recognized for their work. The AUTHORS file is alphabetical by first name.

Also, could you please add an appropriate entry to the `CHANGELOG` for version 2.3.0?

docs/usage.md
```diff
@@ -425,6 +425,50 @@ field_name_mapping = {
 ```
 
 
+### Working with polymorphic resources
+
```
mblayman: This jumps straight into *how* to use polymorphic resources. While that is appropriate for a Usage page, if this is the only place in the documentation that describes polymorphic resources, I think an introductory paragraph providing context and *why* a user would need this would be beneficial. Also, are there going to be related packages that are needed for this work? If so, can we link to them? I'm thinking about `django-polymorphic` here.

docs/usage.md
```diff
@@ -425,6 +425,50 @@ field_name_mapping = {
...
+#### Writing polymorphic resources
+
+A polymorphic endpoint can be setup if associated with a polymorphic serializer.
```
mblayman: "setup" (a noun) should be "set up" (a verb phrase) here. As a native English speaker, I apologize for the stupidity and craziness of the English language. 😄

docs/usage.md
```diff
@@ -425,6 +425,50 @@ field_name_mapping = {
...
+A polymorphic serializer take care of (de)serializing the correct instances types and can be defined like this:
```
mblayman: "take" should be "takes"

docs/usage.md
```diff
@@ -425,6 +425,50 @@ field_name_mapping = {
...
+    current_project = relations.PolymorphicResourceRelatedField(
+        ProjectSerializer, queryset=models.Project.objects.all())
+    future_projects = relations.PolymorphicResourceRelatedField(
+        ProjectSerializer, queryset=models.Project.objects.all(), many=True)
+
+    class Meta:
+        model = models.Company
+```
+
+They must be explicitely declared with the `polymorphic_serializer` (first positional argument) correctly defined.
```
mblayman: "explicitely" should be "explicitly"

docs/usage.md
```diff
@@ -425,6 +425,50 @@ field_name_mapping = {
...
+It must be a subclass of `serializers.PolymorphicModelSerializer`.
+
+<div class="warning">
```
mblayman: sigh... the downside of using Markdown for Sphinx, no directive support. Nothing to do here, just lamenting. 😢

example/models.py
```diff
@@ -86,3 +87,25 @@ class Comment(BaseModel):
 
     def __str__(self):
         return self.body
+
+
+class Project(PolymorphicModel):
```
mblayman: Nice use of the `Project` example. I see the same thing over in the `django-polymorphic` docs. That will be a nice help to those who are coming from viewing that documentation.

example/serializers.py
```diff
@@ -1,6 +1,12 @@
 from datetime import datetime
+from packaging import version
```
mblayman: `packaging` is a third party package. Could this be moved down a line to be grouped with the other third party packages?

example/serializers.py
```diff
@@ -115,3 +121,40 @@ class Meta:
...
+        ProjectSerializer, queryset=Project.objects.all(), many=True)
+
+    included_serializers = {
+        'current_project': ProjectSerializer,
+        'future_projects': ProjectSerializer,
+    }
+
+    class Meta:
+        model = Company
+        if version.parse(rest_framework.VERSION) >= version.parse('3.3'):
```
mblayman: Interesting... I had no idea it was possible to use an `if` statement in a class definition. I learned something new about Python today. Thanks!

rest_framework_json_api/relations.py
```diff
@@ -224,6 +225,43 @@ def get_choices(self, cutoff=None):
         ])
 
 
+class PolymorphicResourceRelatedField(ResourceRelatedField):
```
mblayman: Since this is one of the primary APIs that is exposed to support polymorphic models, I think it should have a docstring describing its function/purpose and any other things that a potential user needs to know.

rest_framework_json_api/serializers.py
```diff
@@ -163,3 +167,140 @@ def get_field_names(self, declared_fields, info):
...
+            raise NotImplementedError(
+                "A PolymorphicModelSerializer must define a `polymorphic_serializers` attribute.")
+        serializer_to_model = {
+            serializer: serializer.Meta.model for serializer in polymorphic_serializers}
+        model_to_serializer = {
+            serializer.Meta.model: serializer for serializer in polymorphic_serializers}
+        type_to_serializer = {
+            get_resource_type_from_serializer(serializer): serializer for
+            serializer in polymorphic_serializers}
+        setattr(new_class, '_poly_serializer_model_map', serializer_to_model)
```
mblayman: Why is this (and the following lines) using `setattr` instead of `new_class._poly_serializer_model_map = serializer_to_model`?

rest_framework_json_api/serializers.py
```diff
@@ -163,3 +167,140 @@ def get_field_names(self, declared_fields, info):
...
+        """
+        received_type = data.get('type')
+        expected_types = self.get_polymorphic_types()
+        if received_type not in expected_types:
+            raise Conflict(
+                'Incorrect relation type. Expected on of [{expected_types}], '
+                'received {received_type}.'.format(
+                    expected_types=', '.join(expected_types), received_type=received_type))
+        serializer_class = self.get_polymorphic_serializer_for_type(received_type)
+        self.__class__ = serializer_class
```
mblayman: This line seems odd to me. Why is the code changing its `__class__` attribute?

rest_framework_json_api/utils.py
```diff
@@ -76,7 +79,11 @@ def get_resource_name(context):
     except AttributeError:
         try:
             serializer = view.get_serializer_class()
-            return get_resource_type_from_serializer(serializer)
+            if issubclass(serializer, PolymorphicModelSerializer) and \
```
mblayman: This is minor, but could we switch the order of these checks? It would be slightly more performant to check the boolean `expand_polymorphic_types` first and only do the subclass check if necessary since `and` will short circuit if the boolean is `False`.

rest_framework_json_api/utils.py
```diff
@@ -236,15 +243,11 @@ def get_related_resource_type(relation):
                     relation_model = parent_model_relation.related.related_model
                 else:
                     relation_model = parent_model_relation.related.model
-            elif parent_model_relation_type is ManyToManyDescriptor:
```
mblayman: What was the driving logic for removing these `elif` clauses? I recall reviewing a PR fairly recently that added some of these cases and I would like to make sure we're not regressing here.

mblayman:
The code looks pretty good to me. My primary concern is that addition of polymorphic support does not disrupt package users who have no need for polymorphic models. As far as I can tell, you've achieved that.

For the actually functionality, there are tests (hurray!), and I presume that you're scratching your own itch and have made this work for polymorphic models.

I think if we can do a bit of cleanup and extend some documentation at spots, this will be ready to merge. Thanks @leo-naeka and @astronouth7303!

leo-naeka:
Thank you for the review @mblayman, I'll work on it today.

example/serializers.py
leo-naeka: Yep, for the record, `packaging` is now a requirement of the recent `setuptools` (coming since python 3.5).

I've considered `distutils.version` which is built in but its classes are not documented (cf. https://docs.python.org/2/distutils/apiref.html#module-distutils.version) and seems less powerful than `packaging.version`

After all that code could be dropped as soon as we drop support of DRF 3.1 and 3.2.

rest_framework_json_api/serializers.py
leo-naeka: No reasons actually: first I was considering formatting the name depending on the serializer but it would have been pointless and overkill. Later, I used `setattr` for consistency...

rest_framework_json_api/utils.py
leo-naeka: Dunno, I suppose it's a merging issue from [0ddf5ca](https://github.com/django-json-api/django-rest-framework-json-api/pull/211/commits/0ddf5ca15867d017cce46bd1fa65990820b9b2de#diff-a16d805758b2ef29db6bd4ff77e9ccfb). Any clue @astronouth7303?

rest_framework_json_api/serializers.py
leo-naeka: That's a trick to replace the current instance class (the polymorphic "parent") by the correct subclass without altering the dynamic data that may have been associated with the instance. It is necessary to get the correct defaults (such as subclass Meta, custom validators) for the rest of the process.

I'm not very proud of that too, I'll search how this case may have been handled by other drf libraries and I'm open to any suggestions.

rest_framework_json_api/utils.py
astronouth7303: My general strategy was to apply the changes this made to the new codebase (ie, a manual rebase).

I don't recall what my logic is here, but it does look like a merge failure.

astronouth7303:
@leo-naeka Is there another way you'd like your name spelled in `AUTHORS`?

@mblayman I think that's all your code comments. However, the documentation still needs improvement.

astronouth7303:
It looks like some of the test environments never ran?

AUTHORS
```diff
@@ -3,6 +3,6 @@ Christian Zosel <https://zosel.ch>
 Greg Aker <greg@gregaker.net>
 Jamie Bliss <astronouth7303@gmail.com>
 Jerel Unruh <mail@unruhdesigns.com>
+Léo S. <leo@naeka.fr>
```
leo-naeka: All good! Thanks.

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 v2.3.0
 
+* Added support for polymorphic models via `django-polymorphic`
```
leo-naeka: Well, so actually this is only tested against `django-polymorphic`, which is the most popular and maintained polymorphic library for Django. But the implemented polymorphic support is library-agnostic and should work with libraries like `django-polymodels`, `django-typed-models` or even Django's `contrib.GenericRelation`...

I'm not sure how to deal with it, maybe an explicit warning, I'm not convinced to add more tests with more libraries... Any thoughts?

mblayman: My sense is that we should make the documentation clear that `django-polymorphic` is the only *tested* polymorphic library (thereby implying *supported* library). As far as the CHANGELOG is concerned, "via `django-polymorphic`" can be deleted.

docs/usage.md
```diff
@@ -425,6 +425,56 @@ field_name_mapping = {
 ```
 
 
+### Working with polymorphic resources
+
+Polymorphic resources (as backed by [django-polymorphic](https://django-polymorphic.readthedocs.io/en/stable/))
+allow you to easily use specialized subclasses without have to have special
```
mblayman: I think you can drop "easily" and replace "have to have" with "requiring."

docs/usage.md
```diff
@@ -425,6 +425,56 @@ field_name_mapping = {
...
+endpoints to expose the specialized versions. For example, if you had a
+`Project` that could be either an `ArtProject` or a `ResearchProject`, you can
+have both kinds at the same URL.
+
```
mblayman: I think we could put a new paragraph here that includes comments about polymorphic libraries (and remove the "as backed by" parenthetical in the previous paragraph). How does the following sound?

> DJA tests its polymorphic support against [django-polymorphic](https://django-polymorphic.readthedocs.io/en/stable/). The polymorphic feature should also work with other popular libraries like `django-polymodels` or `django-typed-models`.

astronouth7303:
Thanks for the feedback!

astronouth7303:
Is there anything else that needs to happen?

mblayman:
I thought the generic relation support oddness was still in here and I was waiting on that, but I see that was cleaned up in https://github.com/django-json-api/django-rest-framework-json-api/pull/211/commits/05cdb515b8e5e5c57c2877d471e84cea092325c8.

Thanks so much to the both of you! I really like how this all came together. I'll see if I can get the publish permissions so I can cut a new release to get this feature out there.

n1ncha:
Is there a reason that polymorphic "list" endpoints aren't supported? 

There is an exception raised by the get_fields method of PolymorphicModelSerializer that just says "("Cannot get fields from a polymorphic serializer given a queryset")" 

Also in the tests there is a "test_polymorphism_on_polymorphic_model_list_post" where you can POST to a polymorphic endpoint, but there is no "test_polymorphism_on_polymorphic_model_list" (ie. no GET project-list)

I've patched this by removing the line that throws the exception in get_fields and then using the following (I changed the models to fit the Project example)

```
class PolyListSerializer(serializers.ListSerializer):
    _poly_force_type_resolution = True

class PolymorphicProjectSerializer(serializers.PolymorphicModelSerializer):
    polymorphic_serializers = [ArtProjectSerializer, ResearchProjectSerializer]

    class Meta:
        model = ProjectSerializer
        list_serializer_class = PolyListSerializer
```

This returns a list where the individual items are listed with the proper "type", unlike before where they were all listed with the parent type.

I haven't tested this very thoroughly but thought I'd submit here for some feedback before digging any deeper.

Darkheir:
For me this merge is breaking some of my existing code.

The problem is that at the end of the parser you add the type to the parsed data:

```python
...
# Construct the return data
parsed_data = {'id': data.get('id')} if 'id' in data else {}
parsed_data['type'] = data.get('type')
parsed_data.update(self.parse_attributes(data))
...
```

In case you have a model that has a `type` attribute and that you send a request without this field it will assign the JSON API type to the field.

sliverc:
@leo-naeka As you were the initial initiator of the polymorphic support within DJA I wonder whether you are still using it nowadays? Would you mind joining our discussion on #1194 so we can get a better picture on how the polymorphic support is used nowadays? Thanks.

Anyone getting notified here and is still using polymorphic would also be great if you could join our discussion at #1194.

PR #213, opened 2016-03-06
[FEATURE]: support using get_serializer_class on view

snewcomer:
Maybe a bit too simplistic.  Wish I had an additional test to add, but then I might be overlapping django rest's tests for the `get_serializer_class` method.  Tests fail if view uses `serializer_class` rather than what is in this commit however.

This should now allow users to specify multiple actions and direct the query to different Serializers depending on the request action.


jerel:
@snewcomer thanks!

PR #216, opened 2016-03-16
fixed get_root_meta for lists

### LINKED ISSUE #198: get_root_meta should add metadata on list endpoints
Currently when viewing a list endpoint metadata can be added to each item in the `data` array but not to the root of the object.

Closes #154 and has a test case demonstrating the bug in PR #179.

adaptivdesign:
Uses tests from #179 and fixes #198

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,6 @@ pip-delete-this-directory.txt
 
 # Tox
 .tox/
+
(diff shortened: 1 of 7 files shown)


jerel:
@adaptivdesign thanks for this fix!

PR #217, opened 2016-03-21
Do not assume multiple queryset evals return in same order

grjones:
If a queryset without an order_by runs multiple times and uses an offset, the returned items are not always in the same order. This results in incorrect, missing, or duplicate data in a list view.

This PR checks to see whether the serializer.instance is a queryset. If so, it matches the serializer data to the model by pk. This might potentially be a breaking change if the serializer is a model serializer but does not expose the id field.

This PR also wraps serializer.instance as a list, so that the queryset is not evaluated many times. In other words, `some_queryset[0]` followed by `some_queryset[1]` results in two SQL queries whereas `list(some_queryset)[0]` followed by `list(some_queryset)[1]` is only one.

### DIFF
--- a/rest_framework_json_api/renderers.py
+++ b/rest_framework_json_api/renderers.py
@@ -5,6 +5,7 @@ import copy
 from collections import OrderedDict
 
 from django.utils import six, encoding
+from django.db.models.query import QuerySet
 from rest_framework import relations
 from rest_framework import renderers
 from rest_framework.serializers import BaseSerializer, ListSerializer, ModelSerializer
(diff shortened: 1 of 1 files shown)


grjones:
There is a failing test that includes a queryset that does not expose an `id` field.  I'm not sure the best way to go here.  I don't think both of these things can be true:
- No `id` on the model serializer
- A queryset that does not include an `order_by` field (since returned order is not guaranteed).
Ideally the entire rendering would be done without needing to run any additional SQL queries.

jerel:
@grjones thanks for this PR. Would you be able to add a test to demonstrate the issue you're having that this PR then fixes?

Also, off the top of my head it seems like `resource_serializer.instance` and `serializer_data` *should* be in sync since the serializer gets its data from the instance.

rest_framework_json_api/renderers.py
```diff
@@ -439,6 +440,17 @@ def render(self, data, accepted_media_type=None, renderer_context=None):
                 resource = serializer_data[position]  # Get current resource
                 resource_instance = resource_serializer.instance[position]  # Get current instance
```
jerel: This line is duplicated by line 452. One should be removed

mblayman:
Hi, @grjones! Thanks for the contribution. This PR is pretty old and has merge conflicts and fails CI. Could you update the PR so that those things are resolved? If not, feel free to close the PR.

I'm happy to review the changes when things are up-to-date. If, on the other hand, this PR is no longer appropriate, we can close it out so it doesn't sit around indefinitely.

Thanks!

mblayman:
Due to inactivity on this PR and no response within a couple of weeks, I'm going to close this PR to clean things up. If this still a relevant issue, please feel free to create a new PR with similar content in the future.

PR #221, opened 2016-04-08
Fixed get_resource_name in case of non-model backed serializer.

### LINKED ISSUE #219: The get_resource_type_from_serializer bug
django-rest-framework-json-api v2.0.0-beta.2
I have defined `resource_name` attribute into Meta class of my serializer:
```python 
class MySerializer(serializers.Serializer):
    email = serializers.EmailField(required=True)
    class Meta:
        resource_name = 'user'
```
In this case the `get_resource_type_from_serializer` will always raise AttributeError, because there isn't defined the `model` attribute.
Code of `get_resource_type_from_serializer`:
```python
def get_resource_type_from_serializer(serializer):
    return getattr(
        serializer.Meta,
        'resource_name',
        get_resource_type_from_model(serializer.Meta.model))
```

scottfisk:
Fixed get_resource_name in case of non-model backed serializer. Seems that `getattr` evaluates the default regardless of whether the default case is used ([Similar SO Question](http://stackoverflow.com/questions/31443989/python-getattr-with-another-attribute-as-the-default)). For serializers without a `meta.Model` this poses a problem. 

Closes #219

### DIFF
--- a/example/tests/unit/test_utils.py
+++ b/example/tests/unit/test_utils.py
@@ -15,6 +15,11 @@ from rest_framework_json_api.utils import get_included_serializers
 pytestmark = pytest.mark.django_db
 
 
+class NonModelResourceSerializer(serializers.Serializer):
+    class Meta:
+        resource_name = 'users'
+
+
 class ResourceSerializer(serializers.ModelSerializer):
     class Meta:
         fields = ('username',)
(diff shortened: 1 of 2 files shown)


jerel:
@scottfisk thanks! This looks good, I'll merge as soon as tests finish

PR #222, opened 2016-04-12
Correct error responses for projects with different DRF-configurations

riklaunim:
This makes it easier to haver DRF-jsonapi alongside normal DRF API. Aside of config you can already set on class, like for example:

```python

class DRFJsonApiConfigurationMixin:
    paginator_class = rest_framework_json_api.pagination.PageNumberPagination
    parser_classes = [
        rest_framework_json_api.parsers.JSONParser,
        rest_framework.parsers.FormParser,
        rest_framework.parsers.MultiPartParser,
    ]
    renderer_classes = [
        rest_framework_json_api.renderers.JSONRenderer,
        rest_framework.renderers.BrowsableAPIRenderer,
    ]
    metadata_class = rest_framework_json_api.metadata.JSONAPIMetadata
    paginate_by = 10
    paginate_by_param = 'page_size'
    max_paginate_by = 100
```

exception handler can't be set that way. To get properly formatted error response the views **handle_exception** must be overridden. The **BlogCustomViewSet** shows how it can be done now. Overwriting error response is likely not the optimal way, bu still it's better than copy-pasting the whole method to just change one line.

Also validation errors should return HTTP 422 response code ( http://jsonapi.org/examples/#error-objects-basics ) so you can set it in that method as well. Things like ember and ember-cp-validations by default expect 422 to show backend validation errors.

The **BlogCustomViewSet** should have its own tests but I couldn't force django to change REST_FRAMEWORK setting to a version with the default error handler, even with override_settings decorator. It likely needs a mock on the self.settings.error_handler in the view or something like that.

### DIFF
--- a/example/views.py
+++ b/example/views.py
@@ -4,10 +4,20 @@ from example.models import Blog, Entry, Author, Comment
 from example.serializers import (
     BlogSerializer, EntrySerializer, AuthorSerializer, CommentSerializer)
 
+from rest_framework_json_api.utils import format_drf_errors
+
 
 class BlogViewSet(viewsets.ModelViewSet):
     queryset = Blog.objects.all()
     serializer_class = BlogSerializer
+
+
+class BlogCustomViewSet(viewsets.ModelViewSet):
+    queryset = Blog.objects.all()
+    serializer_class = BlogSerializer
+
+    def handle_exception(self, exc):
+        return format_drf_errors(super(BlogCustomViewSet, self).handle_exception(exc), self.get_exception_handler_context(), exc)
 
 
 class EntryViewSet(viewsets.ModelViewSet):
--- a/example/views.py
+++ b/example/views.py
@@ -1,5 +1,10 @@
 from rest_framework import exceptions
 from rest_framework import viewsets
+import rest_framework.parsers
+import rest_framework.renderers
+import rest_framework_json_api.metadata
+import rest_framework_json_api.parsers
+import rest_framework_json_api.renderers
 from rest_framework_json_api.views import RelationshipView
 from example.models import Blog, Entry, Author, Comment
 from example.serializers import (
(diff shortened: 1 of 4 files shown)


example/views.py
```diff
@@ -4,12 +4,22 @@
...
     queryset = Blog.objects.all()
     serializer_class = BlogSerializer
 
 
+class BlogCustomViewSet(viewsets.ModelViewSet):
+    queryset = Blog.objects.all()
+    serializer_class = BlogSerializer
+
+    def handle_exception(self, exc):
+        return format_drf_errors(super(BlogCustomViewSet, self).handle_exception(exc), self.get_exception_handler_context(), exc)
```
jerel: Please split this into multiple lines, it's a bit long.

Also would you mind adding some comments to this function so devs reading this file for an example can easily understand that it is an optional override that enables JSON-API to work alongside the regular DRF format?

example/views.py
```diff
@@ -17,6 +20,8 @@ class BlogCustomViewSet(viewsets.ModelViewSet):
     serializer_class = BlogSerializer
 
     def handle_exception(self, exc):
+        if isinstance(exc, exceptions.ValidationError):
+            exc.status_code = HTTP_422_UNPROCESSABLE_ENTITY
```
jerel: This is a good example for setting it manually. jsonapi.org says they don't take a stand on 400 vs 422 so the Ember addons should really support both in my opinion.

example/views.py
riklaunim: will do. I'll also add the other settings that can be set on the class itself - as a full setup example.

example/views.py
riklaunim: There is a method in ember adapter where the code can be customized.

riklaunim:
That should do the trick :)

PR #223, opened 2016-04-12
Support many=True on related serializer method fields.

### LINKED ISSUE #220: many=True not working for SerializerMethodResourceRelatedField
I would like to use the SerializerMethodResourceRelatedField to get multiple models through a function. It works when the function returns only a single model object, but not with a list of objects.
If I try to use the many=True param, the call of the defined function fails with the error message:

` 'Entry' object has no attribute 'get_hot_entries'`

It seems the get_attribute function of the SerializerMethodResourceRelatedField is never called, when passing many=True.

Example:
```
class EntrySerializer(serializers.ModelSerializer):
  hot_entries = relations.SerializerMethodResourceRelatedField(
    source='get_hot_entries',
    many=True,
    read_only=True
  )

  included_serializers = {
    'hot_entries': 'EntrySerializer'
  }

  def get_hot_entries(self, obj):
    return Entry.objects.filter(hot=True)

  class Meta:
    model = Entry
```

I'm using django-rest-framework-json-api v2.0.0-beta.2

jerel:
Replaces #151. Fixes #220.

I'd like some review and it'd be great if you could pull this into your project and give it a test drive.

### DIFF
--- a/example/tests/integration/test_includes.py
+++ b/example/tests/integration/test_includes.py
@@ -31,7 +31,7 @@ def test_included_data_on_detail(single_entry, client):
 
 def test_dynamic_related_data_is_included(single_entry, entry_factory, client):
     entry_factory()
-    response = client.get(reverse("entry-detail", kwargs={'pk': single_entry.pk}) + '?include=suggested')
+    response = client.get(reverse("entry-detail", kwargs={'pk': single_entry.pk}) + '?include=featured')
     included = load_json(response.content).get('included')
 
     assert [x.get('type') for x in included] == ['entries'], 'Dynamic included types are incorrect'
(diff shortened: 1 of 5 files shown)


jsenecal:
I reviewed the code, while I cant test on any project I am currently working on, I don't think this should break anything :)

example/tests/integration/test_includes.py
```diff
@@ -31,7 +31,7 @@ def test_included_data_on_detail(single_entry, client):
 
 def test_dynamic_related_data_is_included(single_entry, entry_factory, client):
     entry_factory()
-    response = client.get(reverse("entry-detail", kwargs={'pk': single_entry.pk}) + '?include=suggested')
+    response = client.get(reverse("entry-detail", kwargs={'pk': single_entry.pk}) + '?include=featured')
```
vivekbhagwat: @jerel why was the test changed here? I get a failure when including a SerializerMethodResourceRelatedField, is this intentional? 

The failure is `'function' object has no attribute 'parent'`

PR #225, opened 2016-04-15
Added get_related_field_name method to RelationshipView

Anton-Shutik:


Anton-Shutik:
Sometimes I need to use other field name in the url rather that it is on e model.
For example:

```python
class Model(models.Model):
    field1 = models.CharField()
```

but relation self link should be:
api/account/123/**field2**

By that way it will be easier to override that name.

What do you think ?

jerel:
It seems good to me to extract this to a method. Could you also add some docs for others that will want to use it?

Anton-Shutik:
@jerel Just added the note

Issue #214, opened 2016-03-08
Using EXCEPTION_HANDLER is not compatible with multiple DRF configurations


riklaunim:
To get proper error responses EXCEPTION_HANDLER must be set as mentioned in #188 however EXCEPTION_HANDLER is the only setting that can't be set on a class, thus preventing multi-configuration projects -  I have vanilla-DRF and JSONApi views in the project.

The source of the problem is in https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/views.py#L413 where exception_handler = self.settings.EXCEPTION_HANDLER

I asked on DRF irc channel about it and it was suggested that the renderer itself could handle that error response formatting. Can't say how that would fit, but it's either that or slight modification to the base view class in DRF so that it's easier to change the exception handler (instead of hacking with deepcopy of self.settings or copy-pasting the whole method).

jerel:
For now I imagine that the best workaround would be to create a mixin that overrides the DRF `handle_exception` and either calls the `super` implementation or does something custom based on other view attributes.

If we moved the exception formatting to the renderer what would be the nicest API to call it with? Are you setting the DRF renderer or the DRF-JA renderer on each view class? I'm happy to review a pull request if you have time to implement something. I would think [this code could easily be extracted](https://github.com/django-json-api/django-rest-framework-json-api/blob/3c9ab1888065e54024c4de9450f0c79e4ff5e0cc/rest_framework_json_api/exceptions.py#L22-L73) to a `format_drf_errors` util method or to the renderer.

riklaunim:
I have one base class with all DRF-JA settings that every API View inherits. Will take a look at it.

riklaunim:
Looking at the DRF interface the best way would be to change it a bit - https://github.com/tomchristie/django-rest-framework/pull/4045 - if that gets merged no change to exception handle would be needed, as it could be set via method override.

jerel:
Closed by #222

PR #233, opened 2016-04-21
Allow exception handler to be used by normal DRF views

### LINKED ISSUE #218: Missing "errors" field in error response
The 400-level errors are responding with a JSON array containing error objects, but the JSON API spec [requires](http://jsonapi.org/format/#document-top-level) a top-level `error` field.

##### To reproduce

1. Use `serializer.is_valid(raise_exception=True)` in an API view
2. Send invalid JSON from the client to *any* API endpoint

    Example:
    ```js
    ajax({
      url: '/my/api',
      contentType: 'application/vnd.api+json',
      type: 'POST',
      data: 'not JSON'
    });
    ```

    Expected response:
    ```json
    {
      "errors": [
        {
          "status": "400",
          "source": {"pointer": "/data"},
          "detail": "JSON parse error - No JSON object could be decoded"
        }
      ]
    }
    ```

    Actual response:    
    ```json
      [
        {
          "status": "400",
          "source": {"pointer": "/data"},
          "detail": "JSON parse error - No JSON object could be decoded"
        }
      ]
    ```

It seems like the exception handler is [properly setting the resource name](https://github.com/django-json-api/django-rest-framework-json-api/blob/v2.0.0-beta.2/rest_framework_json_api/exceptions.py#L72), which *should* be used [here](https://github.com/django-json-api/django-rest-framework-json-api/blob/v2.0.0-beta.2/rest_framework_json_api/renderers.py#L376-L405). The former gets called, but the latter does not. DRF seems to have taken over, responding with this project's `exception_handler` as-is.

##### My requirements

```
...
Django==1.9.1
djangorestframework==3.3.3
djangorestframework-jsonapi==2.0.0-beta.2
...
```

##### My config

```py
REST_FRAMEWORK = {
    'DEFAULT_PERMISSION_CLASSES': (
        'rest_framework.permissions.IsAuthenticated',
    ),
    'DEFAULT_AUTHENTICATION_CLASSES': (
        'rest_framework.authentication.TokenAuthentication',
        'rest_framework.authentication.BasicAuthentication',
    ),
    'DEFAULT_RENDERER_CLASSES': (
    ),
    'DEFAULT_PARSER_CLASSES': (
        'rest_framework_json_api.parsers.JSONParser',
        'rest_framework.parsers.FormParser',
        'rest_framework.parsers.MultiPartParser',
    ),
    'EXCEPTION_HANDLER': 'rest_framework_json_api.exceptions.exception_handler',
}
```

joeyespo:
This fixes #218 and also addresses #214.

If the optional `JSON_API_UNIFORM_EXCEPTIONS` is set to `True`, *all* exceptions will respond with the JSON API format. This is how it used to work, but now for non-JSON-API views, the top-level `errors` object is properly included in the response, fixing #218.

When `JSON_API_UNIFORM_EXCEPTIONS` is `False` (the default), non-JSON-API views will respond with the normal DRF error format, addressing #218 in a way that doesn't require views to manually override `handle_exception` (which is how it was addressed in #222).

### DIFF
--- a/rest_framework_json_api/exceptions.py
+++ b/rest_framework_json_api/exceptions.py
@@ -1,7 +1,16 @@
+from django.conf import settings
 from django.utils.translation import ugettext_lazy as _
 from rest_framework import status, exceptions
 
 from rest_framework_json_api import utils
+from rest_framework_json_api import renderers
+
+
+def rendered_with_json_api(view):
+    for renderer_class in getattr(view, 'renderer_classes', []):
+        if issubclass(renderer_class, renderers.JSONRenderer):
+            return True
+    return False
 
 
 def exception_handler(exc, context):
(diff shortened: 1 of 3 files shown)


jerel:
@joeyespo thanks for this!

PR #240, opened 2016-05-13
DRAFT: Polymorphism post and patch support

ograycode:
This is based off of @leo-naeka's pull #211, and I do not consider it merge-able in it's current state. That said, I wanted to get early feedback from this project's maintainers on the direction of the pull request as well as know what I can do to help get this functionality in.

**The pull request builds on #211 in the following ways:**

- updates leo-naeka's original work to be compatible with current HEAD of develop.
- adds support and tests for `patch` requests on a polymorphic model's `-detail` endpoint.
- adds support and tests for `post` requests on a polymorphic model's `-list` endpoint.
- fixes `includes` json structure and `-detail` json structure to use fields defined by the polymorphic serializer's child fields property. 

**TODO**:

- [x] clean up git history
- [ ] test `relationship` endpoints.
- [x] move a lot of `ProjectSerializer` into a more generic polymorphic serialize.
- [ ] Update documentation.
- [x] remove pdb `# b` comment.

(I did this work because #211 had not been updated in about a month and my team was running up against this issue).


leo-naeka:
Thanks @ograycode, looks great!

Indeed, my preliminary work was not complete then definitely not merge-able. I'm thrilled that there's so much interest in polymorphic support these days.
I had just been able to free myself a little time to work on it these days, so I'm totally open to code with you on this and contribute to this PR.

example/migrations/0002_auto_20160513_0857.py
```diff
@@ -9,7 +9,6 @@
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('contenttypes', '0002_remove_content_type_name'),
```
leo-naeka: `('contenttypes', '0001_initial'),` is still needed

ograycode: I had removed it because I thought the dependency was breaking the older django versions (<1.7) on test. I think I'm wrong though.

ograycode:
@leo-naeka great to see your response. I'll clean up the git history in my repo this afternoon. Then you should be able to base any additional work you do on it. Once that's done, you can submit pull requests against my repo and I'll merge them into the `polymorphism` branch, or, alternatively, I can give you collaborator status on the repo.

example/migrations/0002_auto_20160513_0857.py
leo-naeka: It breaks on Django 1.7 since migration `0002` does not exist (added in 1.8). But for the purpose of this migration, the initial content type migration seems fine as a dependency.

ograycode:
@leo-naeka I have cleaned up the git history of this branch. I don't plan on doing any more interactive rebases, so it can be worked on you now.

leo-naeka:
Thanks @ograycode!

For the moment, I'll fork your code, then fill you some PR. You may want to give me the collaborator status later, if necessary.

ograycode:
my colleague @tpilara will be taking over for me and will be working towards completing this.

stianpr:
> my colleague @tpilara will be taking over for me and will be working towards completing this.

Any progress on this patch @ograycode and @tpilara? If not then can you outline what is missing and how we can get in a better merge state. We haven't tested it yet on our code base, but we would like to help if possible.

leo-naeka:
Since I'm the main developer on this and I want to move forward (but having no permissions on [ZeroCater/django-rest-framework-json-api](https://github.com/ZeroCater/django-rest-framework-json-api)), I propose to take the lead and continue with this on my original PR #211. Then, this one could be closed.
Are you ok with this @ograycode, @tpilara ?

cc @stianpr

birkholz:
@leo-naeka I work at ZeroCater with @ograycode and @tpilara. We replaced our polymorphic models, so it's best if you take over this.

stianpr:
We recently made a simple field for our code to make it work with django-polymorphic. I'm a big fan of a more explicit approach with this field than doing lots of magic in `ResourceRelatedField`. 

The fundamental reason for this to work is that we need to get the model name based on the object ID with the queryset. That will make the type verification work.

```Python
from rest_framework_json_api.exceptions import Conflict
from rest_framework_json_api.utils import get_resource_type_from_instance


class PolymorphicRelatedField(ResourceRelatedField):
    """Polymorphic Related Field.

    A field to ignore failed type verification in
    `ResourceRelatedField.to_internal_value`. Rest Framework JSON API does not
    support Django-polymorphic yet. Its type verification is based on
    retrieving resource name from `queryset.model` on the field. In our case
    that will always return the base model name, which will fail for sub models.
    """

    def to_internal_value(self, data):
        """Override to ignore `Conflict` exception.

        Since the resource name is retrieved from `queryset.model` on the
        field, we make sure it's ignored.
        """

        try:
            return super().to_internal_value(data)

        except Conflict:
            obj = self.queryset.filter(id=data['id']).first()
            if data['type'] != get_resource_type_from_instance(obj):
                raise

            # Luckily the Conflict exception is raised at the end so we can
            # explicitly call super on `ResourceRelatedField`, which is the
            # `PrimaryKeyRelatedField`.
            return super(ResourceRelatedField, self).to_internal_value(
                data['id'])
```

`to_representation()` will just work out of the box because django-polymorphic returns the correct sub type of the model. What do you think?

leo-naeka:
Ok, thanks. So this PR should be closed @jerel, I'm taking over on #211

PR #243, opened 2016-05-30
Optimize/relationship extraction

bor3ham:
Some options to reduce the number of queries made when using relationships.

Currently the number of db queries made explodes out when either:

- Including just the primary key as a field in a serializer, without actually "including" it
- Actually "including" relationships

without prefetching on the viewset's queryset. For complicated models it can be very expensive to defensively prefetch _everything_ every time, so these changes attempt to only fetch them from the db when needed.


bor3ham:
Rebased and patched over at:
https://github.com/django-json-api/django-rest-framework-json-api/pull/264

PR #245, opened 2016-05-31
Require 'id' is included in PATCH request

tpilara:
According to the JSON API spec regarding the updating of resources:

> The PATCH request MUST include a single resource object as primary data. The resource object MUST contain type and id members.

Currently the package will accept PATCH requests made that omit the `id` for the resource object.

This pull request fixes that and adds a test that verifies a 400 error is correctly thrown if a PATCH request is made without the `id`.


scottfisk:
@tpilara Looks like your test failures are due to a misconfigured tox file. Specifically: https://github.com/django-json-api/django-rest-framework-json-api/blob/develop/tox.ini#L11. 

A django alpha 1.10 build dropped on PyPI and that got downloaded instead of 1.9. I will PR a quick fix for that.

tpilara:
@scottfisk Thanks for the quick fix! I've got one more PR that I'd like to make (which will also fail when it gets run due to the same issue). I'm going to open that now, so after this gets merged in do you think you would be able to re-run the checks on my PRs from TravisCI?

If not I can always close and re-open them, just let me know. Thanks!

jerel:
@tpilara I just merged the tox fix so you can now pull from our `develop` branch and push to your PR branches

tpilara:
@jerel Thanks for merging that PR in so quickly. I've merged that into my branch so tests should pass now.

scottfisk:
this lgtm :+1: @jerel Do we want the merge commit in there or a rebase?

jerel:
@scottfisk I'm :+1: also, go ahead and merge it (doing a squash during merge is fine)

PR #246, opened 2016-05-31
Update tox.ini

scottfisk:
Updated tox file to not pull in a django 1.10 build when testing against django 1.9.

The >= resolved to a Django 1.10 alpha build that must have just been dropped. It broke some tests, specifically the usage of [patterns](https://docs.djangoproject.com/en/1.9/ref/urls/#patterns).


tpilara:
Thanks! I just created an identical PR about a minute ago. 👍

PR #247, opened 2016-05-31
Fix return of stale data on PATCH of to-one relationship

tpilara:
I noticed that PATCH requests made to relationship views with to-one relationships correctly update the underlying relationship with the passed data, but the response I get back is stale (it returns the original relationship instead of reflecting the new one).

This is only a problem with to-one relationships, but not to-many, as that code patch updates the `related_instance_or_manager` directly, while in the case of a to-one relationship it is returning a serialized version of the relationship it fetches in the beginning of the function.

I have fixed this by refreshing the `related_instance_or_manager` at the end of the block that updates to-one relationships. I have additionally added asserts to the methods that test updating to-one and to-many relationships to make sure that the response data is equal to the data passed in through the request.

As an aside, I'm not entirely sure if these updates should return a 200 (as they do now) or a 204. The spec has the following to say about the response codes when updating relationships:

> 204 No Content
A server MUST return a 204 No Content status code if an update is successful and the representation of the resource in the request matches the result.

> 200 OK
If a server accepts an update but also changes the targeted relationship(s) in other ways than those specified by the request, it MUST return a 200 OK response. The response document MUST include a representation of the updated relationship(s).

My reading is that maybe this should be a 204 response as the new relationship is exactly the one that has been passed in through the PATCH request, and nothing else about the relationship has changed. That said, regardless of whether it should be a 204 or 200, it shouldn't be returning stale relationship data which is what this PR aims to fix.


example/tests/test_views.py
```diff
@@ -101,6 +101,7 @@ def test_patch_to_one_relationship(self):
         }
         response = self.client.patch(url, data=json.dumps(request_data), content_type='application/vnd.api+json')
         assert response.status_code == 200, response.content.decode()
+        assert response.data == request_data['data']
```
scottfisk: Does these 2 test lines cause the tests to fail without the fix?

tpilara: @scottfisk The line fails on `test_patch_to_one_relationship` without the fix. It passes before and after on `test_patch_to_many_relationship`, but I added it just to make sure that a future change doesn't cause it to stop working since only checking a status code of 200 doesn't make any guarantees about the integrity of the data in the response.

scottfisk:
:+1: thanks for the fix @tpilara

PR #255, opened 2016-06-21
try to use field.remote_field instead of field.related

### LINKED ISSUE #248: field.related RemovedInDjango110Warning
This is likely an easy fix, I think the only tricky thing is making sure that this remains backwards compatible when making the change to field.remote_field.

This warning is seen using Django 1.9.

`rest_framework_json_api/utils.py:194: RemovedInDjango110Warning: Usage of field.related has been deprecated. Use field.remote_field instead.`

[Link to relevant line](https://github.com/django-json-api/django-rest-framework-json-api/blob/54f1dcd86bd571e83988c228e6ab3e3f72093d27/rest_framework_json_api/utils.py#L194)

alexahn:
Closes #248

### DIFF
--- a/rest_framework_json_api/utils.py
+++ b/rest_framework_json_api/utils.py
@@ -191,7 +191,10 @@ def get_related_resource_type(relation):
                 # Django 1.7
                 relation_model = parent_model_relation.related.model
         elif hasattr(parent_model_relation, 'field'):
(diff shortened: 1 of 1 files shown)


scottfisk:
LGTM. Thanks for the fix.

PR #257, opened 2016-06-30
Convert include queries back to snake case

birkholz:
In [parsers](https://github.com/django-json-api/django-rest-framework-json-api/blob/develop/rest_framework_json_api/parsers.py#L32), we convert incoming keys to snake case (undoing whatever formatting was done during [rendering](https://github.com/django-json-api/django-rest-framework-json-api/blob/develop/rest_framework_json_api/renderers.py#L66)). However, include parameters have to exactly match the fields of models/serializers.

This PR brings the same reversal behavior to include parameters as its done for attributes, etc.

Example:
`?include[firstName]=` == `?include[first_name]=`

### DIFF
--- a/rest_framework_json_api/renderers.py
+++ b/rest_framework_json_api/renderers.py
@@ -4,6 +4,7 @@ Renderers
 import copy
 from collections import OrderedDict
 
+import inflection
 from django.utils import six, encoding
 from rest_framework import relations
 from rest_framework import renderers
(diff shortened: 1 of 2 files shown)


jerel:
@scottfisk what are your thoughts on this? This one would be easy to merge but it looks like we also have related work at #204

birkholz:
@jerel I wrote this up before I saw #204 

It seems that #204 is taking things further by also reversing pluralization and adding a setting to specify which format the keys should be reverted to. In my opinion, the effort of that PR should be expanded to the other areas that perform this behavior, namely [parsing](https://github.com/django-json-api/django-rest-framework-json-api/blob/develop/rest_framework_json_api/parsers.py#L32). This could then be considered a stopgap until that larger development is ready.

scottfisk:
@jerel This looks good to me.  Agreed @birkholz I think we can pull this and #204 can continue the more in depth work based on this.

PR #260, opened 2016-07-19
SerializerMethodResourceRelatedField with many=True does not show up in included

tbartelmess:
When using a `SerializerMethodResourceRelatedField` in a serializer, with `many=True` being set, and the request is asking for to include it (via the `?include`) parameter, the field is not included in the result.

This PR adds a test and a fairly hackish fix (by checking is a 'child_relation' is present in the `_kwargs` parameter, when the field is evaluated for includes).

### DIFF
--- a/example/serializers.py
+++ b/example/serializers.py
@@ -35,6 +35,7 @@ class EntrySerializer(serializers.ModelSerializer):
         'authors': 'example.serializers.AuthorSerializer',
         'comments': 'example.serializers.CommentSerializer',
         'featured': 'example.serializers.EntrySerializer',
+        'suggested': 'example.serializers.EntrySerializer',
     }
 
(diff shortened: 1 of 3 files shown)


tbartelmess:
It looks like only one building Travis failed. It looks like the setup failed 
`The command "curl -s -o python-3.5.tar.bz2 ${archive_url}" failed and exited with 18 during .`
not an actual test

khornberg:
Would it be helpful if this was behind a setting so that those with accustomed to the current behavior are not affected? Or is this a bug?  I am preferable to the former.

jerel:
I would view this as a bug: when a field is specified in `include` it should be included.

khornberg:
Yeah, after playing with it, I came to the same conclusion.

PR #261, opened 2016-07-21
Include metadata in JSONParser output

hamedahmadi:
The JSONParser will add metadata in the request under a top level key `_meta`.

This provides access to the data in the `meta` tag of the request.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -47,3 +47,14 @@ Calls a `get_root_meta` function on a serializer, if it exists.
 `build_json_resource_obj(fields, resource, resource_instance, resource_name)`
 
 Builds the resource object (type, id, attributes) and extracts relationships.
+
+## rest_framework_json_api.parsers.JSONParser
+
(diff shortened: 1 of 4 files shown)


jerel:
@hamedahmadi this looks like a nice addition. What is the reason for choosing `_meta` as the key name rather than `meta`?

hamedahmadi:
@jerel using `_meta` to prevent a clash in case the resource has a `meta` attribute.

jerel:
@hamedahmadi ok, makes sense. Would you mind adding some simple documentation yet before I merge? Make a note about parsing incoming meta here: [docs/usage.md#meta](/django-json-api/django-rest-framework-json-api/blob/develop/docs/usage.md#meta) and add a section for `JSONParser` here with `parse_metadata` documented: [docs/api.md](/django-json-api/django-rest-framework-json-api/blob/develop/docs/api.md)

hamedahmadi:
@jerel Thanks for the comment, updated.

Issue #85, opened 2015-09-15
Feedback requested: what version of Django REST Framework do you use?


jerel:
How many of you are using DRF 3.0.x or 2.4.x? Since the JSON API spec just recently stabilized on 1.0 I'm guessing that most JSON API compliant APIs will be built using one of the latest versions of DRF.

If that is the case I would like to just support DRF 3.1 and newer. Django (1.6+) and Python (2.7+) support should remain unchanged.

jsenecal:
DRF 3.2.2
Django 1.8.3
Python 3.4.3

scottfisk:
DRF 3.2.2
Django 1.7.10 (moving to 1.8 very soon)
Python 2.7.10

mkohram:
DRF 3.1.3
Django 1.8.3
Python 3.4.3

holandes22:
DRF 3.2.3
Django 1.8.4
Python 2.7.10

martinmaillard:
DRF 3.2.3
Django 1.8.4
Python 2.7.10

orf:
DRF 3.2.3
Django 1.8.4
Python 2.7.x and 3.4.x

jerel:
So far it's looking like everyone is using DRF 3.1+ so we'll not worry about supporting <= DRF 3.0. Thanks for the feedback.

PR #268, opened 2016-08-16
Implemented py.test runner in setup.py

georgemarshall:
Other highlights:
* Switched to travis build matrix from tox
* Removed runtest.py blob
* Updated tox.ini to reflect documented version compatibility

Suggestions:
* Remove Python 3.2 and Django 1.7 since both are EOL
* Add [coverage reporting](https://codecov.io)


jerel:
@scottfisk what do you think of these changes? I'm happy with merging them if you are.

tox.ini
```diff
@@ -1,8 +1,8 @@
 [tox]
 envlist =
-    py{27,33,34}-django17-drf{31,32},
```
scottfisk: @georgemarshall It looks like python 3.2 wasn't in the environments before here or [travis.yml](https://github.com/django-json-api/django-rest-framework-json-api/pull/268/files#diff-354f30a63fb0907d4ad57269548329e3L7) Since one of your suggestions was to remove it, we can probably just leave it out?

We should probably remove any other references to python 3.2 while we are at it.

scottfisk:
@georgemarshall I do like the coverage reporting as well :+1: 

@jerel Dropped one comment re python 3.2

tox.ini
georgemarshall: Ha, you're right... Must of stayed up later than I should have.

georgemarshall:
Any feedback on Django 1.7 EOL?

scottfisk:
@georgemarshall I was the only one using it last time we did a [poll](https://github.com/django-json-api/django-rest-framework-json-api/issues/85.).  We am now using 1.9, so I am ok with dropping it.

PR #269, opened 2016-08-16
Fixes unhashable type OrderedDict error exposed by DRF 3.4.4

jerel:


scottfisk:
@jerel I am assuming this change was causing the issue? https://github.com/tomchristie/django-rest-framework/compare/3.4.3...3.4.4#diff-9bc5ef9dafcb145b81a6d228c8200fc6L159

jerel:
@scottfisk yeah, due to that naming change `get_choices` was being called on the base class rather than our version (`choices`). And we are serializing `{'type': 'foo', 'id': '1'}` to json since it's not a simple `pk` so just returning `to_representation` caused an `Unhashable type: OrderedDict` to be thrown.

So, my fix is to remove `@property choices` and let the base class cache that (all it does is call `get_choices`)

scottfisk:
lgtm :+1:

PR #272, opened 2016-09-08
Add isort & flake8 settings.

jarekwg:
Isort & flake8 rules for contributors using these tools in their editor.
Also pinned pytest version, to get tests to pass. (pytest-factoryboy doesn't yet support pytest3+).


scottfisk:
@jarekwg This looks good to me. Thanks!

Issue #270, opened 2016-08-26
UnboundLocalError in renderers.py


daveslab:
I recently upgraded my project to Django 1.10 and to DRF JSON API 2.1.0, and I am now seeing an error every time I try to make a simple GET call on one of my resources. The error remains if I use Django 1.10 or Django 1.9. Here is a piece of the error page:
```
UnboundLocalError at /api/abonnes_fichier/
local variable 'relation_instance' referenced before assignment
Request Method:	GET
Request URL:	http://localhost:9000/api/abonnes_fichier/
Django Version:	1.9.9
Exception Type:	UnboundLocalError
Exception Value:	
local variable 'relation_instance' referenced before assignment
Exception Location:	.../site-packages/rest_framework_json_api/renderers.py in extract_relationships, line 202
```
The error comes from the function `extract_relationships` in the class `JSONRenderer`. I'm still a newbie to Django and DRF so I'm having a hard time debugging the function. I think the problematic commit might be b96d79d49995c0faac6174ecdd1715a2b1121d3d.

jerel:
@daveslab thanks for the detailed report. I believe that `and relation_instance is not None` on line 202 is not necessary as `resource_instance` (which must be what the conditional was meaning to target) is already checked on line 78. Could you remove it on your local copy and see if that resolves your issue?

jerel:
@daveslab I'm surprised that a test didn't catch this regression. If you want to add a test to verify the fix [this would be a good place for it.](https://github.com/django-json-api/django-rest-framework-json-api/tree/develop/example/tests/integration)

daveslab:
@jerel Thanks for responding so quickly. As you asked, if I change line 202 to be
```python
if isinstance(field, ListSerializer):
```
Then it works just fine. How do I add a test? (To be honest, like I said, I'm a serious n00b and don't even know how to use git).

daveslab:
@jerel Ahh, however, I just noticed another problem (not sure if I should open another issue), but now that the call works when I change line 202, it **doesn't** return the right `type`. It should return (as it does using django-rest-framework-json-api 2.0.1):
```js
//...
 "relationships": {
        "frequences_recu_fiscal": {
          "data": [
            {
              "type": "FrequenceRecuFiscalAbonneFichier",
              "id": "18470"
            },
           //...
          ]
        },
//...
```
But what it returns is:
```js
//...
 "relationships": {
        "frequences_recu_fiscal": {
          "data": [
            {
[...]

PR #278, opened 2016-09-13
Remove check for relation_instance as it was not available in scope

scottfisk:
Closes issue #270

I added a renderer unit test to catch this error. 

Unfortunately the bug only occurred when there was a single relation on the serializer so I had to make some simplified serializers to reproduce the issue.

### DIFF
--- a/example/tests/unit/test_renderers.py
+++ b/example/tests/unit/test_renderers.py
@@ -1,0 +1,45 @@
+from example.models import Entry, Comment
+from rest_framework_json_api import serializers, views
+from rest_framework_json_api.renderers import JSONRenderer
+
+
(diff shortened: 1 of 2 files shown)


scottfisk:
looks like the travis builds are broken. Not from this PR specifically.

jerel:
Interesting. Looks like a new release in our upstream dependencies must be to blame

jarekwg:
It's an issue with factoryboy not supporting pytest3. I fixed them in my isort/flake8 PR #272. :)

scottfisk:
@jarekwg looks like pulling in and rebasing off #272 did the trick, thanks!

PR #279, opened 2016-09-17
Feature/drf serializer support

### LINKED ISSUE #273: Support for default DRF Serializers (fixing wrong model type)
Split off https://github.com/django-json-api/django-rest-framework-json-api/issues/270

The most recent version broke support for default DRF serializers for reverse and m2m relations (the wrong model type gets returned).

Although the recommended way is to use serializers supplied by jsonapi, we should try to maintain support for drf serializers so long as it's not too big of a job to do so.

- [x] Add failing tests (using `from rest_framework import serializers` vs. `from restframework_json_api import serializers`
- [x] Fix breaking change (probably a bug in `utils.get_related_resource_type`)

jarekwg:
Fixes #273.

I didn't end up explicitly importing drf serializers and writing tests for them. The original plan was to just have an extra set of tests that swap out the jsonapi `serializers` module with drf's original one, however the example app makes use of several serializer features available only to jsonapi, which means i'd've had to write a separate set of plain drf serializers, with their own tests.

For fear of bloating the repo too much (and wanting to save myself some time), I instead just added a `test_utils.py` for testing jsonapi's `utils.py`. This identified the bug, which I fixed.

Please note, this just _happens_ to fix the reverse/m2m functionality for default drf serializers that got broken in 0aedffb, but does not guarantee them to work in general or if other things change.

I feel this is probs the least messy approach, as it might become completely infeasible to support the default drf serializers if jsonapi evolves enough in the future.

~~This is built on #272, please merge that one first.~~
Rebased.

### DIFF
--- a/example/tests/test_utils.py
+++ b/example/tests/test_utils.py
@@ -1,0 +1,22 @@
+"""
+Test rest_framework_json_api's utils functions.
+"""
+from rest_framework_json_api import utils
+
+from ..serializers import EntrySerializer
+from ..tests import TestBase
+
+
+class GetRelatedResourceTests(TestBase):
+    """
+    Ensure the `get_related_resource_type` function returns correct types.
+    """
+
+    def test_reverse_relation(self):
+        """
+        Ensure reverse foreign keys have their types identified correctly.
+        """
+        serializer = EntrySerializer()
+        field = serializer.fields['comments']
+
+        self.assertEqual(utils.get_related_resource_type(field), 'comments')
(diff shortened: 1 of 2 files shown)


jarekwg:
Hrm, hang on. Don't merge yet. I'm going to make the code a bit more explicit. These try-catches don't make things obvious.

jarekwg:
K, done.
As pythonic as `try`/`catch` is, it's not very explicit in this case; making the code look like we're just stumbling around looking for an attribute that exists. By explicitly checking types and django versions, we:

- make the logic readable.
- avoid accidentally referring to an attribute in cases we shouldn't be.
- make it easy to remove conditional branches once older django versions are dropped.

PR #282, opened 2016-09-19
Parsing of included data in JSON request

peterkajan:
Included resources are used to optimize the number of requests that has to be made to the server when fetching an entity. See: http://jsonapi.org/format/#document-compound-documents

Similar concept can be used also when creating entity. If entity contains relation to another entity,  its data can be directly included in the request.

This pull request extends parser so that "included" key is not ignored in JSON request. Instead, it parses the data and adds them under "_included" key in the parsing result.

### DIFF
--- a/example/tests/test_parsers.py
+++ b/example/tests/test_parsers.py
@@ -7,17 +7,19 @@ from rest_framework_json_api.parsers import JSONParser
 
 class TestJSONParser(TestCase):
 
-    def setUp(self):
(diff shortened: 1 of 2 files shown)


jerel:
I don't believe this is something we can accept as it deviates too far from the spec. `includes` is currently defined as a `GET` only feature. For creating or updating relationships refer to the spec here http://jsonapi.org/format/#crud

PR #286, opened 2016-10-07
Add filter feature per JSON API specs.

abdulhaq-e:
Added filter feature. This uses DjangoFilterBackend which either
comes from `rest-framework-filter` or `django-filter`. The former
is a package which adds more features to the latter.

All that is done is reformatting the request query parameters to be
compatible with DjangoFilterBackend. The spec recommend the pattern
`filter[field]=values`. JsonApiFilterBackend takes that pattern and
reformats it to become `field=value` which will then work as typical
filtering in DRF.

The specs only rule towards filtering is that the `filter` keyword
should be reserved. It doesn't however say about the exact format.
It only **recommends** `filter[field]=value` but this can also be
`filter{field}=value` or `filter(field)=value`. Therefore, a new
setting for DJA is introduced: JSON_API_FILTER_KEYWORD. This is a
regex which controls the filter format. Its default value is:

```
JSON_API_FILTER_KEYWORD = 'filter\[(?P<field>\w+)\]'
```

It can be changed to anything as long as the `field` keyword is
included in the regex.

The docs have been updated.

Unfortunately, the editor's beautifier was run and made lots of
other changes to the code style. No harm was done though, that's
why TDD exists :D

### DIFF
--- a/example/filters.py
+++ b/example/filters.py
@@ -1,0 +1,12 @@
+import django_filters
+
+from example.models import Comment
+
+
+class CommentFilter(django_filters.FilterSet):
+
+    class Meta:
+        model = Comment
+        fileds = {'body': ['exact', 'in', 'icontains', 'contains'],
+                  'author': ['exact', 'gte', 'lte'],
+                  }
--- a/rest_framework_json_api/filters.py
+++ b/rest_framework_json_api/filters.py
@@ -1,0 +1,19 @@
+try:
+    import rest_framework_filters
+    DjangoFilterBackend = rest_framework_filters.backends.DjangoFilterBackend
+except ImportError:
+    from rest_framework import filters
+    DjangoFilterBackend = filters.DjangoFilterBackend
+
+from rest_framework_json_api.utils import format_query_params
+
+class JsonApiFilterBackend(DjangoFilterBackend):
+
+    def filter_queryset(self, request, queryset, view):
+
+        filter_class = self.get_filter_class(view, queryset)
+        new_query_params = format_query_params(request.query_params)
+        if filter_class:
+            return filter_class(new_query_params, queryset=queryset).qs
+
+        return queryset
--- a/rest_framework_json_api/utils.py
+++ b/rest_framework_json_api/utils.py
@@ -2,6 +2,7 @@
 Utils.
 """
 import copy
+import re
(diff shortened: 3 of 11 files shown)


rest_framework_json_api/filters.py
```diff
@@ -0,0 +1,20 @@
...
+    DjangoFilterBackend = filters.DjangoFilterBackend
+
+from rest_framework_json_api.utils import format_query_params
+
+class JsonApiFilterBackend(DjangoFilterBackend):
+
+    def filter_queryset(self, request, queryset, view):
+
+        filter_class = self.get_filter_class(view, queryset)
+        print(request.query_params)
```
khornberg: remove print

rest_framework_json_api/utils.py
```diff
@@ -72,7 +74,8 @@ def get_resource_name(context):
                 # The resource name is not a string - return as is
                 return resource_name
 
-            # the name was calculated automatically from the view > pluralize and format
+            # the name was calculated automatically from the view > pluralize
```
khornberg: Changing all these things is not helpful, even if your dev software did it for you. It only adds noise to the PR.

example/filters.py
```diff
@@ -0,0 +1,12 @@
+import django_filters
+
+from example.models import Comment
+
+
+class CommentFilter(django_filters.FilterSet):
```
khornberg: There does not seem to be an example of using DRF filters

rest_framework_json_api/filters.py
abdulhaq-e: Done.

example/filters.py
abdulhaq-e: Can you provide a link that shows an example?

DRF filtering depends on `django_filter` and similar examples are found in DRF docs.

abdulhaq-e:
> Does this allow filtering on multiple fields at the same time?

Yes, it is permitted to send a request such as:

 `GET http://localhost/?filter[author]=1&filter[blog]=10`

> Does this allow filtering on multiple values of the same field?
> Does this allow filtering of array, hstore and other complex/DB specific fields?

The JSON API specs does not get into any of this, it depends on the user implementation. However, this is all possible using `django_filter`.

PR #300, opened 2016-11-21
Fix #299 (change JSONRenderer staticmethods to classmethods)

### LINKED ISSUE #299: Supposedly overridable methods are actually static
According to the [API section](http://django-rest-framework-json-api.readthedocs.io/en/stable/api.html#rest-framework-json-api-renderers-jsonrenderer) of the documentation, "The `JSONRenderer` exposes a number of methods that you may override if you need highly custom rendering control." However, many of these methods are actually static and there are hardcoded calls to them in the code, making them unoverridable. For example:

```python
@staticmethod
def extract_relationships(fields, resource, resource_instance):
    # Avoid circular deps
    # ...

@staticmethod
    # ...
    relationships = JSONRenderer.extract_relationships(fields, resource, resource_instance)
```

Shouldn't these be classmethods instead?

arielpontes:
Remove all hardcoded calls to JSONRenderer methods and change all staticmethods to classmethods in rederers.py so they can be safely overriden.

### DIFF
--- a/example/tests/unit/test_renderer_class_methods.py
+++ b/example/tests/unit/test_renderer_class_methods.py
@@ -37,6 +37,45 @@ def test_build_json_resource_obj():
     assert JSONRenderer.build_json_resource_obj(
         serializer.fields, resource, resource_instance, 'user') == output
 
+def test_can_override_methods():
+    """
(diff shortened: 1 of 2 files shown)


chrisclark:
Looks great, thanks.

Issue #303, opened 2016-11-30
ResourceRelatedField: resource_name doesn't work for nested serializers


czosel:
I found another issue related to support for `resource_name`. While it works fine for `to_representation` when my resource **has many** related resources, i.e.
```
"relationships": {
  "portal-images": {
    "meta": {
      "count": 2
    },
    "data": [
      {
        "type": "ownerportal-image",
        (...)
```
(here, i could return `ownerportal-image` instead of `portal-image` successfully)
it doesn't work when the relationship is the other way around, i.e. my resource is related to only one other resource:
```
"relationships": {
  "image": {
    "data": {
      "type": "image",
```
(here, i would have expected `ownerimage` instead of `image`)

After debugging a bit, i found that the problem is related to (`relations.py`, 140)
```
resource_type = None
root = getattr(self.parent, 'parent', self.parent)
field_name = self.field_name if self.field_name else self.parent.field_name
if getattr(root, 'included_serializers', None) is not None:
```

In the successful case, `self.parent` refers to a `ManyRelatedField`, and `self.parent.parent` to my serializer.
In the unsuccessful case, `self.parent` refers to my serializer, while `self.parent.parent` is `Null`.

For me, changing the initialization of `root` to
```
root = getattr(self.parent, 'parent', self.parent) or self.parent
```
fixes the problem.

Since i'm a bit unsure if my fix is the right way to go, i thought i'd post this as an issue first to get your feedback. Thanks in advance!

mblayman:
Fixed by #306.

PR #304, opened 2016-12-01
Fix resource_name support for ResourceRelatedField

czosel:
See https://github.com/django-json-api/django-rest-framework-json-api/issues/303

I thought i'd open a PR for this already, because it makes it easier to discuss changes like this.


czosel:
Superseded by #306.

PR #307, opened 2016-12-05
Fix included resources not being included on list

rpatterson:


mblayman:
@jerel hi! It looks like you are the current maintainer for DRF JA. Do you need some help with maintenance? The issue associated with this PR is killing my app's performance. I'd love to get it fixed and released, but I understand that things happen and people get swamped.

If I could lend a bit of a hand to help close some of this stuff out, I'd be happy to do that.

jerel:
@mblayman I am; I would definitely appreciate help with maintenance.

mblayman:
@rpatterson It seems like you've pushed some commits that are beyond the original scope of this PR. Is this stuff related? If so, can you explain? I think it would be difficult to review if this is a bunch of things all in one branch.

rpatterson:
@mblayman Sorry, cleaned up my branches such that `develop` in my fork now only contains changes related to this PR, namely missing included resources.

rest_framework_json_api/utils.py
```diff
@@ -186,6 +186,14 @@ def get_related_resource_type(relation):
         relation_model = relation.model
     elif hasattr(relation, 'get_queryset') and relation.get_queryset() is not None:
         relation_model = relation.get_queryset().model
+    elif (
+            getattr(relation, 'many', False) and
+            hasattr(relation.child, 'Meta') and
```
mblayman: I'm wondering if these `hasattr` checks are a bit too much. A serializer doesn't have to be a `ModelSerializer` so there is no requirement for `Meta` to exist, but isn't `model` a required attribute of `Meta` for a `ModelSerializer`? It feels slightly too defensive to me to check that `model` exists.

rest_framework_json_api/utils.py
```diff
@@ -186,6 +186,14 @@ def get_related_resource_type(relation):
...
+            hasattr(relation.child.Meta, 'model')):
+        # For ManyToMany relationships, get the model from the child
+        # serializer of the list serializer
+        # TODO Test coverage
```
mblayman: Could you either add a test to cover this line or delete the `TODO` comment? A test would be preferable, but understand if you can't. I don't think that `TODO` comments should be allowed in a master branch. Experience has showed me that they almost always languish and are forgotten until someone comes by later and deletes the comment.

rest_framework_json_api/utils.py
```diff
@@ -266,8 +274,11 @@ def get_included_resources(request, serializer=None):
 
 
 def get_default_included_resources_from_serializer(serializer):
+    meta = getattr(serializer, 'JSONAPIMeta', None)
+    if meta is None and getattr(serializer, 'many', False):
+        meta = getattr(serializer.child, 'JSONAPIMeta', None)
     try:
-        return list(serializer.JSONAPIMeta.included_resources)
+        return list(meta.included_resources)
```
mblayman: Slightly contrary to what I stated up above, what do you think of changing this logic to have a `hasattr` check here? I'd guess that many (most?) serializers will not have `included_resources`. Relying on an `AttributeError` for normal flow control of the majority case is bound to be slow. I recognize this would slightly increase the scope of this branch. Would you be ok fixing that while you're in here?

mblayman:
I'm excited to get this one fixed for my selfish interest of making my app not hammer my API with requests. I had a couple of minor comments that I think would be worth addressing.

I'm happy to merge once they've been considered. 👍 Thanks, @rpatterson!

rest_framework_json_api/utils.py
rpatterson: I don't quite understand.  If the serializer may not be a `ModelSerializer`, then shouldn't we *not* assume that `Meta` has a `model`?

rest_framework_json_api/utils.py
rpatterson: Removed the TODO comment

rest_framework_json_api/utils.py
mblayman: Ah, you're right. I thought a vanilla serializer made no use of `Meta`, but I can see from the source that that is not correct. By all means, please keep `model`. 😊

rest_framework_json_api/utils.py
rpatterson: Done

rpatterson:
@mblayman I think I addressed all your feedback.  Let me know if I missed anything.

rest_framework_json_api/utils.py
```diff
@@ -266,8 +273,11 @@ def get_included_resources(request, serializer=None):
 
 
 def get_default_included_resources_from_serializer(serializer):
+    meta = getattr(serializer, 'JSONAPIMeta', None)
+    if meta is None and getattr(serializer, 'many', False):
+        meta = getattr(serializer.child, 'JSONAPIMeta', None)
     try:
-        return list(serializer.JSONAPIMeta.included_resources)
+        return list(getattr(meta, 'included_resources', []))
```
mblayman: I think the surrounding `try/except` is no longer needed, right?

rpatterson: Doh!  Done.

mblayman:
Dude, you rock! I'll merge it once CI goes green. Thanks again!

PR #308, opened 2016-12-08
Document DJA's ability to use Compound Documents.

mblayman:
This branch adds documentation for how to include extra resources
under the `included` key.

It also adds Sphinx and the extra required dependencies to build
the docs to `requirements-development.txt`.

I wrote this from the perspective of a user who recently encountered the need to side load extra data and had to scour the source to find it. If it's lacking in additional features that are hidden, I'd love to know and would be happy to add more.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -457,8 +457,53 @@ Adding `url` to `fields` on a serializer will add a `self` link to the `links` k
 
 Related links will be created automatically when using the Relationship View.
 
+### Included
+
+JSON API can include additional resources in a single network request.
+The specification refers to this feature as
+[Compound Documents](http://jsonapi.org/format/#document-compound-documents).
+Compound Documents can reduce the number of network requests
+which can lead to a better performing web application.
+To accomplish this,
+the specification permits a top level `included` key.
+The list of content within this key are the extra resources
+that are related to the primary resource.
+
+To make a Compound Document,
+you need to modify your `ModelSerializer`.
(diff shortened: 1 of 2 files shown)


mblayman:
I noticed this Travis build is failing *hard*. I looked at the errors and they all seem to be network related (which is a relief because I'd be concerned if changing some Markdown broke the build! 😄 ). I'm guessing Travis is having some kind of outage. I would attempt to rebuild, but I don't have permission to do that.

mblayman:
@jerel I feel a bit weird merging my own PR. Could you take a look at this? I'm trying to clear the PR queue to minimize work in progress so DRF JA doesn't feel like a log jam. Thanks!

jerel:
Looks good, thanks so much!

n1ncha:
Wow this was timely, I just came across the need for this while working on the backend for an ember app. I'm still a little confused on what the need is for both the "included_resources" and "included_serializers". I keep re-reading the source and the the documentation posted above, maybe I will realize what they are for once I work with it for a bit... but any clarification would be most appreciated, thanks.

jerel:
@n1ncha `included_serializers` sets up the serializers while `included_resources` specifies what is included by default (without having to be specified in the query string)

n1ncha:
I had been doing ?included=<name> in the querystring (which I discovered in the docs) so I wasn't seeing the difference when included_resources wasn't there. Thanks!

PR #310, opened 2016-12-20
Refactor tests.

kramarz:
Pytest gives cleaner output if you compare list/dictionaries instead of dumped jsons. It is also easier to read tests.


mblayman:
This seems reasonable to me. Comparing native data structures feels like a good fit compared to giant string comparison. 👍

PR #312, opened 2016-12-22
Use 'page_query_param' for pagination instead of hard coded page

RegionSyx:
### What
- Use self.page_query_param to determine the query parameter to use for
  pagination
- Set the default page_query_param to 'page'
- Freeze fake-factory version to 0.7.4

### Why
- Allow the use of different page query parameters to do pagination
- The latest version of fake-factory is now depreciated:
  https://pypi.python.org/pypi/fake-factory/9999.9.9


mblayman:
@RegionSyx Thanks for the contribution! Could you make a few changes?
1. Replace the pinned `fake-factory` dependency with its replacement `Faker` (sans pinning).
2. Add documentation to the Usage area where pagination is mentioned.
3. Remove the default as that same value is set in the base class coming from DRF.

I'll be happy to merge once those changes are made. 👍

mblayman:
@RegionSyx This branch is now in conflict with the main branch. Could you update the branch to resolve the conflicts and address the other review comments? Once that is done, I'll be able to merge this branch. Thanks.

mblayman:
Lookin' good! Thanks @RegionSyx. 👍

PR #317, opened 2017-01-25
Fix get_related_resource_type

bor3ham:
To consider child_relation, eg. the default reverse relationship created by m2m. Otherwise it appears to fall back to the parent serializer type.


bor3ham:
Don't think this is actually required, issue seems to be fixed by using provided json_api model serializers.

Darkheir:
I still have the problem on reverse Many to Many relations using a Through table and I think this commit may fix the issue ...

Darkheir:
Actually it doesn't fix the issue if in the serializer the field is set as read only.

PR #319, opened 2017-02-01
Add support for GenericRelations

### LINKED ISSUE #224: GenericRelation Support
Is GenericRelation support possible?  This is how I would do it with Django rest.  Can't get this to work however as I am getting the `is not JSON serializable` error.  Guessing there might be some serialization issue with json api?

```
# Serializer
class FooSerializer(serializers.ModelSerializer):

    included_serializers = {
        'news': NewsSerializer,
    }

    class Meta:
        model = LSEquity
        fields = ('id', 'news')


# News Serializer
class NewsSerializer(serializers.ModelSerializer):

    class Meta:
        model = News
        fields = ('id', 'description', 'date')


# Foo Model
class FooModel(models.Model):
    news = GenericRelation(News)


# News Model
class News(models.Model):
    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
    description = models.TextField(max_length=200, blank=True)
    date = models.DateTimeField(blank=True, null=True)
    content_type = models.ForeignKey(ContentType)
    object_id = models.UUIDField()
    content_object = GenericForeignKey('content_type', 'object_id')

    class JSONAPIMeta:
        resource_name = 'news'
```

I'd be happy to add some "documentation" through the examples and work through it that way.  That is of course if somebody has not already come across this!

santiavenda2:
Fixed #224 

Also, we fixed

`
ConftestImportFailure: (local('/home/santiago/develop/invgate-django-rest-framework-json-api/example/tests/conftest.py'), (<type 'exceptions.ImportError'>, ImportError('No module named faker',), <traceback object at 0x7f64f3d8c5a8>))
`

[Fake-factory](https://pypi.python.org/pypi/fake-factory) has been deprecated, changing its name to [Faker](https://pypi.python.org/pypi/Faker)


example/factories/__init__.py
```diff
@@ -58,3 +58,11 @@ class Meta:
     body = factory.LazyAttribute(lambda x: faker.text())
     author = factory.SubFactory(AuthorFactory)
 
+
+class EntryTaggedItemFactory(factory.django.DjangoModelFactory):
```
mblayman: Could you call this `TaggedItemFactory` instead of `EntryTaggedItemFactory`? That seems more in line with the other factory naming conventions here.

example/models.py
```diff
@@ -3,6 +3,9 @@
 
 from django.db import models
 from django.utils.encoding import python_2_unicode_compatible
+from django.contrib.contenttypes.models import ContentType
```
mblayman: After reviewing this package's code, imports are usually alphabetically sorted. Could you move these `contrib` imports about the `db` import?

rest_framework_json_api/utils.py
```diff
@@ -30,11 +30,14 @@
 if django.VERSION >= (1, 9):
     from django.db.models.fields.related_descriptors import ManyToManyDescriptor, ReverseManyToOneDescriptor
     ReverseManyRelatedObjectsDescriptor = type(None)
+    from django.contrib.contenttypes.fields import ReverseGenericManyToOneDescriptor
```
mblayman: This method of having two mirrored imports that set one to the None type is something I found confusing. After digging through the Django contenttypes source, I see that what is happening is trying to deal with a rename of the descriptor from 1.8 to 1.9. Unfortunately, this method leads to extra branches in an already large if/elif section.

How about using an import alias instead?

```python
if django.VERSION >= (1, 9):
    # ... snip unmodified lines ...
    from django.contrib.contenttypes.fields import ReverseGenericManyToOneDescriptor
else:
    from django.contrib.contenttypes.fields import ReverseGenericRelatedObjectsDescriptor as ReverseGenericManyToOneDescriptor
```
I believe this would clean up the code, be more efficient, and capture the fact that the descriptor was renamed. It seems this pattern was already followed for the first two imports that follow the `else`.

requirements-development.txt
```diff
@@ -2,6 +2,6 @@
 pytest>=2.9.0,<3.0
 pytest-django
 pytest-factoryboy
-fake-factory
+Faker
```
mblayman: Thanks for picking up this package name change. This change has bit me on a few projects already.

mblayman:
Thanks @santiavenda2 for the contribution! Making this work for `GenericRelations` is a feature that I'm sure many users will value considering the popularity of the `contenttypes` system.

This feels like a feature worth documenting. Would you be able to do some of that? At the very least, I think starting a v2.X entry with a bullet point about `GenericRelations` should happen.

I'd be happy to merge after some minor modifications occur. 👍

rest_framework_json_api/utils.py
santiavenda2: Done!

mblayman:
Nice work. Thanks, @santiavenda2!

Issue #321, opened 2017-02-05
Bug on render() method concerning renderer_context


erichonkanen:
Hello, we are using this package and ran into an issue with our unit tests where it seems that the rest_framework APIRequestFactory which is behind the test client does not pass in a `renderer_context`:

```
    def render(self, data, accepted_media_type=None, renderer_context=None):

>       view = renderer_context.get("view", None)
E       AttributeError: 'NoneType' object has no attribute 'get'

/usr/local/lib/python3.5/site-packages/rest_framework_json_api/renderers.py:419: AttributeError
```
I think you could fix it by doing `renderer_context={}`

The code from this library will break at the following lines which are not guarded properly:

https://github.com/django-json-api/django-rest-framework-json-api/blob/develop/rest_framework_json_api/renderers.py#L419

also:
https://github.com/django-json-api/django-rest-framework-json-api/blob/develop/rest_framework_json_api/renderers.py#L401
https://github.com/django-json-api/django-rest-framework-json-api/blob/develop/rest_framework_json_api/renderers.py#L420

Here is the corresponding rest_framework.test line that you can see does not include `renderer_context` as an argument:

https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/test.py#L183

Let me know if this makes sense or if I'm off course!

Thanks

sliverc:
This should have been fixed with #396  hence closing this issue. If there are still issues concerning this error, please comment.

Issue #320, opened 2017-02-05
Add support for cursor based pagination


hoIIer:
After lots of head scratching we just determined that our test was failing because it has cursor pagination enabled, does this package support it or what would be needed to make it work?

```
class FooViewSet(viewsets.ModelViewSet):
    serializer_class = serializers.FooSerializer
    pagination_class = FooCursorPagination

    def get_queryset(self):
        qs = models.Foo.objects.all()
        return qs
```

sliverc:
Marking as help wanted if someone wants to pick up the PR. As a side node `renderer_context` won't fail anymore when it is `None` - this has been fixed with #396

Issue #275, opened 2016-09-08
AttributeError: 'NoneType' object has no attribute 'get'


ahmed-taj:
Hi, there

Since yesterday, I'm stuck on this error in my test script:

    Creating test database for alias 'default'...
    E
    ======================================================================
    ERROR: test_create_account (api.accounts.tests.test_crud.AccountTests)
    ----------------------------------------------------------------------
    Traceback (most recent call last):
      File "/app/api/accounts/tests/test_crud.py", line 22, in test_create_account
        response = self.client.post(url, data)
      File "/usr/local/lib/python3.5/site-packages/rest_framework/test.py", line 169, in post
        path, data=data, format=format, content_type=content_type, **extra)
      File "/usr/local/lib/python3.5/site-packages/rest_framework/test.py", line 90, in post
        data, content_type = self._encode_data(data, format, content_type)
      File "/usr/local/lib/python3.5/site-packages/rest_framework/test.py", line 65, in _encode_data
        ret = renderer.render(data)
      File "/usr/local/lib/python3.5/site-packages/rest_framework_json_api/renderers.py", line 419, in render
        view = renderer_context.get("view", None)
    AttributeError: 'NoneType' object has no attribute 'get'

    ----------------------------------------------------------------------
    Ran 1 test in 0.019s

    FAILED (errors=1)
    Destroying test database for alias 'default'...

Particularly, this is the line of my code causing the error:

`response = self.client.post(url, data)`

And this is the line of [your code](https://github.com/django-json-api/django-rest-framework-json-api/blob/develop/rest_framework_json_api/renderers.py#L419) that responsible of the error

`view = renderer_context.get("view", None)`

Simply because    `renderer_context` is None

All my test was passing before switching to this library (using default DRF renders)

Here is my configurations: 

    REST_FRAMEWORK = {
        'PAGE_SIZE': 5,
        'EXCEPTION_HANDLER': 'rest_framework_json_api.exceptions.exception_handler',
        'DEFAULT_PAGINATION_CLASS':
            'rest_framework_json_api.pagination.PageNumberPagination',
        'DEFAULT_PARSER_CLASSES': (
            'rest_framework_json_api.parsers.JSONParser',
            'rest_framework.parsers.FormParser',
            'rest_framework.parsers.MultiPartParser'
        ),
        'DEFAULT_RENDERER_CLASSES': (
            'rest_framework_json_api.renderers.JSONRenderer',
[...]

PR #322, opened 2017-02-06
Add CursorPagination, enhance test client to handle client.post better

erichonkanen:
Opening this preemptively for discussion:

Aside from adding a cursor paginator I think we need to address an issue that was already brought up by @abdulhaq-e  and others where testing code that uses this package will break on `APITestClient().post` as discussed here https://github.com/django-json-api/django-rest-framework-json-api/issues/275

Will resolve https://github.com/django-json-api/django-rest-framework-json-api/issues/321 and  https://github.com/django-json-api/django-rest-framework-json-api/issues/320

### DIFF
--- a/rest_framework_json_api/renderers.py
+++ b/rest_framework_json_api/renderers.py
@@ -5,7 +5,7 @@ import copy
 from collections import OrderedDict
 
 import inflection
-from django.db.models import Manager, QuerySet
+from django.db.models import Manager
 from django.utils import six, encoding
 from rest_framework import relations
 from rest_framework import renderers
@@ -396,7 +396,7 @@ class JSONRenderer(renderers.JSONRenderer):
             resource_data.append(('links', {'self': resource[api_settings.URL_FIELD_NAME]}))
         return OrderedDict(resource_data)
 
-    def render_relationship_view(self, data, accepted_media_type=None, renderer_context=None):
+    def render_relationship_view(self, data, accepted_media_type=None, renderer_context={}):
         # Special case for RelationshipView
         view = renderer_context.get("view", None)
         render_data = OrderedDict([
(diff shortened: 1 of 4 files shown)


eriktrom:
👍

rest_framework_json_api/renderers.py
```diff
@@ -396,7 +396,7 @@ def build_json_resource_obj(cls, fields, resource, resource_instance, resource_n
             resource_data.append(('links', {'self': resource[api_settings.URL_FIELD_NAME]}))
         return OrderedDict(resource_data)
 
-    def render_relationship_view(self, data, accepted_media_type=None, renderer_context=None):
+    def render_relationship_view(self, data, accepted_media_type=None, renderer_context={}):
```
santiavenda2: Using a mutable default parameter is generally a bad idea in python, due to:

> Python's default arguments are evaluated once when the function is defined, not each time the function is called (like it is in say, Ruby). This means that if you use a mutable default argument and mutate it, you will and have mutated that object for all future calls to the function as well.

[more details](https://www.google.com.ar/url?sa=t&rct=j&q=&esrc=s&source=web&cd=2&cad=rja&uact=8&ved=0ahUKEwiI7uyWk4HSAhUGjJAKHRWMAGIQFggcMAE&url=http%3A%2F%2Fdocs.python-guide.org%2Fen%2Flatest%2Fwriting%2Fgotchas%2F&usg=AFQjCNF849mrqlMC96-oZkr80MSgOb_R4w)

erichonkanen: @santiavenda2 fair point, I guess this package needs to add better exception handling then since right off the bat if `renderer_context` is the default of `None`, the first line of code will fail. We're looking at other options though as this package lacks documentation and doesn't appear to be active anymore. The added complexity layer on top of DRF which is on top of Django just makes it very hard to deal with.

mblayman: Does this mean you're ok closing this PR? This comment makes it seem to me like this is not important to you anymore, but I might be misreading. I'm wondering if I should spend the time to review or not.

PR #330, opened 2017-03-01
Fixed crash when patching many-to-many relationships in Django >= 1.9

DmitriyZverev:
The app crash with error: `TypeError: add() got an unexpected keyword argument 'bulk'`, when trying to patch many-to-many relationships. In file `example/tests/test_views.py` was added the test catching this error. It's error is possible only in Django >= 1.9, because was added the `bulk` argument in `add()` method for all `Manager` classes, excluding `ManyRelatedManager` class. In file `rest_framework_json_api/views.py` was added check the class name to fix this error.

### DIFF
--- a/example/tests/test_views.py
+++ b/example/tests/test_views.py
@@ -118,12 +118,31 @@ class TestRelationshipView(APITestCase):
         response = self.client.get(url)
         assert response.data == request_data['data']
 
-    def test_patch_to_many_relationship(self):
+    def test_patch_one_to_many_relationship(self):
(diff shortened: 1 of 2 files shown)


mblayman:
@DmitriyZverev, thanks for the contribution! I'll check out the details of the branch when I have time to investigate the issue and your proposed solution.

mblayman:
This looks pretty good to me. Doing an `if` check on a class name feels non-optimal to me, but Django does not seem to provide many reasonable alternatives.

Thanks! 👍

PR #333, opened 2017-03-14
Fix/auto prefetch with m2m

bor3ham:
Previously including relations with the optimised viewset through m2m's would crash as it was not correctly getting the model from the related descriptor

For example from the tests:

```
response = client.get(reverse("entry-list") + '?include=comments.author.bio')
```

The tests already covered the right use case, but were not using the prefetching viewset.

I had to update the test models / serializers slightly to support the auto prefetching (which requires the field names in the serializer match the models). I will create an issue to address this as an improvement

### DIFF
--- a/rest_framework_json_api/views.py
+++ b/rest_framework_json_api/views.py
@@ -6,13 +6,17 @@ from django.db.models.query import QuerySet
 from django.db.models.manager import Manager
 if django.VERSION < (1, 9):
     from django.db.models.fields.related import (
-        ReverseSingleRelatedObjectDescriptor as ForwardManyToOneDescriptor,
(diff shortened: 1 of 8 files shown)


rest_framework_json_api/views.py
```diff
@@ -32,7 +36,7 @@
 
 class ModelViewSet(viewsets.ModelViewSet):
     def get_queryset(self, *args, **kwargs):
-        qs = super().get_queryset(*args, **kwargs)
+        qs = super(ModelViewSet, self).get_queryset(*args, **kwargs)
```
mblayman: Oh, nice! Thanks for this one.

mblayman:
@bor3ham thanks for this work. I really like the improved coverage. 👍

PR #336, opened 2017-04-02
Avoid choices evaluation on relational fields getting field metadata

santiavenda2:
On relation fields line 131 is getting all the related objects.


mblayman:
Looks good to me. 👍

PR #345, opened 2017-05-07
Fix for apps that don't load contenttypes module.

amw:
PR #319 brought support for generic relations. Unfortunately apps that don't add contenttypes to it's INSTALLED_APPS would crash and burn:

```
RuntimeError: Model class django.contrib.contenttypes.models.ContentType doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS.
```

I would like to add a test for this, but it would require two separate Django configs. Not sure how you would want to handle that in py.test.

Also using `type(something) is object()` comparison as a safer alternative to `type(something) is type(None)`. If `something` happened to be `None` we would enter a branch that was never supposed to run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ v2.3.0
 * When `JSON_API_FORMAT_KEYS` is False (the default) do not translate request
   attributes and relations to snake\_case format. This conversion was unexpected
   and there was no way to turn it off.
+* Fix for apps that don't use `django.contrib.contenttypes`.
 
 v2.2.0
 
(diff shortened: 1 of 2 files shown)


mblayman:
I agree that a test would be nice, but this might be one of those cases where the setup for that test would be extreme overkill since it involves things that happen at import time.

What do you think of adding a `# pragma: no cover` to the `if` clause that checks for contenttypes in `INSTALLED_APPS`? It seems like that might be a clue to future readers that this was considered and declared "not worth testing."

amw:
Added.

mblayman:
👍

PR #346, opened 2017-05-07
Fix Django deprecation warning

amw:
```
RemovedInDjango20Warning: Importing from django.core.urlresolvers is deprecated in favor of django.urls
```

### DIFF
--- a/rest_framework_json_api/views.py
+++ b/rest_framework_json_api/views.py
@@ -1,6 +1,9 @@
 import django
 from django.core.exceptions import ImproperlyConfigured
-from django.core.urlresolvers import NoReverseMatch
+if django.VERSION >= (1, 10):
+    from django.urls import NoReverseMatch
+else:
+    from django.core.urlresolvers import NoReverseMatch
 from django.db.models import Model
 from django.db.models.query import QuerySet
 from django.db.models.manager import Manager


mblayman:
You were busy today. 😄 Thanks, @amw! 👍

lucacorti:
This now breaks `drf-j-a` on Django 2.0 rc.

mblayman:
Do you mean that this fix breaks DJA on Django 2.0? Or do you mean that, because this fix is not in a released version on PyPI, DJA is broken on Django 2.0? I'm trying to figure out if the thing to fix this issue is to upload a release to PyPI.

lucacorti:
@mblayman Sorry, I mean this needs a PyPI release to be compatible with Django>=2.0.0 beacause the issue fixed by this this pull request has upgraded from deprecation to error in the 1.11 -> 2.0 upgrade.

lucacorti:
Also, this could be improved since Django 2.0.0 release notes are encouraging app authors to drop support for Django <= 1.11 so the version check on import could be removed if a new release is targeting Django>=1.11.

mblayman:
Yeah, I agree. I think I need to put out a release that includes support for the older versions and the newer versions so there is a migration path, then the project can start dropping some of the older stuff (which will be nice because some of the imports are a mess).

lucacorti:
Yes, these two could be probably be tackled separately. A minor release as-is to make drf-ja work with Django>=2.0.0 and a new major with a larger refactoring to cleanup the compatibility stuff.

lucacorti:
`relations.py` has the same import and errors out on Django >= 2.0. Can this be fixed as a point release?

lucacorti:
See #392

PR #347, opened 2017-05-11
Added nested included serializer support for remapped relations

sliverc:
In ResourceRelatedField it is possible to rename/remap a field so it can be represented in the api with a different name using the source attribute. This works well however if I request to include this relationship in `included` it won't be returned.

I have added a test which should make it clearer what issue this PR addresses.

### DIFF
--- a/rest_framework_json_api/renderers.py
+++ b/rest_framework_json_api/renderers.py
@@ -273,6 +273,30 @@ class JSONRenderer(renderers.JSONRenderer):
         return utils.format_keys(data)
 
     @classmethod
+    def extract_relation_instance(cls, field_name, field, resource_instance, serializer):
+        relation_instance = None
+
+        try:
+            relation_instance = getattr(resource_instance, field_name)
+        except AttributeError:
+            try:
+                # For ManyRelatedFields if `related_name` is not set
+                # we need to access `foo_set` from `source`
+                relation_instance = getattr(resource_instance, field.child_relation.source)
+            except AttributeError:
+                if hasattr(serializer, field.source):
+                    serializer_method = getattr(serializer, field.source)
+                    relation_instance = serializer_method(resource_instance)
+                else:
+                    # case when source is a simple remap on resource_instance
+                    try:
+                        relation_instance = getattr(resource_instance, field.source)
+                    except AttributeError:
+                        pass
+
+        return relation_instance
+
+    @classmethod
     def extract_included(cls, fields, resource, resource_instance, included_resources):
         # this function may be called with an empty record (example: Browsable Interface)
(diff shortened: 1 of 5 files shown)


mblayman:
Hey, @sliverc, this branch now has some conflicts because of the work you did on the flake8 branch. Could you fix that up and also add a CHANGELOG entry for this too? I'm not sure I'll have time to review this today, but I will get to it when I can. Thanks!

sliverc:
@mblayman Solved merge conflicts and added CHANGELOG entry

rest_framework_json_api/renderers.py
```diff
@@ -273,6 +273,30 @@ def extract_relationships(cls, fields, resource, resource_instance):
         return utils.format_keys(data)
 
     @classmethod
+    def extract_relation_instance(cls, field_name, field, resource_instance, serializer):
```
mblayman: Could you please add a docstring which provides a conceptual overview of what's going on here? I'm finding the four nested levels to be really intense. If you could use PEP 257 style, that would be great.

```python
"""Extract ...

<relevant details that might help understand what's going on here>
"""
```

This many levels of exceptions makes me wonder if there might performance challenges here (because of the cost of going through the exception stack over and over). I don't think it's fair to ask you for a performance test so I'm mostly wondering "out loud" if this is too much exception handling as a way to do control flow. It does feel like a bit of a code smell. Maybe that's the cost of working with a spec that's so dynamic ¯\\_(ツ)_/¯

sliverc: I have added a docstring.

Considering the levels of exceptions I also think this is certainly not ideal but I cannot think of an alternative way to do it though.

For example calling `getattr` with default argument `None` wouldn't work, as an attribute could actually be `None`. Calling `hasattr` wouldn't make things better either as internally it also does a try-except (https://docs.python.org/3/library/functions.html#hasattr).
It could be argued whether the serializer method case is really needed but otherwise I guess we will need to live with it.

mblayman: Thanks for thinking about it. The only alternative that I can think of is to have multiple returns. What do you think of something like the following:

```python
         try:
             return getattr(resource_instance, field_name)
         except AttributeError:
             pass

         try:
             # For ManyRelatedFields if `related_name` is not set
             # we need to access `foo_set` from `source`
             return getattr(resource_instance, field.child_relation.source)
         except AttributeError:
             pass

        <and so on...>
```

Is that any "better?" Or is it worse? It would remove the deep nesting even if the logic is functionally the same.

mblayman:
Thanks, @sliverc. I think this will work well. I appreciate the contribution and dialog.

PR #353, opened 2017-05-24
Fix for pagination ‘last’ link to give the real last page

YanivP:
We're working on a project using django-json-api and came across unintuitive behavior for the pagination. The current pagination implementation for the 'last' link is basically a 'total count' - 'limit'. Consider the following example:

Objects = [0-21]
Limit = 10
(first page - offset 0)
0,1,2,3,4,5,6,7,8,9

(clicking 'next' - offset 10)
10,11,12,13,14,15,16,17,18,19

(clicking 'next' - offset 20)
20,21

(clicking 'last' - offset 12)
12,13,14,15,16,17,18,19,20,21

This behavior would make the user completely disoriented in the paging. It makes more sense that 'last' will offset to 20 just the same as clicking 'next'->'next'->'next' and simply return two objects instead of 10.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -4,4 +4,4 @@ Greg Aker <greg@gregaker.net>
 Jerel Unruh <mail@unruhdesigns.com>
 Matt Layman <http://www.mattlayman.com>
 Oliver Sauder <os@esite.ch>
-
+Yaniv Peer <yanivpeer@gmail.com>
--- a/example/tests/unit/test_pagination.py
+++ b/example/tests/unit/test_pagination.py
@@ -41,7 +41,7 @@ class TestLimitOffset:
         offset = 10
         limit = 5
         count = len(self.queryset)
-        last_offset = count - limit
+        last_offset = (count // limit) * limit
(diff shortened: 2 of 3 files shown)


mblayman:
Hi, @YanivP. Thanks for your contribution! It looks like you're a first time contributor so I'm excited that you took the time to help.

The PR, in its current state, can't be merged because the test suite is reporting failure (you can use the error at https://travis-ci.org/django-json-api/django-rest-framework-json-api/builds/235614695?utm_source=github_status&utm_medium=notification). Either the fix is not quite correct and the tests are saving us from making a mistake, or the test is not correct. Could you please look into what the issue is?

Also, please add your name to the `AUTHORS` file to give you some credit for your contribution. The listing is alphabetical by first name.

Let me know if you have any questions. Thanks again!

YanivP:
Hey @mblayman 
You're right I forgot to run the tests, fixed and also added myself to AUTHORS

mblayman:
👍 Good stuff. Thank you!

Issue #354, opened 2017-05-30
Reverse Many to Many relationships returns wrong type


Darkheir:
The reverse m2m relationships are not returning the good type.

mblayman:
Fixed by #355

PR #355, opened 2017-05-30
Fixes types in reverse m2m relationships

Darkheir:
In case of reverse m2m relationships the wrong type was returned.

I created a test to reproduce this situation.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -4,4 +4,5 @@ Greg Aker <greg@gregaker.net>
 Jerel Unruh <mail@unruhdesigns.com>
 Matt Layman <http://www.mattlayman.com>
 Oliver Sauder <os@esite.ch>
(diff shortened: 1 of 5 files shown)


Darkheir:
fixes #354

Darkheir:
It seems that my new test triggers an error on Django 1.8 version.

The error is not on the lines added by my commits

mblayman:
Hi, @Darkheir. Thanks so much for your contribution! The extra test coverage is really valued. I like the code that I see here. I'd like for all contributors to be recognized for their work so could you add your name to the `AUTHORS` file? The list of names is alphabetical by first name.

I'm happy to merge after that.

Darkheir:
Hi, thank you for your kind answer. I just added my name to the contributors.

mblayman:
Great, thanks again!

Issue #358, opened 2017-06-21
related_link_url_kwarg doesnt work with non pk value


ola-t:
Seems like `related_link_lookup_field` is not respected in `SerializerMethodResourceRelatedField`:

```
class BookSerializer(serializers.HyperlinkedModelSerializer):

    authors = relations.SerializerMethodResourceRelatedField(
        source='get_authors',
        model=Author,
        many=True,
        read_only=True,
        related_link_view_name='books-authors-list',
        related_link_url_kwarg='ext_book_id',
        related_link_lookup_field='ext_book_id',
    )

```

```
  File "/src/django-rest-framework-json-api/rest_framework_json_api/renderers.py", line 140, in extract_relationships
    field_links = field.get_links(resource_instance)
  File "/src/django-rest-framework-json-api/rest_framework_json_api/relations.py", line 125, in get_links
    related_kwargs = {self.related_link_url_kwarg: kwargs[self.related_link_lookup_field]}
KeyError: 'ext_book_id'
```

Looks like https://github.com/django-json-api/django-rest-framework-json-api/blob/develop/rest_framework_json_api/renderers.py#L140 `get_links` is missing `lookup_field`
```
kwargs: {'ext_book_id': 256} http://localhost:8000/api/books/256/authors/
```

Issue #356, opened 2017-06-10
Demo site 500 


elena:
Following links on demo site return Server Error 500:
[http://json-api.jerel.co/](http://json-api.jerel.co/)

```
       "blogs": "http://json-api.jerel.co/blogs",
       "entries": "http://json-api.jerel.co/entries",

```

Note other links work:
```
        "authors": "http://json-api.jerel.co/authors",
        "comments": "http://json-api.jerel.co/comments"

```
Cheers.

mblayman:
Thanks, @elena. I believe this runs on the personal website of @jerel. If I had to venture a guess, the site is probably missing some database migrations that were recently added the demo project.

jerel:
It's running on a private server. I'll try to take a look, probably an
issue with the reset script or migrations

On Jun 10, 2017 7:32 PM, "Matt Layman" <notifications@github.com> wrote:

Thanks, @elena <https://github.com/elena>. I believe this runs on the
personal website of @jerel <https://github.com/jerel>. If I had to venture
a guess, the site is probably missing some database migrations that were
recently added the demo project.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/django-json-api/django-rest-framework-json-api/issues/356#issuecomment-307598252>,
or mute the thread

jerel:
So, it's throwing `ImportError: No module named polymorphic`. Merely installing `django-polymorphic` doesn't fix it. Evidently there's a package missing from the requirements file. But does, or should, the demo even use polymorphism out of the box?

mblayman:
@jerel, I believe that the polymorphic branch that was merged *did* update the demo project to use polymorphic models in some locations. I'm surprised (saddened?) that installing `django-polymorphic` doesn't solve the problem.

Do you have any details on what it is reporting after you install `django-polymorphic`? Did you get past the `ImportError`?

jerel:
So, I've worked with this a bit and found the demo hadn't applied migrations correctly. The `/blogs` endpoint is working fine but `/entries` is throwing `Could not resolve URL for hyperlinked relationship using view name "entry-relationships".`

I'm out of time for the moment but hopefully that gets somebody else pointed in the right direction. You can replicate by checking out the `develop` branch and running `django-admin.py runserver --settings=example.settings`

mblayman:
[...]

PR #359, opened 2017-06-21
respect custom related_link_url_kwarg

ola-t:
fix https://github.com/django-json-api/django-rest-framework-json-api/issues/358

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -8,3 +8,4 @@ Matt Layman <http://www.mattlayman.com>
 Oliver Sauder <os@esite.ch>
 Raphael Cohen <raphael.cohen.utt@gmail.com>
 Yaniv Peer <yanivpeer@gmail.com>
+Ola Tarkowska <ola@red-aries.com>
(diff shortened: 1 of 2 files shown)


mblayman:
Hi, @ola-t! Thanks for the contribution. The fix looks good to me. 👍

As this is a community run project, we'd like for every contributor to get credit for their work. Could you add your name to the AUTHORS file (it's alphabetical by first name)? I'm happy to merge the work after that.

Darkheir:
This doesn't fix the problem in the many-many relations. 

The folowing line should also be corrected :
https://github.com/ola-t/django-rest-framework-json-api/blob/ef3e61f55e9a89b30e5c6147eeafe92145a82d27/rest_framework_json_api/renderers.py#L193

To be
```python
field_links = field.child_relation.get_links(resource_instance, field.child_relation.related_link_lookup_field)
```

mblayman:
@Darkheir, thanks for thinking about that case. If you could make a separate pull request, that would probably be the quickest fix for the many-to-many issue. I'm going to go ahead and merge this one.

ola-t:
When are you planning to release these bug fixes?

mblayman:
I would like to see a release soon. I either need to get upload permission to PyPI to do it or @jerel will need to release the package. Either action will require some input from him that's out of my control.

Before the release, I think the demo site needs to be fixed so #356 is a blocker that should be addressed.

PR #360, opened 2017-06-26
Add isort to CI process

### LINKED ISSUE #344: Add isort to the PR/CI process
Specifying the order of imports is definitely something that can be handled with a tool. To make the project's imports consistent, isort can be added to the CI toolchain. This will remove the need to make comments on PRs that might seem silly.

mblayman:
This branch adds `isort` to the CI process so that import ordering issues can be left to a tool. It also does an initial run through on the code so that everything respects the isort settings.

Fixes #344

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,10 @@ install:
   - python setup.py install
 script:
   - flake8
-  - isort --check-only --verbose --recursive --diff rest_framework_json_api example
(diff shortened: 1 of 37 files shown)


mblayman:
This is an annoying dev vs CI issue. I'll keep poking on this later, but the tl;dr is that isort works locally in my dev env, but then fails to consider django-polymorphic and pytest_factoryboy as third party packages on Travis.

PR #362, opened 2017-06-30
[WIP] Fix example app

### LINKED ISSUE #356: Demo site 500

astronouth7303:
Some miscellaneous bugs, some of which stemmed from polymorphic. Fixes #356 

* [x] The example now requires `packaging` (`example/serializers.py`). I'm guessing this is the actual cause of the bug.
* [ ] `/entries` produces `ImproperlyConfigured at /entries`: `Could not resolve URL for hyperlinked relationship using view name "entry-relationships".`
* [x] `/projects` produces `AttributeError at /projects`: `'list' object has no attribute '_meta'`

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -69,7 +69,9 @@ From Source
 ## Running the example app
 
     git clone https://github.com/django-json-api/django-rest-framework-json-api.git
-    cd django-rest-framework-json-api && pip install -e .
+    cd django-rest-framework-json-api
+    pip install -e .
+    pip install -r example/requirements.txt
     django-admin.py runserver
 
 Browse to http://localhost:8000
--- a/example/requirements.txt
+++ b/example/requirements.txt
@@ -1,0 +1,2 @@
+# Requirements specifically for the example app
+packaging
--- a/example/tests/integration/test_polymorphism.py
+++ b/example/tests/integration/test_polymorphism.py
@@ -78,6 +78,30 @@ def test_polymorphism_on_polymorphic_model_list_post(client):
     assert content['data']['attributes']['artist'] == test_artist
 
 
+def test_polymorphic_model_without_any_instance(client):
+    expected = {
+        "links": {
+            "first": "http://testserver/projects?page=1",
+            "last": "http://testserver/projects?page=1",
+            "next": None,
+            "prev": None
+        },
+        "data": [],
+        "meta": {
+            "pagination": {
+                "page": 1,
+                "pages": 1,
+                "count": 0
+            }
+        }
+    }
+
+    response = client.get(reverse('project-list'))
+    assert response.status_code == 200
+    content = load_json(response.content)
+    assert expected == content
+
+
 def test_invalid_type_on_polymorphic_model(client):
     test_topic = 'New test topic {}'.format(random.randint(0, 999999))
     test_artist = 'test-{}'.format(random.randint(0, 999999))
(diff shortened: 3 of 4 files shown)


astronouth7303:
@leo-naeka if you know this better, I would love some insight. I'm not actually _that_ familiar with DRF.

leo-naeka:
I'll check this today.

leo-naeka:
Actually the `ImproperlyConfigured at /entries: Could not resolve URL for hyperlinked relationship using view name "entry-relationships".` bug was introduced before the polymorphic feature merge: occurs in 629d1d6.

The last error is polymorphic-related indeed.

mblayman:
@Darkheir, one of the errors that's happening on the demo site might be related one of your commits. Could you take a look at the issue? Once we fix the demo site, I'd love to put out a release that includes everyone's work from the last couple of months.

Darkheir:
Hi,

I quickly checked using the debugger and the code I added is not called during the process ...

leo-naeka:
After some investigation this endpoint mays have been errored since Nov. 8th 2016 with commit 1941c34.

- 1941c34 introduced the `Could not resolve URL for hyperlinked relationship using view name "entry-relationships".`
- fbe49a1 introduced the `'EntrySerializer' object has no attribute 'partial'` error which supersede 1941c34's one but occurred only with `DRF>=3.6.3`
- a12d98d fixed the `'EntrySerializer' object has no attribute 'partial'` and we got back the first one 'til now...

leo-naeka:
The fact is that [relationship patterns](https://github.com/django-json-api/django-rest-framework-json-api/blob/develop/example/urls_test.py#L38-L53) are not included in regular `urls.py` as they are in `urls_test.py`. I'm not very familiar with the relationships code, but adding them seems to fix the bug...

leo-naeka:
@astronouth7303 found the bug in the `PolymorphicModelSerializer` causing the last error. Fixed and added a test case locally. Could you please grant me push to your repo in order to fix all of these in this PR?

leo-naeka:
@mblayman @astronouth7303 @jsenecal any thoughts on this? In particular the second error and the missing relationship patterns in `urls.py`? @astronouth7303 the last error has been fixed, so should be checked on the OP.

mblayman:
@leo-naeka I won't have much time to look at this for a while, but I hope to give it some more time in the future. I took a quick peek and I have a question. What do you mean by "missing relationship patterns"? The second bullet is an error with `/entries` and it looks like that is in `urls.py`. I'm not clear on your meaning of "missing" in this context. 

btw, thanks for looking into these issues!

leo-naeka:
This is an error with the`entry-relationships` endpoint which is not part of the `entries` viewset.  
This is only defined in `urls_test.py` [[code]](https://github.com/django-json-api/django-rest-framework-json-api/blob/develop/example/urls_test.py#L38-L53) but missing in `urls.py` [[code]](https://github.com/django-json-api/django-rest-framework-json-api/blob/develop/example/urls.py#L24)

 I guess this is why tests aren't failing whereas demo does.

mblayman:
Could we solve the remaining problems by moving the routes from the tests to the regular `urls.py` file? I think this branch has made things better and if moving the routes is something that would fix the demo site, then I think that would be good enough to merge so we can get closer to releasing.

mblayman:
Time flies. I think this code is good enough to merge because it solved some problem. Since there hasn't been movement on the remaining task, I'll merge and we can solve that later.

PR #364, opened 2017-07-20
Add Django Debug Toolbar, and add a failing test for query explosions

aidanlister:
This PR adds Django Debug Toolbar, makes the example app easier to performance test in the browsable API (e.g. eliminates extra queries caused by lack of ordering on the models, and stops the browsable forms from making queries) and adds a failing test for query explosions. I also flake8'd a few files (mainly fixed import * stuff).

This should give everyone a pretty clean base to play with JSON API with.


mblayman:
Hey @aidanlister! Thanks for the PRs. I'm excited to get some of these performance rough edges smoothed out. Both of the current PRs have failing builds, if you can fix those things up, I can take a look at the code once the CI build is green.

PR #366, opened 2017-07-21
Issue #337

### LINKED ISSUE #337: Query explosion even on simple models with no included serializers
I love the idea of this module but it seems like it's impossible to build fast APIs with it.

Let's say I have Books and Authors. I fetch a list of books, even with no "included" specified I get a query for each author, just to display the FK?

As a broader issue, as Author increases in complexity developers will add more included_serializers to that serialiser, which inadvertently affects my Books viewset with a query explosion.

Could we define a lambda for each included_serializer to alter the query? Anything more robust than magic.

jsenecal:
- Adds Django Debug Toolbar, and add a failing performance (query count) test
- Adds a helper for specifying a prefetch_for_related attribute on your viewset to help with prefetching includes.
- Fixes some of the example code

Thanks to @aidanlister for his contribution on this.

Merging this will close #337


aidanlister:
Looks great!

mblayman:
@jsenecal It looks like Travis is failing because of isort (curiously, only Python 3.4 🤔 ). I think isort gets a little tricky because the config file is trying to be correct for the rest_framework_json_api, but it also has to satisfy the included example package. I'm guessing you can solve this by including `--thirdpart packaging` to the isort command that checks example.

```bash
$ isort --check-only --verbose --recursive --diff --thirdparty pytest \
    --thirdparty polymorphic --thirdparty pytest_factoryboy \
    --thirdparty packaging example
```

I'll look at the code closer to give it some review.

.gitignore
```diff
@@ -40,4 +40,7 @@ pip-delete-this-directory.txt
 # VirtualEnv
 .venv/
 
+# Developers
 *.sw*
+manage.py
```
mblayman: Why is this ignoring `manage.py`?

example/settings/dev.py
```diff
@@ -74,7 +79,13 @@
...
     'DEFAULT_RENDERER_CLASSES': (
         'rest_framework_json_api.renderers.JSONRenderer',
-        'rest_framework.renderers.BrowsableAPIRenderer',
+
+        # If you're performance testing, you will want to use the browseable API
+        # without forms, as the forms can generate their own queries.
+        # If performance testing, enable:
+        'example.utils.BrowsableAPIRendererWithoutForms',
+        # Otherwise, to play around with the browseable API, enable:
+        # 'rest_framework.renderers.BrowsableAPIRenderer',
```
mblayman: Since `example` is what runs on the demo site, should we leave the form version on by default instead of the perf version so that people can add extra data to play with it?

rest_framework_json_api/relations.py
```diff
@@ -1,10 +1,15 @@
...
+from collections import OrderedDict
 
 import inflection
+import six
+from django.core.exceptions import ImproperlyConfigured
+from django.core.urlresolvers import NoReverseMatch
 from django.utils.translation import ugettext_lazy as _
 from rest_framework.fields import MISSING_ERROR_MESSAGE
-from rest_framework.relations import *  # noqa: F403
+from rest_framework.relations import MANY_RELATION_KWARGS, PrimaryKeyRelatedField
```
mblayman: 👏 I'm a fan of any time a `*` import can be removed.

mblayman:
I like what's here a lot!

What do you think about documentation for this stuff? IMO, features don't really exist until they are documented so that users know how to use them. With that stated, it doesn't matter to me if the features are documented in this branch or in some other branch as long as there is some prose before in the docs before the next release.

Once Travis is green, I'm ok with merging. If you decide to add some docs, I can give it another look.
 
Thanks for pulling this together with @aidanlister. 👍

.gitignore
jsenecal: @aidanlister must have a custom one.

We're not using it to start the app anyways (`$ django-admin.py runserver --settings=example.settings`)

example/settings/dev.py
jsenecal: I guess it would make sense indeed - let me fix this

example/views.py
```diff
@@ -74,8 +74,13 @@ class AuthorViewSet(ModelViewSet):
 
 
 class CommentViewSet(ModelViewSet):
-    queryset = Comment.objects.all()
+    queryset = Comment.objects.select_related('author', 'entry')
```
jsenecal: I still have an issue with the way relations are extracted first level... I had to prefetch things that I shouldn't have to only to make tests pass... Feels like the queryset might be over reaching to overcome an issue with drf_jsonapi.

jsenecal: So, it really happens in https://github.com/ABASystems/django-rest-framework-json-api/commit/b96d79d49995c0faac6174ecdd1715a2b1121d3d merged a year ago.

@jerel do you remember the conversation that was around this ?

jsenecal:
@mblayman I did add some docs. 

See my notes about [this](https://github.com/django-json-api/django-rest-framework-json-api/pull/366/files/a7130b9a29fb3ce5325d8f571a11766294dfc26d#diff-e009eaa5df551557d6bbf2300acb072b) and b96d79d49995c0faac6174ecdd1715a2b1121d3d

Should we open another issue for that ?

aidanlister:
@jsenecal thanks for including my cite, might be better to give some context there though? Let me write something:

aidanlister:
> Using the helper to prefetch, rather than attempting to minimise queries via select_related might give you better performance depending on the characteristics of your data and database.
> 
> For example:
> 
> > If you have a single model, e.g. Book, which has four relations e.g. Author, Publisher, CopyrightHolder, Category.
>  > To display 25 books and related models, you would need to either do:
>  > a) 1 query via selected_related, e.g. SELECT * FROM books LEFT JOIN author LEFT JOIN publisher LEFT JOIN CopyrightHolder LEFT JOIN Category
>  > b) 4 small queries via prefetch_related.
>  >
>  > If you have 1M books, 50k authors, 10k categories, 10k copyrightholders in the select_related scenario, you've just created a in-memory table with 1e18 rows which will likely exhaust any available memory and slow your database to crawl. The prefetch_related case will issue 4 queries, but they will be small and fast queries.

example/views.py
jsenecal: After further investigation, the problem does not occur inside extract_relationships - Still looking

aidanlister: @jsenecal the problem is in two places, setting use_pk_only_optimization to False, and extract_attributes. You have to whack them both, which made it so hard to find the source of the explosion.

I think you should probably remove the select_related in this because the tests should fail until we fix that other issue?

jsenecal: Thanks for this - I merged already but will update my branch to reflect this conversation

Issue #158, opened 2015-11-22
exception on import of the serializer


patroqueeet:
Dear,

my code:
```
from rest_framework_json_api import serializers
```

when using `from rest_framework import serializers` all is working.

the error when executing runserver and triggering a request:
```
Traceback (most recent call last):
  File "/usr/lib/python2.7/wsgiref/handlers.py", line 85, in run
    self.result = application(self.environ, self.start_response)
  File "/srv/www/vbnet/.ve/local/lib/python2.7/site-packages/django/contrib/staticfiles/handlers.py", line 64, in __call__
    return self.application(environ, start_response)
  File "/srv/www/vbnet/.ve/local/lib/python2.7/site-packages/django/core/handlers/wsgi.py", line 168, in __call__
    self.load_middleware()
  File "/srv/www/vbnet/.ve/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 46, in load_middleware
    mw_instance = mw_class()
  File "/srv/www/vbnet/.ve/local/lib/python2.7/site-packages/django/middleware/locale.py", line 23, in __init__
    for url_pattern in get_resolver(None).url_patterns:
  File "/srv/www/vbnet/.ve/local/lib/python2.7/site-packages/django/core/urlresolvers.py", line 372, in url_patterns
    patterns = getattr(self.urlconf_module, "urlpatterns", self.urlconf_module)
  File "/srv/www/vbnet/.ve/local/lib/python2.7/site-packages/django/core/urlresolvers.py", line 366, in urlconf_module
    self._urlconf_module = import_module(self.urlconf_name)
  File "/usr/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/srv/www/vbnet/site/project/urls.py", line 13, in <module>
    from location.views import VenueDetailView, VenueListView, \
  File "/srv/www/vbnet/site/location/views.py", line 19, in <module>
    from services.rest.serializers import RatingSerializer
  File "/srv/www/vbnet/site/services/rest/serializers.py", line 7, in <module>
    from rest_framework_json_api import serializers
  File "/srv/www/vbnet/.ve/local/lib/python2.7/site-packages/rest_framework_json_api/serializers.py", line 5, in <module>
    from rest_framework_json_api.relations import ResourceRelatedField
  File "/srv/www/vbnet/.ve/local/lib/python2.7/site-packages/rest_framework_json_api/relations.py", line 7, in <module>
[...]

PR #372, opened 2017-09-14
support polymorphic list with proper type/fields resolution

robertobarreda:
Render polymorphic list of items with different fields/type


mblayman:
Hi @robertobarreda. Thanks for the contribution. The project's policy is to only merge work that passes all continuous integration tests. If you're still interested in this branch, could you make sure that CI is passing. I'm happy to review once that check is out of the way. If that's not something you have time for, then we can close out the PR.

mblayman:
Ah, it looks like CI might actually be failing because of a factory boy dependency problem. Once #373 is merged, you can probably merge develop into your branch to fix CI.

mblayman:
hi @robertobarreda, it looks like this might be close to ready. I think if you merge the `develop` branch into this, the branch will get the necessary package update to make the CI build work.

robertobarreda:
Sorry @mblayman, I merged the develop branch but CI is still failing :(

mblayman:
Looks like there are a couple of issues:
1. The branch has a flake8 error which needs to be fixed.
2. There's an import problem. I did a Google search on the exception and found another issues from *this* repository that's similar. https://github.com/django-json-api/django-rest-framework-json-api/issues/158 My best guess is that there is some kind of circular import error. Maybe, instead of importing PolymorphicModelSerializer directly, you can import `serializers` and access the model serializer like `serializers.PolymorphicModelSerializer`.

Thanks for your diligence in working on this. I'm sure that users who use the polymorphic features will be grateful.

adsized:
Just found this pull request and realised that it would probably solve the issue #400 I raised as well as perhaps help with issue #401 which seemed similar.

I can see that some tests still seemed to be failing on some of the travis builds that are configured to use older versions of python. @robertobarreda @mblayman Do you know what was blocking getting this merged in?

mblayman:
<snark>Travis CI</snark> 😉 To keep our sanity on the project, we only merge stuff that passes Travis builds. I suspect there is a problem with circular imports with this PR, but I'm not certain. If someone has time to check out the branch, figure the problem out, and get the branch updated to work in CI, I'd be happy to merge this.

adsized:
@mblayman It makes perfect sense only merging stuff that passes the CI builds, I can perfectly understand that :smile:

I have checked out out the code from this pull request and had a look at where the errors are coming from. It does seems they were being caused by the serializers import in the renderers file.

From what I have read, the from x import y syntax requires a module to be completely initialized, whereas a simple import does not. By changing this, all of the tests now pass in my environment under both python2.7 and python3.6.

I'd be quite happy to commit my changes. Is there any easy way I can add a commit to this pull request or would I need to open a new one?

mblayman:
@adsized, thanks for looking into this. I'd suggest putting up a separate PR. We'll make sure that both you and @robertobarreda get credit for the work and get those issues closed out. Please note that we're now working off of `master` instead of the old `develop` branch. I'm not sure if that will cause more work for you or not.

I look forward to seeing the Pull Request! 😄

mblayman:
@adsized fixed the import issues in his PR. Thanks @robertobarreda!

PR #373, opened 2017-09-20
Fix factory boy dependency

santiavenda2:
Factoy Boy 2.9.+ versions are raising this problem when we run the test suite using py.test

ConftestImportFailure: (local('/home/santiago/develop/invgate-django-rest-framework-json-api/example/tests/conftest.py'), (<type 'exceptions.AttributeError'>, AttributeError("'DjangoOptions' object has no attribute 'postgen_declarations'",), <traceback object at 0x7fb1c6b123b0>))

Related issue https://github.com/pytest-dev/pytest-factoryboy/issues/47

To fix this problem temporally  we can switch to FactoryBoy 2.8.1.

### DIFF
--- a/requirements-development.txt
+++ b/requirements-development.txt
@@ -5,6 +5,7 @@ isort
 mock
 pytest>=2.9.0,<3.0
 pytest-django
+# factory_boy is currently broken at 2.9 and above. See: https://github.com/pytest-dev/pytest-factoryboy/issues/47
 factory-boy<2.9.0
 pytest-factoryboy
 recommonmark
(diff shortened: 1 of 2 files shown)


mblayman:
Thanks @santiavenda2. Would you mind adding a comment with a link to the factory boy issue to the code and/or requirements.txt file so that future developers will be able to know when that dependency limitation can be removed in the future?

mblayman:
Nevermind. I took care of adding the comment. Thanks for the fix!

santiavenda2:
Thanks @mblayman  :+1:

PR #374, opened 2017-10-09
Avoid query explosion for related fields where I can user pk_only_opt…

santiavenda2:
…imization.

Also, fix AutoPrefetchMixin for ReverseOneToOneDescriptors.


sliverc:
I tested this change in on one of my projects and it works smoothly. Thanks.

It seems that in drf 3.1 `_readable_fields` doesn't exist so CI is failing. Not sure whether there is an alternative but I would be in favor of dropping support for 3.1 which is already more than 2.5 years old.

santiavenda2:
I have fixed  the problem we had in DRF<3.2 (_readable_field property didn't exist on Serializer class)

I think the change is ready to be merged

mblayman:
I'm going to re-target this PR to `master` since the main development branch is now over there.

mblayman:
@santiavenda2 because this is now targeting `master`, there is a conflict on the branch. `master` dropped a bunch of supported Django version so I think that might provided an opportunity to simplify the PR code.

santiavenda2:
@mblayman done

rest_framework_json_api/serializers.py
```diff
@@ -170,6 +170,47 @@ def get_field_names(self, declared_fields, info):
         fields = super(ModelSerializer, self).get_field_names(declared, info)
         return list(fields) + list(getattr(self.Meta, 'meta_fields', list()))
 
+    def to_representation(self, instance):
```
mblayman: Could you explain why this is overriding `to_representation` from the `ModelSerializer`? It seems like DJA is likely to diverge from the DRF upstream if this method is added. Is there some extra stuff added here that you could call out?

aidanlister:
I've reviewed the PR and we're now testing this PR internally - everything seems to be working as expected - very impressive. Next step we'd do a production release in two weeks, but I would love to have this merged before then.

rest_framework_json_api/serializers.py
aidanlister: @mblayman this is returning the JSON API identifier of the object without diving down the recursive rabbit whole. It looks like an elegant solution.

mblayman: Thanks for the second look, @aidanlister. I took a look at the DRF implementation and the method seems small enough that it would be hard for DJA to diverge too much from upstream.

mblayman:
With the added input from @aidanlister, I think we've had enough eyes looking at this to declare victory. Thanks for the contribution, @santiavenda2!

PR #376, opened 2017-10-11
Allow type field on none polymorphic serializers

sliverc:
Changes in #211 caused serializers with type field to fail, breaking backwards compatibility. This change fixes this by allowing type fields again on none polymorphic serializers.


sliverc:
For CI to run integrated #373

mblayman:
@sliverc, thanks for going through all that extra effort for a relatively small change. 👍

PR #377, opened 2017-10-30
removed duplicate requirement

gregorypease280:
"mock" is included twice in the development requirements. This causes pip install to fail with 
```
Double requirement given: mock (from -r requirements-development.txt (line 15)) (already in mock (from -r requirements-development.txt (line 5)), name='mock')
```


mblayman:
Thanks! I guess there was a merge that got mixed up. 👍

PR #385, opened 2017-12-02
Fix import NoReverseMatch

mapeveri:
Could you please tell me when you will release a new version?


mblayman:
@mapeveri Thanks for the contribution. This branch is currently failing on the import sort order. The fix looks good to me so I'm happy to merge it once the import order is updated (please see a CI build for detail).

mapeveri:
@mblayman I don't understand because this PR failed. I just modified the tests, but it's not that :(

The strange thing is that in travis it's ok. I do not understand where the error is.

image

mblayman:
The CI build runs more than tests. If you look at some of the builds, you can observe some flake8 failures.
image

mblayman:
I've had more time to think on this PR. This import fix is solving a problem that will affect Django 2.0 users (because the old import moved). The currently release version of DJA (2.3.1) does not officially support DRF 3.7.x. The DRF 3.6.x series does not support Django 2.0 (and I don't think it ever will).

PR #387 removes support for the older versions of Django where this older import is required. I think DRF 3.7 support will come after that and, with that, Django 2.0 compatibility.

In light of all of that version information, I think it makes sense to close this PR (and its cousin, #392) in favor of #387. DJA 2.4.x will be the first version to support DRF 3.7.x and Django 2.0.

Even though I won't be merging this PR, I totally appreciate your effort. It lit the fire under me to get out the 2.3.1 DJA release and motivated me to get DJA compatible with the latest versions in the Django ecosystem. Thanks a ton!

mapeveri:
@mblayman When the version that supports Django 2.0 comes out? I mean the date aprox. Thanks!!

mblayman:
I don't know for certain, however, I'd hope to have it done before the new year.

mapeveri:
@mblayman great. Thanks!

PR #387, opened 2017-12-03
Drop support for Django 1.8 - 1.10

### LINKED ISSUE #386: Drop support for 1.8, 1.9, and 1.10
DJA 2.3.0 supports 1.8 - 1.11. At the end of the month, Django 1.10 will no longer be supported by the Django core team and the 1.8 LTS support is ending early next year. Since the 2.3.0 release can act as a transition release, I think it's time for future versions of DJA to support 1.11 and greater. This will permit the project to clean out some weird import hoop jumping that was required to support all these versions.

image

mblayman:
The effect of dropping support for Django 1.8 - 1.10 is pretty huge. Lots of test environments can go away.

Python 3.3 is EOL so I dropped that.

Django 1.8 - 1.10 are (basically) EOL so I dropped them.

Django REST Framework doesn't support Django 1.11 until a bug fix release of DRF 3.6.x so I dropped those too.

If the tests go green, I'll begin removing the old bits that are in the package files. 

Fixes #386

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,39 +2,12 @@
 language: python
 sudo: false
 cache: pip
-matrix:
-  exclude:
-    - python: "3.3"
-      env: DJANGO=">=1.9,<1.10" DRF=">=3.3,<3.4"
-    - python: "3.3"
-      env: DJANGO=">=1.9,<1.10" DRF=">=3.4,<3.5"
-    - python: "3.3"
-      env: DJANGO=">=1.10,<1.11" DRF=">=3.4,<3.5"
-    - python: "3.3"
-      env: DJANGO=">=1.11,<1.12" DRF=">=3.4,<3.5"
-    - python: "3.3"
-      env: DJANGO=">=1.11,<1.12" DRF=">=3.5,<3.6"
-    - python: "3.3"
-      env: DJANGO=">=1.11,<1.12" DRF=">=3.6,<3.7"
 python:
   - "2.7"
-  - "3.3"
   - "3.4"
   - "3.5"
   - "3.6"
 env:
-  - DJANGO=">=1.8,<1.9" DRF=">=3.1,<3.2"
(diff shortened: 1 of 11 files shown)


mblayman:
I'll leave this open for comment for a week or two. If people are cool with it, we can merge it and move into the future! 😄

sliverc:
Great... I like it when this gets cleaned up.
Technically Django 1.8 is still supported till April next year - I do not mind to drop it others might comment on that though (they could still use old versions though)

There are several spots in the code where django.VERSION is checked... I think some of those checks could be removed now.

mblayman:
Right about Django 1.8. I figured DJA 2.3.1 can be the end of the line for 1.8.

This PR clears out *all* the `django.VERSION` checks. 😄

sliverc:
Nice :+1: 

It might be good to reference the issue why only version DRF 3.6.3 supports Django 1.11. I haven't found it with a quick search.

mblayman:
That would be nice. I didn't base that off an issue because I couldn't find one. I went source diving and looked at the DRF tags on GitHub and browsed through the tags until I found the first one that mentions 1.11 on the README. 3.6.3 was it.

sliverc:
The PRs seem to be:

https://github.com/encode/django-rest-framework/pull/5081
https://github.com/encode/django-rest-framework/pull/5062

which were only released with 3.6.3. So this looks good.

mblayman:
Awesome. I appreciate you checking on that. It's nice to have confirmation of when the support for 1.11 was added.

amw:
I fully approve and eagerly await release. For some reason an older release of DRF JSON API that I have been using works for me after upgrading to Django 2.0, but I feel uncomfortable about it.

mblayman:
I open this 10 day ago and have only seen positive feedback so I'm going to go ahead and merge it.

PR #392, opened 2017-12-04
Fix NoReverseMatch import for Django >= 2.0

lucacorti:
This is the same as #385 but isort is happy.


mblayman:
This is resolved by the recently merged branch that removes support for much older Django versions.

Issue #367, opened 2017-07-25
Add bulk operation support


zivSher:
Hi guys,
I'm trying to post an array of objects. (it's basically a list of dictionaries).
I'm getting 
'if data.get('type') != resource_name and request.method in ('PUT', 'POST', 'PATCH'):
AttributeError: 'list' object has no attribute 'get''

from the parsers.py.
Seems like it doesn't know to handle a case where the data is a list.
Is that functionality missing or am I doing something wrong?

I'm using 'many=true' in the serializer. 

This is the data I'm sending:
{
                "data": [
                    {
                        "type": "admin_mappings",
                        "attributes": {
                            "oxygen_id": "2222",
                            "type": "1",
                            "my_role": "manager"
                        },
                        "relationships": {
                            "container": {
                                "data": {
                                    "type": "containers",
                                    "id": "5d0554d1-22b3-441c-89f5-d6d5cdd74451"
                                }
                            }
                        }
                    },
                    {
                        "type": "admin_mappings",
                        "attributes": {
                            "oxygen_id": "3333",
                            "type": "1",
                            "my_role": "manager"
                        },
                        "relationships": {
                            "container": {
                                "data": {
                                    "type": "containers",
                                    "id": "5d0554d1-22b3-441c-89f5-d6d5cdd74451"
                                }
                            }
                        }
                    }]
}

Thanks!

sliverc:
What you suggest is adding bulk support for json api. This is actually not part of json api standard but might still be valid to add. Best have a look at https://github.com/django-json-api/django-rest-framework-json-api/pull/393#issuecomment-353579753 for more details.

donaldwasserman:
So this is actually something being worked on right now: https://github.com/json-api/json-api/pull/1254

I'd actually be interested in working on this, but it's not to a point yet where I've had enough time to see where it's at (the PR to the spec is still open, but I think it's mostly complete)

puremath:
Could someone tell me what that state of this issue is and what might be the most promising path from here on? As I see it this is where we stand:

* The discussed [sideposting draft](https://github.com/json-api/json-api/pull/1197) did not make it into JSON-API [1.1RC](https://jsonapi.org/format/1.1) and seems pretty abandoned.

* There is a chance that [operations](https://github.com/json-api/json-api/pull/1254) will be included in a 1.2RC some months from now, but there is no guarantee for this to actually happen nor can it be considered a stable specification at this point.

[...]

PR #393, opened 2017-12-13
Overall improvements to parser.

amw:
This PR is something that I've mentioned long time ago in #290. It adds support for sending list data in views other than `RelationshipView` allowing users to create bulk operation views (Bulk operations are one of JSON API extensions). But it also fixes a few "500 Server Error" instances where parser tried to use `data` like a dictionary even when it wasn't one.

Related issue by another user: #367 

- Less exceptions caused by data type assumptions
- Better handling of list data
- Does not special-case RelationshipView
- Support `meta` in Resource Identifier Objects (per spec)

### DIFF
--- a/rest_framework_json_api/parsers.py
+++ b/rest_framework_json_api/parsers.py
@@ -6,7 +6,8 @@ from django.utils import six
 from rest_framework import parsers
 from rest_framework.exceptions import ParseError
 
-from . import exceptions, renderers, serializers, utils
+from . import exceptions, renderers, utils
+from .serializers import PolymorphicModelSerializer, ResourceIdentifierObjectSerializer
 
 
 class JSONParser(parsers.JSONParser):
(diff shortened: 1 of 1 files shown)


mblayman:
@amw, thanks, this looks pretty cool. If I'm reading your description correctly, this is adding new behavior to the JSONParser. Do you think some new tests could be included to hit some of the extra lines in the parser? Looking at the codecov report, lines 127 and 139 do not appear to be covered.

amw:
Sure, I'll add some tests. Just wanted to wait for initial feedback. I might have a bit of time during the holiday break.

sliverc:
Thanks @amw for bringing this PR forward. I have also struggled with bulk operation support with json:api for a while and have been following the different discussions.

What you are suggesting is to implement the experimental bulk extension as described [here](https://github.com/json-api/json-api/blob/v1.0/extensions/bulk/index.md)

This seems to have been [deprecated](http://jsonapi.org/extensions/) and json:api is properly moving to a completely different approach in version 1.1 as discussed in https://github.com/json-api/json-api/pull/1197

I think though it is still legit to implement bulk extension in DJA as DJA currently targets json:api version "1.0". I think though as it is an experimental feature and media type is different ('application/vnd.api+json; ext=bulk') bulk support logic should be separated into its own parser. Reason being as the standard doesn't allow list primary data when [creating](http://jsonapi.org/format/#crud-creating) or [updating](http://jsonapi.org/format/#crud-updating) resources so the jsonapi parser shouldn't allow it either.

Instead a bulk parser could be designed like this:

```python
import io
import json

from django.conf import settings
from rest_framework_json_api import parsers

class JSONAPIBulkParser(parsers.JSONParser):
    media_type = 'application/vnd.api+json; ext=bulk'

    def parse(self, stream, media_type=None, parser_context=None):
        data = []
        encoding = parser_context.get('encoding', settings.DEFAULT_CHARSET)
        data_collection = json.loads(stream.read().decode(encoding))

        for single_data in data_collection['data']:
            single_data = {'data': single_data}
            sub_data = super(JSONAPIBulkParser, self).parse(
                io.BytesIO(bytes(json.dumps(single_data), encoding)),
                media_type=media_type,
                parser_context=parser_context
            )
            data.append(sub_data)

        return data
```

For this parser to work a view mixin would be needed which could look something like this:

```python
class BulkExtensionMixin(object):
     def get_serializer(self, *args, **kwargs):
        is_bulk = isinstance(self.request.data, list)
        kwargs['many'] = kwargs.pop('many', is_bulk)
        return super().get_serializer(*args, **kwargs)
```

This way if a user of DJA wants to use the experimental bulk feature it would need to add new parser to `DEFAULT_PARSER_CLASSES` and extend views with mixin which actually need bulk support. In the respective class it could be clearly documented that this is an experimental feature.

What do you think?

sliverc:
@amw 
There hasn't been any progress on this since quite a while so I am closing this PR. If you still want to work on this please comment and I can reopen. Or if someone else wants to work on this change feel free to open a new PR.

PR #394, opened 2017-12-18
Add DRF 3.7 and Django 2.0 to the test matrix.

### LINKED ISSUE #388: Add support for Django REST Framework 3.7 and Django 2.0
DJA should be tested against DRF 3.7.x and document that support.

mblayman:
Fixes #388

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,6 @@ matrix:
     - python: 3.4
       env: DJANGO=">=1.11,<2.0" DRF=">=3.7.0,<3.8"
     - python: 3.4
-      env: DJANGO=">=2.0,<2.1" DRF=">=3.6.3,<3.7"
-    - python: 3.4
       env: DJANGO=">=2.0,<2.1" DRF=">=3.7.0,<3.8"
 
     - python: 3.5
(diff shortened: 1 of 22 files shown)


mblayman:
The DJA test suite relies on the example app which, in turn, relies on django-polymorphic. There is an open issue with django-polymorphic (https://github.com/django-polymorphic/django-polymorphic/issues/324) that prevents it from running on Django 2.0.

Once django-polymorphic works on Django 2.0, I can come back to this branch and (hopefully), merge it.

sliverc:
Maybe we should think about making the dependency django-polymorphic optional as the majority of  users are properly not using it. So if someone wants to use it they already have the dependency set in their own project to create the models and DJA will detect it that it is installed.

This way DJA wouldn't be blocked by issues like this to upgrade to a newer Django and DRF version. The tests in this PR would simply need to exclude polymorphic tests to be run on Django 2.0 till django-polymorphic has fixed it.

mblayman:
I considered that route, but ultimately decided against it because I thought it would be too confusing.

If DJA released a version compatible with Django 2.0, but didn't work with django-polymorphic because polymorphic still doesn't work with 2.0, then I feel like the documentation would be lying to users. I'm not comfortable stating DJA supports Django 2.0 and django-polymorphic at the same time.

In addition (and probably the bigger problem), the example app uses a polymorphic model, and I don't know how to conditionally make models in the example app in a way that won't cause migrations oddities. One part of this project that I find puzzling is that the entire test suite is extremely coupled to the example project. I understand that it's for historical reasons, but I think that is the blocker.

I think that the 2.0 branch for polymorphic is blocked (https://github.com/django-polymorphic/django-polymorphic/pull/318) so maybe the best thing we could do is go help them get unblocked. 😄

sliverc:
Fair enough. Not sure whether I will find the time myself but it would properly be a good idea to assist django-polymorphic with 2.0 support.

mblayman:
Cool, it looks like there was progress on the django-polymorphic issue. I'll keep my eyes open for a new release that has support for Django 2.0.

mblayman:
Excellent! Many folks are looking for a Django 2.0 compatible release so I'm going to merge this and make a release in the next couple of days.

dbinetti:
Hey @mblayman thanks for your efforts here.  Any chance you could cut a release?

dbinetti:
Ooops! I checked 'latest' but didn't look deeply enough.  Apologies @mblayman  for the interruption....  deleting comment.

PR #396, opened 2018-01-05
Allow json api renderer to be used as test request renderer class

sliverc:
This allows to define json api format as default in tests. Following needs to be configured:

```python
REST_FRAMEWORK {
   ...
    'TEST_REQUEST_RENDERER_CLASSES': (
    ),
    'TEST_REQUEST_DEFAULT_FORMAT': 'vnd.api+json'
   ...
}
```

With this change tests can be simplified and there is no need to specifically assign json api content type and dump resp. load data as json. This PR changes tests accordingly.

More detail documentation on this rest framework feature can be found [here](http://www.django-rest-framework.org/api-guide/testing/#configuration)

### DIFF
--- a/example/tests/utils.py
+++ b/example/tests/utils.py
@@ -1,21 +1,7 @@
 import json
 
-from django.utils.encoding import force_bytes, force_text
+from django.utils.encoding import force_text
 
 
 def load_json(data):
     return json.loads(force_text(data))
-
-
-def dump_json(data):
-    '''
-    Converts a Python object to a JSON formatted string.
-    '''
-
-    json_kwargs = {
-        'sort_keys': True,
-        'indent': 4,
-        'separators': (', ', ': ')
-    }
-
-    return force_bytes(json.dumps(data, **json_kwargs))
--- a/example/tests/utils.py
+++ b/example/tests/utils.py
@@ -1,7 +1,0 @@
-import json
-
-from django.utils.encoding import force_text
-
-
-def load_json(data):
-    return json.loads(force_text(data))
--- a/rest_framework_json_api/utils.py
+++ b/rest_framework_json_api/utils.py
@@ -40,7 +40,7 @@ def get_resource_name(context, expand_polymorphic_types=False):
 
     # Sanity check to make sure we have a view.
     if not view:
-        raise APIException(_('Could not find view.'))
+        return None
 
     # Check to see if there is a status code and return early
     # with the resource_name value of `errors`.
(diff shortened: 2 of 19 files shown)


rest_framework_json_api/utils.py
```diff
@@ -40,7 +40,7 @@ def get_resource_name(context, expand_polymorphic_types=False):
 
     # Sanity check to make sure we have a view.
     if not view:
-        raise APIException(_('Could not find view.'))
```
mblayman: Why was this switched to returning None instead of raising an exception?

example/tests/utils.py
```diff
@@ -1,21 +1,7 @@
 import json
 
-from django.utils.encoding import force_bytes, force_text
+from django.utils.encoding import force_text
 
 
 def load_json(data):
```
mblayman: Many of the code changes removed `load_json`. Is it still used somewhere or could it be deleted too?

sliverc: I would like to remove it as well. Unfortunalty there is one [test](https://github.com/django-json-api/django-rest-framework-json-api/blob/master/example/tests/integration/test_non_paginated_responses.py#L117) which still needs it.
Reason being is that it uses `RequestFactory` and not `APIClient`. In such a case response doesn't seem to provide a `.json()` function.

rest_framework_json_api/utils.py
sliverc: DRF uses the test renderer to parse data passed on to `APIClient`. In such a case it won't have a renderer context and therefore there is no view provided.
The json api renderer should still not fail in such a case and render data as is. Hence I had to remove this exception.
Interesting enough is though that the json api renderer was already aware of `resource_name` with value `None`.

example/tests/utils.py
mblayman: Would it be possible to convert the test to use `APIClient` instead?

rest_framework_json_api/utils.py
mblayman: ok, if I understand this correctly based on what you wrote, missing a view should not be possible in a non-test setting? Does that sound true?

sliverc: yep this is true...

example/tests/utils.py
sliverc: Possibly. I have a look.

sliverc: I have found an ok way to convert it to `APIClient`. Now also removed `example/tests/utils.py`.

PR #398, opened 2018-01-08
Fixed import of NoReverseMatch exception

mrhanky17:
In relations.py the NoReverseMatch exception isn't correctly imported for Django versions >= 1.10.

### DIFF
--- a/rest_framework_json_api/relations.py
+++ b/rest_framework_json_api/relations.py
@@ -4,8 +4,8 @@ from collections import OrderedDict
 
 import inflection
 import six
+import django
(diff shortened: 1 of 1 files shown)


mblayman:
@mrhanky17, while I appreciate your contribution, I'm going to close this PR. It's not because your fix is invalid. The reason is that the next version of DJA is dropping support for Django < 1.11. A fix is already applied in #394, but it has not been merged yet because we are waiting for a package dependency update to support Django 2.0.

Thanks for your efforts anyway.

PR #408, opened 2018-02-01
Remove factory-boy version constraint.

mblayman:
PR #407 is failing CI currently because of a package version issue and the PR is only changing a typo in the usage file. Some dependency requirement must have changed recently.

I'm trying this PR to see if it's now possible to have factory-boy working on a newer version.


mblayman:
This makes the test suite green again so I'm going to go ahead and merge it.

PR #409, opened 2018-02-04
Defer to lookup_field/custom id fields

ola-t:
Thsi PR allow support for custom pk https://github.com/django-json-api/django-rest-framework-json-api/issues/155

There are so many places where PK is assumed in
djangorestframework-jsonapi and so many places/ways to change IDs/URLs
in DRF itself, that getting coverage on all the edge cases will be quite
a lot of work.  Some of that work may be better spent re-factoring
things to make the code easier to understand and reduce the number of
edge cases in the first place.

### DIFF
--- a/rest_framework_json_api/renderers.py
+++ b/rest_framework_json_api/renderers.py
@@ -112,7 +112,7 @@ class JSONRenderer(renderers.JSONRenderer):
                         OrderedDict([
                             ('type', relation_type),
                             ('id', encoding.force_text(getattr(
-                                related_object.pk, field.lookup_field)))
+                                related_object, field.lookup_field)))
                         ])
                     )
 
@@ -223,14 +223,14 @@ class JSONRenderer(renderers.JSONRenderer):
                     )
 
                     if hasattr(field.child_relation, 'lookup_field'):
-                        _id = getattr(nested_resource_instance,
-                                      field.child_relation.lookup_field)
+                        pk = getattr(nested_resource_instance,
+                                     field.child_relation.lookup_field)
                     else:
-                        _id = nested_resource_instance.pk
+                        pk = nested_resource_instance.pk
 
                     relation_data.append(OrderedDict([
                         ('type', nested_resource_instance_type),
-                        ('id', encoding.force_text(_id))
+                        ('id', encoding.force_text(pk))
                     ]))
                 data.update({
                     field_name: {
--- a/rest_framework_json_api/renderers.py
+++ b/rest_framework_json_api/renderers.py
@@ -111,7 +111,8 @@ class JSONRenderer(renderers.JSONRenderer):
                     relation_data.append(
                         OrderedDict([
                             ('type', relation_type),
-                            ('id', encoding.force_text(related_object.pk))
+                            ('id', encoding.force_text(getattr(
+                                related_object.pk, field.lookup_field)))
                         ])
                     )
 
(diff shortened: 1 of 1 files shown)


rest_framework_json_api/renderers.py
```diff
@@ -111,7 +111,8 @@ def extract_relationships(cls, fields, resource, resource_instance):
                     relation_data.append(
                         OrderedDict([
                             ('type', relation_type),
-                            ('id', encoding.force_text(related_object.pk))
+                            ('id', encoding.force_text(getattr(
```
mblayman: What's happening with this `getattr`? If my understanding of `getattr` is correct, then something seems off here. Let's say `fields.lookup_field = 'foo'`. Wouldn't this `getattr` call be equivalent to `related_object.pk.foo`?

rest_framework_json_api/renderers.py
```diff
@@ -244,14 +256,23 @@ def extract_relationships(cls, fields, resource, resource_instance):
...
                         nested_resource_instance_type = (
                             relation_type or
                             utils.get_resource_type_from_instance(nested_resource_instance)
                         )
 
+                        instance_pk_name = (
+                            getattr(field, 'lookup_field', None) or
+                            nested_resource_instance._meta.pk.name)
+                        if instance_pk_name in nested_resource_data:
+                            pk = nested_resource_data[instance_pk_name]
```
mblayman: Could you change this to use a consistent name? Higher in this diff, the variable passed to `force_text` was `_id`. Now it's `pk`. I think we should pick one and stick with it.

rest_framework_json_api/renderers.py
ola-t: something went wrong when cherry-picking, sorted now

SubhamBhattacharjee:
@mblayman Please merge if you are okay with the changes.

ola-t:
any chance to merge this PR please?

rest_framework_json_api/renderers.py
```diff
@@ -216,9 +223,15 @@ def extract_relationships(cls, fields, resource, resource_instance):
                         utils.get_resource_type_from_instance(nested_resource_instance)
                     )
 
+                    if hasattr(field.child_relation, 'lookup_field'):
```
sliverc: couldn't this code block be simplified if we add a `lookup_field` to `ResourceRelatedField` defaulting to `pk`?

rest_framework_json_api/renderers.py
```diff
@@ -150,18 +151,24 @@ def extract_relationships(cls, fields, resource, resource_instance):
...
             ):
-                resolved, relation = utils.get_relation_instance(
-                    resource_instance, '%s_id' % source, field.parent
-                )
-                if not resolved:
-                    continue
-                relation_id = relation if resource.get(field_name) else None
+                lookup_field = getattr(field, 'lookup_field', None)
+                if lookup_field:
+                    relation_id = getattr(
```
sliverc: could the lookup field not also be a relation?

rest_framework_json_api/renderers.py
```diff
@@ -463,9 +485,19 @@ def build_json_resource_obj(cls, fields, resource, resource_instance, resource_n
         # Determine type from the instance if the underlying model is polymorphic
         if force_type_resolution:
             resource_name = utils.get_resource_type_from_instance(resource_instance)
+        if resource_instance is None:
+            pk = None
+        elif 'id' in fields:
+            pk = fields['id'].get_attribute(resource_instance)
+        elif 'url' in fields:
```
sliverc: Instead of 'url' `api_settings.URL_FIELD_NAME` should be used.

sliverc:
@ola-t
There hasn't been any progress on this since quite a while so I am closing this PR. If you still want to work on this please comment and I can reopen. Or if someone else wants to work on this change feel free to open a new PR.

PR #411, opened 2018-02-07
Issue 410

n2ygk:
I was a bit confused by the various sources of urlpatterns, etc. and hopefully got it right. I ran both the example app with runserver and did a tox run too to make sure I didn't break that. 

For whatever reason, I was unable to get this to work by doing `export DJANGO_SETTINGS_MODULE=example.settings`  and various `django-admin.py` stuff as I got an error about module example not found. I'm not quite sure why that happened.

However, doing basically the same thing by creating `manage.py` works and, since `manage.py` is the usual documented approach, I think that's OK.

It's a little different from the usual pattern in that there's not a project directory with settings and then an example app directory; both are the same, but that's OK I think.

There's still one problem in here in that `http://127.0.0.1:8000/entries/2/relationships/suggested` returns a 404. I must have missed something. @mblayman Maybe you can find it quicker than I can.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -70,7 +70,6 @@ From Source
 
     git clone https://github.com/django-json-api/django-rest-framework-json-api.git
     cd django-rest-framework-json-api
-    pip install -e .
     pip install -r example/requirements.txt
     django-admin.py runserver
 
@@ -78,5 +77,5 @@ Browse to http://localhost:8000
 
 ## Running Tests
 
-    python runtests.py
+    tox
 
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -73,6 +73,7 @@ From Source
     python -m venv env
     source env/bin/activate
     pip install -r example/requirements.txt
+	pip install -e .
     django-admin.py startproject example .
     python manage.py migrate
     python manage.py runserver
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -70,12 +70,17 @@ From Source
 
     git clone https://github.com/django-json-api/django-rest-framework-json-api.git
     cd django-rest-framework-json-api
+    python -m venv env
+    source env/bin/activate
     pip install -r example/requirements.txt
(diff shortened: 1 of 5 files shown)


n2ygk:
Ugh, looks like my edit to example/urls.py messed up.

n2ygk:
@mblayman Looks like now it's safe to merge. I had a bit of a learning curve with tox, travis, etc. `manage.py test` still fails though the travis and tox tests don't catch it.

docs/getting-started.md
```diff
@@ -70,13 +70,17 @@ From Source
 
     git clone https://github.com/django-json-api/django-rest-framework-json-api.git
     cd django-rest-framework-json-api
-    pip install -e .
```
mblayman: How does this work? I don't understand where the DJA package would be available to Python if it's not installed in editable mode.

n2ygk: It's available because you are in the source tree for DJA so `import rest_framework_json_api` just works. I suppose retaining `pip install -e .` can't hurt. Pushing a new commit.
```
django-rest-framework-json-api$ tree -L 1
.
├── AUTHORS
├── CHANGELOG.md
├── LICENSE
├── MANIFEST.in
├── README.rst
├── docs
├── drf_example
├── env
├── example
├── manage.py
├── pytest.ini
├── requirements-development.txt
├── requirements.txt
├── rest_framework_json_api
├── setup.cfg
├── setup.py
└── tox.ini
```

n2ygk: Sorry, neglected to answer the question: "Did you happen to run the steps you listed in Getting Started from a totally fresh clone to see if it would start the example app properly?"

Yes. Several times.

mblayman:
Thanks! I appreciate all your work on this branch. I forgot to ask you to add your name to the AUTHORS file. We do that to give you credit for your contribution and recognize that you have some copyright claim to this project. The list is alphabetical by first name in case that's somehow unclear.

PR #412, opened 2018-02-15
Speed up JSONRenderer.extract_included

amw:
I've hit a nasty performance issue with one of my index-type API endpoints. It returns only about 600 elements each with 3 additional resources included (one of them nested under another). The included resources are widely shared (loaded with `prefetch_related`). All database queries for this request taken together execute for less than 10ms, but the entire request takes over 2 seconds to complete. That's not reasonable at all.

Turns out main view code was taking *"only"* about 500ms (still a lot for this amount of data) while `JSONRenderer.render` was taking almost 2 seconds.

In this PR I am introducing some optimizations to `JSONRenderer.extract_included` – instead of serializing all related resources and then deduplicating them near the end, I pass a cache of serialized resources and skip processing when encountering the same resource.

Here are some measurements I've put around the `render` method. Before my changes:
```
Successfully completed rendering in 1.807s
Successfully completed rendering in 1.790s
Successfully completed rendering in 1.788s
Successfully completed rendering in 1.809s
```

After my changes:
```
Successfully completed rendering in 0.465s
Successfully completed rendering in 0.460s
Successfully completed rendering in 0.463s
Successfully completed rendering in 0.463s
```

Just to give the idea of how much time `extract_included` takes for my request and what is the ceiling for our optimizations I've ran another test with `extract_included` short-circuited to return immediately without extracting anything:
```
Successfully completed rendering in 0.054s
Successfully completed rendering in 0.056s
Successfully completed rendering in 0.054s
Successfully completed rendering in 0.054s
```

## Warning

Optimization in this PR is aggressive and there is a theoretical scenario where it could fail to extract some resources. It is related to nested included resources, e.g. `["contributors", "author.opinions"]`. If `author` happened to be first extracted via a different path (like `contributors`) then their related `opinions` would never get extracted.

I do not run this risk, because I don't include the same resource type via multiple paths in a single endpoint. I think this risk is acceptable for this library overall. Including one resource type via multiple paths in a single endpoint is already unsafe operation today – if you were to use different `Serializer` classes for these different paths then only one serialization would be kept for `included` key at the end of rendering. The versions of the same serialized resource would be discarded as long as they have the same `type` and `id` values.

## Caveat

I have focused on optimizing extraction via a simple `ResourceRelatedField`. There are more optimizations that can be made for `ListSerializer` and the `included_cache` dictionary that I have introduced paves the way for others to implement them.

### DIFF
--- a/rest_framework_json_api/exceptions.py
+++ b/rest_framework_json_api/exceptions.py
@@ -2,12 +2,13 @@ from django.conf import settings
 from django.utils.translation import ugettext_lazy as _
 from rest_framework import exceptions, status
 
-from rest_framework_json_api import renderers, utils
+from rest_framework_json_api import utils
 
 
 def rendered_with_json_api(view):
+    from rest_framework_json_api.renderers import JSONRenderer
     for renderer_class in getattr(view, 'renderer_classes', []):
-        if issubclass(renderer_class, renderers.JSONRenderer):
+        if issubclass(renderer_class, JSONRenderer):
             return True
     return False
 
--- a/rest_framework_json_api/renderers.py
+++ b/rest_framework_json_api/renderers.py
@@ -2,7 +2,7 @@
 Renderers
 """
 import copy
-from collections import OrderedDict
+from collections import OrderedDict, defaultdict
 
 import inflection
 from django.db.models import Manager
@@ -13,6 +13,7 @@ from rest_framework.settings import api_settings
 
 import rest_framework_json_api
 from rest_framework_json_api import utils
+from rest_framework_json_api.relations import ResourceRelatedField
 
 
 class JSONRenderer(renderers.JSONRenderer):
@@ -313,12 +314,12 @@ class JSONRenderer(renderers.JSONRenderer):
         return relation_instance
 
     @classmethod
-    def extract_included(cls, fields, resource, resource_instance, included_resources):
+    def extract_included(cls, fields, resource, resource_instance, included_resources,
+                         included_cache):
         # this function may be called with an empty record (example: Browsable Interface)
         if not resource_instance:
             return
 
-        included_data = list()
         current_serializer = fields.serializer
         context = current_serializer.context
         included_serializers = utils.get_included_serializers(current_serializer)
@@ -350,9 +351,6 @@ class JSONRenderer(renderers.JSONRenderer):
             if isinstance(relation_instance, Manager):
                 relation_instance = relation_instance.all()
 
-            new_included_resources = [key.replace('%s.' % field_name, '', 1)
-                                      for key in included_resources
-                                      if field_name == key.split('.')[0]]
             serializer_data = resource.get(field_name)
 
             if isinstance(field, relations.ManyRelatedField):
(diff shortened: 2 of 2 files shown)


mblayman:
@amw I'm definitely open to this fix. I think the warning you described sounds like acceptable risk to me. Looks like there is an isort error in the CI build (I think it's an ASCII sorting issue thing). I'm happy to review once Travis is green.

amw:
I've sent a fix for `isort`, but Travis is still complaining – I don't see about what though – all the logs seem empty.

mblayman:
Looks like you fixed isort. Now you have a flake8 error.
```
$ flake8
./rest_framework_json_api/renderers.py:318:101: E501 line too long (103 > 100 characters)
./rest_framework_json_api/renderers.py:401:33: E126 continuation line over-indented for hanging indent
./rest_framework_json_api/renderers.py:406:29: E121 continuation line under-indented for hanging indent
./rest_framework_json_api/renderers.py:424:29: E126 continuation line over-indented for hanging indent
The command "flake8" exited with 1.
```

amw:
All green.

rest_framework_json_api/renderers.py
```diff
@@ -610,20 +614,11 @@ def render(self, data, accepted_media_type=None, renderer_context=None):
...
-                    seen.add(type_tuple)
-                    unique_compound_documents.append(included_dict)
-
-            # Sort the items by type then by id
-            render_data['included'] = sorted(
-                unique_compound_documents, key=lambda item: (item['type'], item['id'])
-            )
+        if included_cache:
+            render_data['included'] = list()
+            for included_type in sorted(included_cache.keys()):
```
mblayman: Calling `.keys()` on this line and the next is not necessary. `sorted` applies to the keys of a dictionary by default.

mblayman: Alternatively, this could be compacted with a nested list comprehension:
```python
render_data['included'] = [
    instance for instance_type, type_values in sorted(instance_cache.items())
             for id, instance in sorted(type_values.items())]
```
This works because sorted on `items` works on the keys. This version eliminates a bunch of append operations. There's probably a small perf gain here, but I bet it's tiny. Here's an example to illustrate:
```python
>>> instance_cache = {'c': {'b': 6, 'a': 5}, 'a': {'b': 2, 'a': 1}, 'b': {'b': 4, 'a': 3}}
>>> [instance for instance_type, type_values in sorted(instance_cache.items()) for id, instance in sorted(type_values.items())]
[1, 2, 3, 4, 5, 6]
```

rest_framework_json_api/exceptions.py
```diff
@@ -2,12 +2,13 @@
 from django.utils.translation import ugettext_lazy as _
 from rest_framework import exceptions, status
 
-from rest_framework_json_api import renderers, utils
+from rest_framework_json_api import utils
 
 
 def rendered_with_json_api(view):
+    from .renderers import JSONRenderer
```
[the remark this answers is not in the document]
amw: Done

rest_framework_json_api/renderers.py
```diff
@@ -313,12 +315,12 @@ def extract_relation_instance(cls, field_name, field, resource_instance, seriali
         return relation_instance
 
     @classmethod
-    def extract_included(cls, fields, resource, resource_instance, included_resources):
+    def extract_included(cls, included_cache, fields, resource, resource_instance,
```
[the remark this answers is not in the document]
amw: My thinking was that `included_cache` is the target of the operation, almost a subject like `self` and I usually place these first. But I don't care – changed as requested.

rest_framework_json_api/renderers.py
amw: If you don't mind I prefer explicitness in code. Most people know that iterating a dictionary in Python yields keys, but using `keys()` makes it obvious for everyone – seniors and juniors from whatever programming language background.

I feel similarly about that nested list comprehension. I didn't know that a thing like that existed in Python, but I have to say it is very difficult to read. Seems arcane.

mblayman: I can appreciate your point of view here. Thanks for at least considering it.

PR #419, opened 2018-04-16
Add json api settings module

sliverc:
This is inspired by api_settings of REST framework.

Advantages:
* Centralizing all json api settings and defaults in one module
* Simplifying refactoring and deprecation of settings in the future
* Properly use override_settings in tests instead of monkey patching


sliverc:
@n2ygk Could you have a look at this? Thanks.

sliverc:
Thanks for review. Merging it therefore. Yep, we could certainly think about moving to a dict approach as DRF does. One step at the time... ;).

Issue #230, opened 2016-04-21
`write_only` field included in response


martinmaillard:
Declaring a `write_only` on a serializer should not change the response. Currently, `write_only` fields are included in the response with a `null` value.

PR #421, opened 2018-04-17
Don't output key for write only relations

peakwinter:
If the `write_only` key is `True` on a relationship, its key is still included in the response body with an empty resource object (`{'data': null}`).

This was fixed in #230 for normal fields but not for relations.


sliverc:
Thanks for your contribution. Looks good to me but could you write a test case for this? Thanks.

Issue #424, opened 2018-04-24
Request for new maintainer


mblayman:
I picked up the maintainer torch about a year ago when the project seemed to be stuck. I was using DJA on a side project, and I needed to fix a bug and get a new release out. Today, I'm no longer working on the side project and my interest in maintaining DJA has waned significantly.

As I'm not being a great steward for the project, I'd like to request that someone else pick it up. This is very much a community effort and code contributions come from a wide variety of contributors. The new maintainer would be most involved with:
1. Review of new Pull Requests and Issues.
2. Deploying new releases to PyPI.

I hope that I've added the right tools that can make management of this project easier for a future maintainer. Static code analysis has removed a lot of friction that might come up as a primary PR reviewer.

If you're in a decent place to step into the maintainer role, please let me know and I can work to transition over project handling to you. I'm happy to help with a bit of training if needed or improving release documentation.

mblayman:
Yes, I think I'm the only active maintainer. My chronology may not be exactly right, but @gaker started the project, @jsenecal maintained it after that, @jerel followed him, and I picked it up from @jerel.

These guys chime in from time to time, but I don't believe that any of them are actively maintaining DJA.

Switching to Jazzband is probably a pretty solid idea if there is no one with the time availability to act as a regular maintainer.

sliverc:
I use DJA in several projects and therefore would be interested in a well maintained DJA. I do not think though I would have the time to be a sole active maintainer and I also think it won't be good for the project in the long run.

For the future of the project moving it to JazzBand could be a good idea. If it is decided to go down this way I would be willing to help out in the transition.

n2ygk:
I would be happy to back you up @sliverc. I just joined Jazzband, FWIW.

mblayman:
I'm happy to pass the baton to the two of you as you both have actively contributed within the last year. I think I would need to add you to the django-json-api team. Ultimately, I can give one or both of you the upload permission to the PyPI package.

I think I should probably start a `CONTRIBUTING.md` file so that if the team ultimately decides to move to Jazzband, that work will at least be done.

sliverc:
@mblayman If you could add me to the team and give PyPi upload rights that would be great.
I see when I get around but I will try to get the current PRs reviewed and get a release out.

[...]

PR #425, opened 2018-04-26
Add a CONTRIBUTING.md file.

mblayman:
This branch adds a skeleton of contribution guidelines and cleans up
little bits that I noticed along the way.

For #424

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -13,6 +13,7 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
+import datetime
 import sys
 import os
 import shlex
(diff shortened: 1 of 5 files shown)


mblayman:
@sliverc, @n2ygk, you both might want to look at this and let me know what you think I should add. I'm not sure what bits of knowledge about this project are locked away in my brain that I need to let out. 😄

docs/conf.py
```diff
@@ -52,8 +53,9 @@
 
 # General information about the project.
 project = 'Django REST Framework JSON API'
-copyright = '2015, Jerel Unruh and contributors'
-author = 'Jerel Unruh'
+year = datetime.date.today().year
```
n2ygk: @mblayman Sorry for this late response. Shouldn't the copyright date not just be the current year but a range or list of years starting from when the code was first released? Or just leave it at the date of first publication?  There are [lots of opinions](https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated) around this but it seems that by advancing the copyright notice automatically you are

docs/conf.py
```diff
@@ -52,8 +53,9 @@
...
+copyright = '{}, Django REST Framework JSON API contributors'.format(year)
```
n2ygk: Also, I would think @jerel's name should be retained with special emphasis as the original creator. "and contributors" for the rest...

n2ygk: Oh wait, git says @gaker for the initial commit. Hmm... I'm still uncomfortable with removing a specific name from a copyright statement.

PR #434, opened 2018-05-05
Issue 430: pagination enhancement

### LINKED ISSUE #430: spec compliance: pagination
Split #416 out into pagination improvements which comply with the jsonapi 1.0 [spec](http://jsonapi.org/format/#fetching-pagination)

n2ygk:
Fixes #430

## Description of the Change
Adds configurable pagination query parameter names along with settings defaults for them.
Note that changing the defaults of `page` and `page_size` would be a breaking change so
those are left as-is but documentation shows how to change them to `page[number]` and
`page[size]` for example.

## Checklist

- [x]  PR only contains one change (considered splitting up PR)
- [x] unit-test added (was already present)
- [x]  documentation updated

### DIFF
--- a/example/tests/unit/test_pagination.py
+++ b/example/tests/unit/test_pagination.py
@@ -87,7 +87,7 @@ class TestLimitOffset:
 
 
 # TODO: This test fails under py27 but it's not clear why so just leave it out for now.
-@pytest.mark.xfail((sys.version_info.major, sys.version_info.minor) == (2,7),
+@pytest.mark.xfail((sys.version_info.major, sys.version_info.minor) == (2, 7),
                    reason="python2.7 fails for unknown reason")
 class TestPageNumber:
     """


sliverc:
I think we should try to avoid adding new settings which DRF doesn't provide but simply add an example how to overwrite the class in the documentation (as you actually already do in this PR).

See http://www.django-rest-framework.org/api-guide/pagination/#modifying-the-pagination-style

Question is then how we want to solve problem how to move to new query param names..

What about adding a new class `JsonApiPageNumberPagination` and the like with the proper parameter names?

We could then deprecate the other classes `PageNumberPagination` and `LimitOffsetPagination`. To solve deprecation user can the move to JsonApi pagination classes or overwrite it to set parameters as needed.

What do you think?

In terms of Travis problem as workaround: You can simply push to git again and travis should pick it up.

e.g.
```bash
git commit --amend -m "Change to a slight different commit message"
git push -f
```

n2ygk:
Sounds like a plan. I’ll take a look in the next couple of days.

I had pushed another commit to try to wake up Travis....

n2ygk:
@sliverc I think the workflow changes that were implemented are preventing Travis from running. It almost looks like it won't submit the Travis job until after an approving review.

mblayman:
@n2ygk what workflow changes are you referring to? There shouldn't be anything that would block Travis from running on PR branches. Travis normally runs everything before a review even happens.

I'm not sure why your PR is not running. If it would help us debug, I can try to push up a dummy PR to see if CI will run my branch. That might helps us determine if the problem is for the whole project or something with your specific repo or branch.

n2ygk:
Yes please do. I did request a review this time. I don’t recall if I’ve
done that before.

sliverc:
@n2ygk 
Changes look good.

In terms of deprecation this could be done by adding a warning to `__init__` of both classes like:
```python
warnings.warn('PageNumberPagination is deprecated. Use JsonApiPageNumberPagination instead or create custom pagination (link to documentation)', DeprecationWarning)
```

Maybe this way we don't have to document deprecation in the documentation or this is simply additional. What do you think?

n2ygk:
@sliverc Do you mean linking to the usage.md in the git repo or at readthedocs.io? The latter won't be there until release... Also the former won't be at the right place until merged.

sliverc:
@n2ygk Fair point you would need to derive classes the other way around. Maybe a link to where it is gonna be linked to once it is released on readthedocs.io what do you think?

n2ygk:
I guess I'm using something newer than available in [Python-2.7](https://travis-ci.org/django-json-api/django-rest-framework-json-api/builds/379834096?utm_source=github_status&utm_medium=notification)

example/tests/unit/test_pagination.py
```diff
@@ -76,3 +77,22 @@ def test_valid_offset_limit(self):
...
         assert content == expected_content
+
+    def test_LimitOffset_deprecation(self):
+        with pytest.warns(DeprecationWarning) as record:
+            pagination.LimitOffsetPagination()
+        assert len(record) == 1
+        assert 'LimitOffsetPagination' in str(record[0].message)
+
+
+# TODO: This test fails under py27 but it's not clear why so just leave it out for now.
```
sliverc: There is a pytest feature called xfail. See https://docs.pytest.org/en/latest/skipping.html

So instead of commenting this out you can mark it as `@pytest.mark.xfail` and still leave you comment with `TODO`.

This is that `CI` doesn't fail even when test is not successful but it still counts towards coverage.

example/tests/unit/test_pagination.py
```diff
@@ -76,3 +77,22 @@ def test_valid_offset_limit(self):
 
         assert queryset == list(range(offset + 1, next_offset + 1))
         assert content == expected_content
+
+    def test_LimitOffset_deprecation(self):
```
sliverc: Methods name should always be all snake cased and lowercase.

example/tests/unit/test_pagination.py
n2ygk: done. Got caught by flake8 again (#436 will address this if and when done)

PR #440, opened 2018-06-12
Use tox to run test in travis

### LINKED ISSUE #391: Test with tox in CI
2.3.0 is currently broken unless a user's environment happens to have `six` installed. If the CI tests ran at least one test that minimally installs the package and runs some basic tests to import things, then that kind of problem would be avoided.

sliverc:
This way we can harmonize tox and travis configuration

Fixes #391

## Checklist

- [ ] unit-test added (not needed as only build change)
- [ ] documentation updated (not needed as only build change)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,9 @@ pip-delete-this-directory.txt
 # PyTest cache
 .pytest_cache/
 
+# Coverage
+.coverage
+
 # Tox
 .tox/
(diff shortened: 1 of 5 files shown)


sliverc:
@mblayman I would have been too lucky to get it right in the first run. Coverage was not working anymore and is now fixed. Could you have a look again? Thanks.

Issue #296, opened 2016-11-15
Include only a link in the relationships object


arielpontes:
According to the JSONAPI specification, if I have an `Article` model with a to-many relationship to a `Comment` model and I want to express this when I retrieve an `Article` resource, there are multiple valid ways I can do this:

1. Include the comments by default in the response, having a list of resource identifier objects in the `relationships` object and a list with the actual resource objects in the `included` object.
2. Return a list of resource identifier objects in the `relationships` object only, without including the actual attributes of the comments.
3. Include only a link (e.g. `/articles/1/comments`) in the `relationships` object.

I have a model that links to a lot of related objects so the 3rd option would be the best for me, but I can't configure the ResourceRelatedField to act like that. It always returns a `data` object with the full list of related resource identifiers. Is there a way I can avoid this? Or do I have to extend the class myself? Doesn't seem very trivial to do that in a clean way.

Ideally, since the JSONAPI specs accept these 3 options, I think this library should allow these 3 configurations out of the box.

benjaminlong:
Hello @arielpontes, have you managed to remove the `data` attribute when trying to use ResourceRelatedField ? I would like to have the same behavior.

benjaminlong:
Hello ! if the community thinks is the good improvement, I will have some time to work on this in the next month. Let me know if there are any first thoughts/ideas. 
- Should we add a `link_only` option ?
- Should we...
  - return the link, when `related_link_view_name` and `related_link_lookup_field` options  of the `ResourceRelatedFiled` are presents. 
  - return the  data if those optiosn aren't presents.
- A third option ?

All the best.

sliverc:
There is a `relations.HyperlinkedRelatedField` do you think this does what you expect?

Anton-Shutik:
@sliverc I didn't find a way I can use `relations.HyperlinkedRelatedField` and get `links` only in the relationship payload. May be I doing smth wrong ? Anyway, I think that impossible to do via DRF serializers or fields, because I've walked through `extract_relationships` and there is no way to get `data` key omitted and have `links` only

@arielpontes I've made a [PR](https://github.com/django-json-api/django-rest-framework-json-api/pull/443) that skips the "data", I think that is what you need to get your #3. Hope that helps you

sliverc:
[...]

Issue #178, opened 2015-12-23
Pagination of nested relationships


maryokhin:
Hello, recently started playing with this package, and while I am a bit struggling with somewhat scarce documentation, I really love the idea and how it works. 

One of the quirks I have is how to paginate nested relationships. My data can easily reach thousands of relations per object and that sometimes just hangs the server/browser.

I read the spec and [found out](http://jsonapi.org/format/#document-resource-object-relationships) that it is possible to have the server paginate also the internal relationships. 

> A relationship object that represents a to-many relationship MAY also contain pagination links under the links member.

I would guess that this is not implemented yet? If not, can you evaluate on how hard it would be to do something like this and maybe point me in the right direction on where to start?

maryokhin:
Here is what I found regarding implementation: http://stackoverflow.com/questions/15617595/paginate-relationship-in-django-rest-framework.

jerel:
Based on that stackoverflow the first thing I would suggest trying is `SerializerMethodResourceRelatedField`. It allows you to declare a method on the serializer who's data will show up in the `relationships` json. I don't believe it currently works to return a queryset of many items (pull request #151 is meant to address that) so that PR might need to be cleaned up and merged first.

maryokhin:
You mean an in-house implementation? I was thinking of maybe a generic implementation for the package. What I had in mind was:
* Having settings like `DEFAULT_RELATIONSHIP_PAGINATION_CLASS`, `RELATIONSHIP_PAGE_SIZE` etc. that are turned off by default.
* If setting is set, then paginate the relationships under the hood, serialize the links and so on.

So I was think how hard it would be to achieve something like this and how you look at that type of functionality being in the package. It just looks like functionality that anyone who has more than 20 relations per object would probably want.

jerel:
I meant in-house implementation in case you needed it right now. We would like to see a proper solution in this package so if you feel to contribute it would be most welcome. @jsenecal and I had discussed the setting approach and if I remember right we decided it would work well but haven't had time to implement yet.

It seems to me like pagination styles should be uniform across the API so perhaps the main pagination class (`rest_framework_json_api.pagination.PageNumberPagination`) could be used by default with `DEFAULT_RELATIONSHIP_PAGINATION_CLASS` being optional? I haven't thought through the technical side of it, might not be possible

maryokhin:
Should be possible if it doesn't break everyones API structure as soon as they upgrade, will take a look.

[...]

PR #443, opened 2018-07-11
"Data" key optional in relationships

Anton-Shutik:
Fixes #

## Description of the Change
This PR adds a flag `render_data` on ResourceRelatedField. Passing `render_data=False` will skip rendering "data" in responce, thus saving sql queries and improveing performance.

Json api docs says:
>A “relationship object” MUST contain at least one of the following: 
links, data, meta

Before:
```
"relationships": {
    "comments": {
        "data": [
             {
                 "type": "comments",
                  "id": "874"
             },
             {
                  "type": "comments",
                   "id": "877"
              },
              {
                  "type": "comments",
                  "id": "2273"
              }
       ],
       "links": {
           "self": "http://127.0.0.1:9000/drf/users/1/relationships/comments/",
           "related": "http://127.0.0.1:9000/drf/users/1/comments/",
           "first": "http://127.0.0.1:9000/drf/users/1/comments/?page=1",
            "last": "http://127.0.0.1:9000/drf/users/1/comments/?page=1",
            "next": null,
            "prev": null
       }
},
```

Using `comments=ResourceRelatedField(render_data=False, ...):

```
"relationships": {
    "comments": {
       "links": {
           "self": "http://127.0.0.1:9000/drf/users/1/relationships/comments/",
           "related": "http://127.0.0.1:9000/drf/users/1/comments/",
           "first": "http://127.0.0.1:9000/drf/users/1/comments/?page=1",
            "last": "http://127.0.0.1:9000/drf/users/1/comments/?page=1",
            "next": null,
            "prev": null
       }
},
```

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/rest_framework_json_api/relations.py
+++ b/rest_framework_json_api/relations.py
@@ -7,8 +7,8 @@ import six
 from django.core.exceptions import ImproperlyConfigured
 from django.urls import NoReverseMatch
 from django.utils.translation import ugettext_lazy as _
-from rest_framework.fields import MISSING_ERROR_MESSAGE
-from rest_framework.relations import MANY_RELATION_KWARGS, PrimaryKeyRelatedField
+from rest_framework.fields import MISSING_ERROR_MESSAGE, SkipField
+from rest_framework.relations import MANY_RELATION_KWARGS, PrimaryKeyRelatedField, ManyRelatedField as DRFManyRelatedField
 from rest_framework.reverse import reverse
 from rest_framework.serializers import Serializer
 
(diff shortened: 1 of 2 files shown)


sliverc:
Thanks for proposing this feature. I think this is mostly related to issue #296 (I guess #178 is a slightly different approach)

I have a feeling such a change should not be added to `ResourceRelatedField` but to a different field resp. serializer. There is already serializer `serializer.HyperlinkedModelSerializer` and `rest_framework.relations.HyperlinkedRelatedField` which according to the documentation on [DRF](http://www.django-rest-framework.org/api-guide/serializers/#hyperlinkedmodelserializer) should be used for such a use case. What do you think?

In terms of performance resp. too many sql queries: What I usually do is to use prefetch related data as documented [here](https://django-rest-framework-json-api.readthedocs.io/en/stable/usage.html?highlight=performance#performance-improvements). Maybe this helps you as well.

sliverc:
Continuing discussion from #296 
> We cant have 3 Fields like LinksField, DataField, MetaField, and use them separately.

Maybe this would be the best but I am not sure whether there is really a use case for all three fields even though the specification allows it.

I think for now we should stick to the two use cases:
* resource representing all parts (meta, links, data)
* resource only represented as links and optional meta (for better performance)

> I've tried to create HyperlinkedRelatedField based on DRF's but came up rewriting same code from ResourceRelatedField about link handling.

What about moving the `get_links` logic into a mixin? This could then be used from `ResourceRelatedField` and `HyperlinkedRelatedField`. If this solves the issue that we do not need to use `SkipField` and overwriting `ManyReleatedField` I would certainly prefer it.
As you have tried this in case of deriving from `HyperlinkedRelatedField` did you still need to overwrite `ManyReleatedField` field or not?

Anton-Shutik:
@sliverc I've tried to implement it in this [PR](https://github.com/django-json-api/django-rest-framework-json-api/pull/445)
Current PR looks easier and safer to me.

sliverc:
I am closing this in favor of #445

PR #451, opened 2018-08-03
Add support for related links using parent view and its permissions

### LINKED ISSUE #426: Right way to represent relationships links.related view?
I'm trying to correctly return a links.related view (not links.self) and haven't been able to find a good example. In looking at the example app it appears to have the same problem. Here's a fragment of my urlconf and view definitions:
```python
router = routers.DefaultRouter()
router.register(r'courses', views.CourseViewSet)
router.register(r'course_terms', views.CourseTermViewSet)

urlpatterns = [
    url(r'^v1/', include(router.urls)),
    # http://127.0.0.1:8000/v1/courses/f009e671-b615-42c7-b35f-8500d7ef5d24/relationships/course_terms
    url(r'^v1/courses/(?P<pk>[^/.]+)/relationships/(?P<related_field>\w+)',
        views.CourseRelationshipView.as_view(),
        name='course-relationships'),
    # http://127.0.0.1:8000/v1/courses/f009e671-b615-42c7-b35f-8500d7ef5d24/course_terms/
    url(r'^v1/courses/(?P<fk>[^/.]+)/course_terms/',  # need fk (course_id) not pk (id)
        views.CourseTermViewSet.as_view({'get': 'list'}),
        name='course-course_terms'),
]
```

```python
class CourseBaseViewSet(AuthnAuthzMixIn, SortMixin, FilterMixin, viewsets.ModelViewSet):
    pass


class CourseViewSet(CourseBaseViewSet):
    # API endpoint that allows course to be viewed or edited.
    queryset = Course.objects.all()
    serializer_class = CourseSerializer


class CourseTermViewSet(CourseBaseViewSet):
    # API endpoint that allows CourseTerm to be viewed or edited.
    queryset = CourseTerm.objects.all()
    serializer_class = CourseTermSerializer
#    queryset = queryset.filter(course_id='ec008c20-79a1-4ca7-931a-019d62c219c9')


class CourseRelationshipView(AuthnAuthzMixIn, RelationshipView):
    queryset = Course.objects
    self_link_view_name = 'course-relationships'
```

A get of a course looks like this:
`GET http://127.0.0.1:8000/v1/courses/ec008c20-79a1-4ca7-931a-019d62c219c9/`

```json
{
    "data": {
        "type": "courses",
        "id": "ec008c20-79a1-4ca7-931a-019d62c219c9",
        "attributes": {
            "school_bulletin_prefix_code": "XCEFK9",
            "suffix_two": "00",
            "subject_area_code": "PSYB",
            "course_number": "00241",
            "course_identifier": "PSYC1138X",
            "course_name": "SOCIAL PSYCHOLOGY-LEC",
            "course_description": "SOCIAL PSYCHOLOGY-LEC",
            "effective_start_date": null,
            "effective_end_date": null,
            "last_mod_user_name": "loader",
            "last_mod_date": "2018-03-11"
        },
        "relationships": {
            "course_terms": {
                "data": [
                    {
                        "type": "course_terms",
                        "id": "3ce01e27-9a68-4970-b48a-e6a83166ca41"
                    }
                ],
                "links": {
                    "self": "http://127.0.0.1:8000/v1/courses/ec008c20-79a1-4ca7-931a-019d62c219c9/relationships/course_terms",
                    "related": "http://127.0.0.1:8000/v1/courses/ec008c20-79a1-4ca7-931a-019d62c219c9/course_terms/"
                },
                "meta": {
                    "count": 1
                }
            }
        },
        "links": {
            "self": "http://127.0.0.1:8000/v1/courses/ec008c20-79a1-4ca7-931a-019d62c219c9/"
        }
    }
}
```

But `GET http://127.0.0.1:8000/v1/courses/ec008c20-79a1-4ca7-931a-019d62c219c9/course_terms/` returns all the course_terms, not just those with `fk=ec008c20-79a1-4ca7-931a-019d62c219c9` because the foreign key is not being applied. (You can see where I tested and with `queryset = queryset.filter(course_id='ec008c20-79a1-4ca7-931a-019d62c219c9')`) so I'm wondering what the **right** way to do this is.

Am I missing something? I've reproduced this with the example app and it also does it "wrong". I think I need to extend the `view` function to check for some kwargs (e.g. fk) and filter the manager.

Thanks in advance for any help.

### LINKED ISSUE #450: How do we check permissions for related entities?
Let's say we  have config like this:

views.py
```python
class CustomerProfileViewSet(viewsets.ModelViewSet):
    permission_classes = (IsCustomerOrCreation,)


class ProductViewSet(viewsets.ModelViewSet):
    permission_classes = []

    def get_object(self,):
        customer_pk = self.kwargs.get('customer_pk', None)
            if customer_pk is not None:
                customer = get_object_or_None(Customer, pk=customer_pk)
                if customer is not None:
                    return customer.favourite_product
        return super(ProductViewSet, self).get_object()
```

urls.py
```python
url(r'^api/customer/(?P<pk>\d+)/$', CustomerProfileViewSet.as_view({'get': 'retrieve'})),
url(r'^api/customer/(?P<customer_pk>\d+)/favorite-product$', ProductViewSet.as_view({'get': 'retrieve'})),
```

`GET api/customer/1/` Will give 403 error for any user whose id differs from 1. And that is fine.
Then if anon user goes to `api/customer/1/favourite-product/` they will get 200 response with product payload, because permissions for product checked in `ProductViewSet`, which have no permissions for any role. But I think it is wrong and should return 403, since we have to check permission for customer first. We actually just told everybody what is customer's 1 favorite product!

**I understand that permissions it is out of json api spec, but may be there is a nice way I can handle permissions with this lib correctly?**

For now I do it like below:
A RelatedMixin class will get a related entity and will find corresponding a serializer for that. And all that stuff will happen **after**  we check permissions for parent ("customer" in the case) entity

```python
class RelatedMixin(object):
    serializer_mapping = {}
    field_name_mapping = {}

    def get_related(self, request, *args, **kwargs):
        serializer_kwargs = {}
        instance = self.get_related_instance()

        if callable(instance):
            instance = instance()

        if hasattr(instance, 'all'):
            instance = instance.all()

        if instance is None:
            return Response(data=None)

        if isinstance(instance, Iterable):
            serializer_kwargs['many'] = True

        serializer = self.get_serializer(instance, **serializer_kwargs)
        return Response(serializer.data)

    def get_serializer_class(self):
        field_name = self.get_related_field_name()
        return self.serializer_mapping[field_name]

    def get_related_field_name(self):
        field_name = self.kwargs['related_field']
        if field_name in self.field_name_mapping:
            return self.field_name_mapping[field_name]
        return field_name

    def get_related_instance(self):
        try:
            return getattr(self.get_object(), self.get_related_field_name())
        except AttributeError:
            from rest_framework.exceptions import NotFound
            raise NotFound
```

and then just make a view that will handle all Customer's related entities:

```python
class CustomerProfileRelatedViewSet(RelatedMixin, CustomerProfileViewSet):
    serializer_mapping = {
        'favourite_product': ProductSerializer,
        #'order_list': OrderSerializer
}
```
urls.py # Only one url route required for all related entities
```python
url(r'^api/customer/(?P<customer_pk>\d+)/(?P<related_field>\w+)/$', CustomerProfileRelatedViewSet.as_view({'get': 'get_related'})),
```

What do you think about it ? If you like the idea I'm ready to come up with PR

Anton-Shutik:
Fixes #450 
Fixes #426

## Description of the Change

Add `RelatedMixin`. This introduces new approach in handling related urls/entities. `RelatedMixin` will handle all related entities, configured in `related_serializers` dict with no related views required. Also it will check permissions for parent object for any related entity.

## Checklist

- [x] unit-test added
- [x] documentation updated

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -443,6 +443,50 @@ class LineItemViewSet(viewsets.ModelViewSet):
 not render `data`. Use this in case you only need links of relationships and want to lower payload
 and increase performance.
 
+#### Related urls
+
+There is a nice way to handle "related" urls like `/orders/3/lineitems/` or `/orders/3/customer/`.
+All you need is just add to `urls.py`:
+```python
+url(r'^orders/(?P<pk>[^/.]+)/(?P<related_field>\w+)/$',
+        OrderViewSet.as_view({'get': 'retrieve_related'}),
+        name='order-related'),
+```
+Make sure that RelatedField declaration has `related_link_url_kwarg='pk'` or simply skipped (will be set by default):
+```python
+    line_items = ResourceRelatedField(
+        queryset=LineItem.objects,
+        many=True,
+        related_link_view_name='order-lineitems-list',
+        related_link_url_kwarg='pk',
+        self_link_view_name='order_relationships'
+    )
+
+    customer = ResourceRelatedField(
+        queryset=Customer.objects,
+        related_link_view_name='order-customer-detail',
+        self_link_view_name='order-relationships'
+    )
+```
+And, the most important part - declare serializer for each related entity:
+```python
+class OrderSerializer(serializers.HyperlinkedModelSerializer):
+    ...
+    related_serializers = {
+        'customer': 'example.serializers.CustomerSerializer',
+        'line_items': 'example.serializers.LineItemSerializer'
+    }
+```
+Or, if you already have `included_serializers` declared and your `related_serializers` look the same, just skip it:
+```python
+class OrderSerializer(serializers.HyperlinkedModelSerializer):
+    ...
+    included_serializers = {
+        'customer': 'example.serializers.CustomerSerializer',
+        'line_items': 'example.serializers.LineItemSerializer'
+    }
+```
+
 ### RelationshipView
 `rest_framework_json_api.views.RelationshipView` is used to build
 relationship views (see the
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -448,6 +448,9 @@ and increase performance.
 There is a nice way to handle "related" urls like `/orders/3/lineitems/` or `/orders/3/customer/`.
 All you need is just add to `urls.py`:
 ```python
+url(r'^orders/(?P<pk>[^/.]+)/$',
+        OrderViewSet.as_view({'get': 'retrieve'}),
+        name='order-detail'),
 url(r'^orders/(?P<pk>[^/.]+)/(?P<related_field>\w+)/$',
         OrderViewSet.as_view({'get': 'retrieve_related'}),
         name='order-related'),
@@ -457,14 +460,14 @@ Make sure that RelatedField declaration has `related_link_url_kwarg='pk'` or sim
     line_items = ResourceRelatedField(
         queryset=LineItem.objects,
         many=True,
-        related_link_view_name='order-lineitems-list',
+        related_link_view_name='order-related',
         related_link_url_kwarg='pk',
-        self_link_view_name='order_relationships'
+        self_link_view_name='order-relationships'
     )
 
     customer = ResourceRelatedField(
         queryset=Customer.objects,
-        related_link_view_name='order-customer-detail',
+        related_link_view_name='order-related',
         self_link_view_name='order-relationships'
     )
 ```
--- a/rest_framework_json_api/relations.py
+++ b/rest_framework_json_api/relations.py
@@ -116,7 +116,11 @@ class HyperlinkedMixin(object):
         })
         self_link = self.get_url('self', self.self_link_view_name, self_kwargs, request)
 
-        related_kwargs = {self.related_link_url_kwarg: kwargs[self.related_link_lookup_field]}
+        if self.related_link_url_kwarg == 'pk':
+            related_kwargs = self_kwargs
+        else:
+            related_kwargs = {self.related_link_url_kwarg: kwargs[self.related_link_lookup_field]}
+
         related_link = self.get_url('related', self.related_link_view_name, related_kwargs, request)
 
         if self_link:
--- a/rest_framework_json_api/relations.py
+++ b/rest_framework_json_api/relations.py
@@ -118,9 +118,11 @@ class HyperlinkedMixin(object):
 
         """
         Assuming RelatedField will be declared in two ways:
-        1. url(r'^authors/(?P<pk>[^/.]+)/(?P<related_field>\w+)/$', AuthorViewSet.as_view({'get': 'retrieve_related'}))
-        2. url(r'^authors/(?P<author_pk>[^/.]+)/bio/$', AuthorBioViewSet.as_view({'get': 'retrieve'}))
-        In other words if related_link_url_kwarg == 'pk' it will add 'related_field' parameter to reverse()
+        1. url(r'^authors/(?P<pk>[^/.]+)/(?P<related_field>\w+)/$',
+                AuthorViewSet.as_view({'get': 'retrieve_related'}))
+        2. url(r'^authors/(?P<author_pk>[^/.]+)/bio/$',
+                AuthorBioViewSet.as_view({'get': 'retrieve'}))
+        So, if related_link_url_kwarg == 'pk' it will add 'related_field' parameter to reverse()
         """
         if self.related_link_url_kwarg == 'pk':
             related_kwargs = self_kwargs
--- a/rest_framework_json_api/relations.py
+++ b/rest_framework_json_api/relations.py
@@ -116,6 +116,12 @@ class HyperlinkedMixin(object):
         })
         self_link = self.get_url('self', self.self_link_view_name, self_kwargs, request)
 
+        """
+        Assuming RelatedField will be declared in two ways:
+        1. url(r'^authors/(?P<pk>[^/.]+)/(?P<related_field>\w+)/$', AuthorViewSet.as_view({'get': 'retrieve_related'}))
+        2. url(r'^authors/(?P<author_pk>[^/.]+)/bio/$', AuthorBioViewSet.as_view({'get': 'retrieve'}))
+        In other words if related_link_url_kwarg == 'pk' it will add 'related_field' parameter to reverse()
+        """
         if self.related_link_url_kwarg == 'pk':
             related_kwargs = self_kwargs
         else:
--- a/rest_framework_json_api/views.py
+++ b/rest_framework_json_api/views.py
@@ -106,7 +106,6 @@ class RelatedMixin(object):
     This mixin handles all related entities, whose Serializers are declared in "related_serializers"
     """
     related_serializers = {}
-    field_name_mapping = {}
 
     def retrieve_related(self, request, *args, **kwargs):
         serializer_kwargs = {}
@@ -129,7 +128,7 @@ class RelatedMixin(object):
 
     def get_serializer_class(self):
         if 'related_field' in self.kwargs:
-            field_name = self.get_related_field_name()
+            field_name = self.kwargs['related_field']
             class_str = self.related_serializers.get(field_name, None)
             if class_str is None:
                 raise NotFound
@@ -138,8 +137,12 @@ class RelatedMixin(object):
 
     def get_related_field_name(self):
         field_name = self.kwargs['related_field']
-        if field_name in self.field_name_mapping:
-            return self.field_name_mapping[field_name]
+        # Making sure we're getting correct model field/property/method name
+        try:
+            return super(RelatedMixin, self).get_serializer_class()().fields[field_name].source
+        except KeyError:
+            # Looks like the field was not declared on the serializer
+            pass
         return field_name
 
     def get_related_instance(self):
--- a/rest_framework_json_api/views.py
+++ b/rest_framework_json_api/views.py
@@ -156,10 +156,18 @@ class RelatedMixin(object):
         return field_name
 
     def get_related_instance(self):
-        try:
-            return getattr(self.get_object(), self.get_related_field_name())
-        except AttributeError:
-            raise NotFound
+        parent_obj = self.get_object()
+        parent_serializer = self.serializer_class(parent_obj)
+        field_name = self.get_related_field_name()
+        field = parent_serializer.fields.get(field_name, None)
+
+        if field is not None:
+            return field.get_attribute(parent_obj)
+        else:
+            try:
+                return getattr(parent_obj, field_name)
+            except AttributeError:
+                raise NotFound
 
 
 class ModelViewSet(AutoPrefetchMixin,
--- a/rest_framework_json_api/views.py
+++ b/rest_framework_json_api/views.py
@@ -133,20 +133,22 @@ class RelatedMixin(object):
 
             # Try get the class from related_serializers
             if hasattr(parent_serializer_class, 'related_serializers'):
-                class_str = parent_serializer_class.related_serializers.get(field_name, None)
-                if class_str is None:
+                _class = parent_serializer_class.related_serializers.get(field_name, None)
+                if _class is None:
                     raise NotFound
 
             elif hasattr(parent_serializer_class, 'included_serializers'):
-                class_str = parent_serializer_class.included_serializers.get(field_name, None)
-                if class_str is None:
+                _class = parent_serializer_class.included_serializers.get(field_name, None)
+                if _class is None:
                     raise NotFound
 
             else:
                 assert False, \
                     'Either "included_serializers" or "related_serializers" should be configured'
 
-            return import_class_from_dotted_path(class_str)
+            if not isinstance(_class, type):
+                return import_class_from_dotted_path(_class)
+            return _class
 
         return parent_serializer_class
 
--- a/rest_framework_json_api/views.py
+++ b/rest_framework_json_api/views.py
@@ -106,6 +106,7 @@ class RelatedMixin(object):
     This mixin handles all related entities, whose Serializers are declared in "related_serializers"
     """
     related_serializers = {}
+    related_field_mapping = {}
 
     def retrieve_related(self, request, *args, **kwargs):
         serializer_kwargs = {}
@@ -137,12 +138,8 @@ class RelatedMixin(object):
 
     def get_related_field_name(self):
         field_name = self.kwargs['related_field']
-        # Making sure we're getting correct model field/property/method name
-        try:
-            return super(RelatedMixin, self).get_serializer_class()().fields[field_name].source
-        except KeyError:
-            # Looks like the field was not declared on the serializer
-            pass
+        if field_name in self.related_field_mapping:
+            return self.related_field_mapping[field_name]
         return field_name
 
     def get_related_instance(self):
--- a/rest_framework_json_api/views.py
+++ b/rest_framework_json_api/views.py
@@ -1,3 +1,5 @@
+from collections import Iterable
+
 from django.core.exceptions import ImproperlyConfigured
 from django.db.models import Model
 from django.db.models.fields.related_descriptors import (
@@ -96,6 +98,54 @@ class AutoPrefetchMixin(object):
                 qs = qs.prefetch_related(included.replace('.', '__'))
 
         return qs
+
+
+class RelatedMixin(object):
+    """
+    This mixin handles all related entities, whose Serializers are declared in "related_serializers"
+    """
+    related_serializers = {}
+    field_name_mapping = {}
+
+    def retrieve_related(self, request, *args, **kwargs):
+        serializer_kwargs = {}
+        instance = self.get_related_instance()
+
+        if hasattr(instance, 'all'):
+            instance = instance.all()
+
+        if callable(instance):
+            instance = instance()
+
+        if instance is None:
+            return Response(data=None)
+
+        if isinstance(instance, Iterable):
+            serializer_kwargs['many'] = True
+
+        serializer = self.get_serializer(instance, **serializer_kwargs)
+        return Response(serializer.data)
+
+    def get_serializer_class(self):
+        if 'related_field' in self.kwargs:
+            field_name = self.get_related_field_name()
+            _class = self.related_serializers.get(field_name, None)
+            if _class is None:
+                raise NotFound
+            return _class
+        return super(RelatedMixin, self).get_serializer_class()
+
+    def get_related_field_name(self):
+        field_name = self.kwargs['related_field']
+        if field_name in self.field_name_mapping:
+            return self.field_name_mapping[field_name]
+        return field_name
+
+    def get_related_instance(self):
+        try:
+            return getattr(self.get_object(), self.get_related_field_name())
+        except AttributeError:
+            raise NotFound
 
 
 class ModelViewSet(AutoPrefetchMixin, PrefetchForIncludesHelperMixin, viewsets.ModelViewSet):
--- a/rest_framework_json_api/views.py
+++ b/rest_framework_json_api/views.py
@@ -105,8 +105,6 @@ class RelatedMixin(object):
     """
     This mixin handles all related entities, whose Serializers are declared in "related_serializers"
     """
-    related_serializers = {}
-    related_field_mapping = {}
 
     def retrieve_related(self, request, *args, **kwargs):
         serializer_kwargs = {}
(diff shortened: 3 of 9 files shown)


sliverc:
@Anton-Shutik Thanks for your PR. I hope I will get to review it in the next few days.

rest_framework_json_api/views.py
```diff
@@ -98,6 +100,54 @@ def get_queryset(self, *args, **kwargs):
         return qs
 
 
+class RelatedMixin(object):
+    """
+    This mixin handles all related entities, whose Serializers are declared in "related_serializers"
+    """
+    related_serializers = {}
```
sliverc: could we not use `included_serailizers` on the defined serializer instead?

rest_framework_json_api/views.py
```diff
@@ -98,6 +100,54 @@ def get_queryset(self, *args, **kwargs):
...
+    field_name_mapping = {}
```
sliverc: what's the use case that this mapping is needed?

rest_framework_json_api/views.py
```diff
@@ -98,6 +100,54 @@ def get_queryset(self, *args, **kwargs):
...
+
+        serializer = self.get_serializer(instance, **serializer_kwargs)
+        return Response(serializer.data)
+
+    def get_serializer_class(self):
+        if 'related_field' in self.kwargs:
+            field_name = self.get_related_field_name()
+            _class = self.related_serializers.get(field_name, None)
+            if _class is None:
+                raise NotFound
```
sliverc: this seems to be a configuration issue, so rather seems to be a 500 error to me which could contain more detail what is missing. Or do you have a specific purpose to use `NotFound` here?

rest_framework_json_api/views.py
```diff
@@ -98,6 +100,54 @@ def get_queryset(self, *args, **kwargs):
...
+        field_name = self.kwargs['related_field']
+        if field_name in self.field_name_mapping:
+            return self.field_name_mapping[field_name]
+        return field_name
+
+    def get_related_instance(self):
+        try:
+            return getattr(self.get_object(), self.get_related_field_name())
+        except AttributeError:
+            raise NotFound
```
sliverc: same here as above.

rest_framework_json_api/relations.py
```diff
@@ -116,7 +116,11 @@ def get_links(self, obj=None, lookup_field='pk'):
         })
         self_link = self.get_url('self', self.self_link_view_name, self_kwargs, request)
 
-        related_kwargs = {self.related_link_url_kwarg: kwargs[self.related_link_lookup_field]}
+        if self.related_link_url_kwarg == 'pk':
```
sliverc: You have more insight into generation of this links. Can you briefly explain why this is needed? In what case would this if match?

rest_framework_json_api/views.py
```diff
@@ -98,6 +100,54 @@ def get_queryset(self, *args, **kwargs):
         return qs
 
 
+class RelatedMixin(object):
```
sliverc: Is there any reason we shouldn't directly derive `ModelViewSet` from this mixin? This would be easier to use but have to think about whether it breaks backwards compatibility.

rest_framework_json_api/views.py
Anton-Shutik: I could, but I think it will be better to keep it separate. Otherwise all related entities described here will be available via `include` by default, and there is no way to remove it from `include` list.

rest_framework_json_api/views.py
Anton-Shutik: That is used when url `related_field` name differs from model field name. Let's say we have field `Author.bio`,  but url we want is `/author/1/biography/`. Then we do:
```python
field_name_mapping = {'biography': 'bio'}
```

rest_framework_json_api/views.py
Anton-Shutik: I think it is correct. What if we go to `/author/1/unexisting_relation/`  ? I think it should return 404 as the url simply does not exist. There is no such a relation on Author model.
If there is a field on a Author model, but there is no serializer declaration for that field - it will give 404 as well. That might be used for "closing/hiding" some related resources.

http://jsonapi.org/format/#fetching-relationships-responses-404

Spec does not describe exactly our case, but I think 404 is correct behavior.

rest_framework_json_api/views.py
Anton-Shutik: If there is no attribute on model - return 404. This is similar behavior to the answer above

rest_framework_json_api/relations.py
Anton-Shutik: Oh yeah!
The `if` would match in any case you didn't pass the argument when declaring a field, or the argument value equals 'pk'.

When `related_link_url_kwarg == 'pk'` we can suppose the url was declared like this:
```python
url(r'^authors/(?P<pk>[^/.]+)/(?P<related_field>\w+)/$',
        AuthorViewSet.as_view({'get': 'retrieve_related'}),
        name='author-related'),
```
and the route has 2 parameters: `pk` and `related_field`. Exactly like `self_link`. That is why we do: `related_kwargs = self_kwargs` in the next line.

In any other cases (`related_link_url_kwarg != 'pk'`) we assume that we have "old-style" urls declaration like this:
```python
url(r'^authors/(?P<author_pk>[^/.]+)/bio/$',
        AuthorBioViewSet.as_view({'get': 'retrieve'}),
        name='author-bio'),
```

Anton-Shutik:
@sliverc thanks for you comments. 

So, changes I'm going to do:
1. Inherit `ModelViewSet` from `RelatedMixin`
2. Raise 500 in `get_serializer_class` if it is not found in `related_serializers` and give some description

How that works for you ?

rest_framework_json_api/views.py
sliverc: Would it be not more consistent from an api caller when a relationship can be either fetch per link and per include? An api caller can then decide what he prefers. This way it would also be DRY.

rest_framework_json_api/views.py
sliverc: Is this not the same which can already be done in the serializer with `source` attribute on a field? Would it be not more consistent to take the mapping information form the serializer directly?

rest_framework_json_api/views.py
sliverc: fair enough, makes sense.

rest_framework_json_api/views.py
sliverc: that would be more user friendly and a good improvement.

rest_framework_json_api/relations.py
sliverc: makes sense. What do you think about adding a small doc string explaining this?

Also do you think this could break backwards compatibility in case someone used `pk` in its related link? Although documentation states differently I guess.

rest_framework_json_api/views.py
sliverc: I think so too so let's extend `ModelViewSet` and `ReadOnlyViewSet` with this mixin

sliverc:
@Anton-Shutik 
Your list looks good + see my inline comments above for further details.

rest_framework_json_api/views.py
Anton-Shutik: I'm not against of that, but how do I add `included` serializer, without adding related link ?
Let's say I want GET api/post/1/?include=author, but at the same time I want api/post/1/author/ return 404 ?

The only way I see is to use `included_serializers` if `related_serializer` is None, and use `related_serializer` otherwise. Also we need to declare related_serializer as None by default.

rest_framework_json_api/views.py
Anton-Shutik: Just tried to do that. But what if we want to force api/author/1/bio/ return 404 ? It looks like impossible, because it will start giving 500 all the time. So, I think we have to leave it as is.

rest_framework_json_api/views.py
Anton-Shutik: Just tried. No. We trying to get related_attribute from **model**, but not from serializer, so it won't work. I can remove it, but there will be no opportunity to rename relation in the url.

rest_framework_json_api/views.py
Anton-Shutik: Done

rest_framework_json_api/views.py
sliverc: Are you sure? I assume for consistency the url relation part should have the same name as the relation in the serializer right?

What we could to ensure this is get the field declaration from the serializer class, read the source attribute and use this to get the relation from the model we have access to in case it is set. This should work or not?

rest_framework_json_api/views.py
sliverc: That's true this wouldn't be possible anymore to have a include parameter allowed but not a accessible relationship.
I am more wondering whether there is really use case for this where this is needed?

In the end it is the same data returned in included and as the relationship and with `RelationMixin` permissions are also defined by parent resource which is the same with included.
Or do you have a specific use case in mind where this could be handy?

rest_framework_json_api/views.py
sliverc: I guess it could be by looking up whether the related field is defined on the serializer. If yes then it would be a 500 error if it is not defined or if it is defined a 404.

However I see this might be a bit of a overkill just for an error message. Let's leave it as it is and make sure that it is well documented.

rest_framework_json_api/relations.py
sliverc: Your added docstring looks good. Thanks.

rest_framework_json_api/views.py
sliverc: Great.

sliverc:
@Anton-Shutik see my comments above

rest_framework_json_api/views.py
Anton-Shutik: Well, probably I have the case. Let's say we want client to "enforce" fetching related resource from the "related" url rather then just by including it within parent resource. My example is:

```python
{
    "data": {
        "type": "product",
        "id": "1",
        "attributes": {
            ...
        },
        "relationships": {
            "reviews": {
                "data": [
                    {
                        "type": "product-review",
                        "id": "1"
                    },
# A lot of object identifiers
                    {
                        "type": "product-review",
                        "id": "1000"
                    },
                ],
                "meta": {
                    "count": 1000
                },
                "links": {
                    "self": "api/product/1/relationships/reviews/",
                    "related": "api/product/1/reviews/"
                }
            },
```
Let's say I do not want to give client an ability to fetch such a huge response from api/product/1/ endpoint, because it is cached by product id for example. Of source we can use full_url (with get params like `include=reviews`) as cache key in the case, but who knows how do people use it.

So, what I would do here is:
Use included_serializers by default when related_serializer is None. And use related_serializers when it is not None. Such a way we give both opportunities and people can use `included_serializers` for both inclusion and related links. When they need different behavior - they use related_serializers dict.
Sounds like a plan ? :)

rest_framework_json_api/views.py
Anton-Shutik: >Are you sure? I assume for consistency the url relation part should have the same name as the relation in the serializer right?

100% agree with you. But there are some cases when we need to have different name. For example Author.entry_set. It would be nicer to have url like /author/1/entries/, right ?

>What we could to ensure this is get the field declaration from the serializer class, read the source attribute and use this to get the relation from the model we have access to in case it is set. This should work or not?

I [tried](https://github.com/Anton-Shutik/django-rest-framework-json-api/commit/c6a77654873fa5d2460c43bc6b7efa4474ccdb46) (If you know better/nicer way to do it - pls let me know ). Looks like we have to create Serializer instance to be able to get the field. That looks more complex, but seems like it works. 
What do we do when serializer does not have such a field ? Should we just return 500 or try to use related_name we got from the url ? Will it be possible to have a related link without declaring the field on the serializer ? You probably will want to deny such a case, but seems like json api spec allows it, or at least does not deny.

rest_framework_json_api/views.py
Anton-Shutik: >I guess it could be by looking up whether the related field is defined on the serializer. If yes then it would be a 500 error if it is not defined or if it is defined a 404.
Not sure I got you :) So, if there is no attribute on a model, we check if the attribute is declared on a serializer. If it is declared - we return 500 saying smth like "Wrong config" and if it is not declared - return 404 ? 

I will agree with case return 404, because in "500" case we should just get correct model attribute name from related field "source".

Well, honestly speaking I don't like the fact that RelatedMixin depends on parent's entity serializer class, like fetching field source name and so on. Probably that is unreasonable, but who knows, haha. I like keep thing as simple as possible, then its easier to support it.

So, I would keep it as is. Any minds ?

Anton-Shutik:
@sliverc Answered/addressed your notes, thank you.

For now I see 1 unresolved place here: Who do we resolve model's related field name. Options:

1. Declare `related_name_mapping` and use it. (That was initial version). Simple, but have to declare the dict if url name differs from model field name.
2. Create a parent serializer class instance and figure out corresponding "source" and use it. A little bit more complex, but no additonal declaration required (note, only in case url's and model's names do not match). Oh, but it requires instantiating a parent serializer class. And I already see one case it won't work
```python
class AuthorSerializer():
    featured_post = SerializerMethodRelatedField (source='get_featured_post')
```
That way we have to declare `get_featured_post` on the **model** and serializer.

I would give my 80% for 1st approach, and only 20% for the second. How do you think which one is better ? May be we need somebody's else opinion ?

rest_framework_json_api/views.py
sliverc: As said above I also think this is a bit of a overkill. It was just some thoughts on how it could be done.

But let's leave it as is and let's make sure the documentation is clear on it.

rest_framework_json_api/views.py
sliverc: Still not sure on the use case :). But yeah that's a plan let's do it like you suggested. Users might come up with use case when they use it I guess.

sliverc:
I really like the idea of `RelatedMixin` a lot. It makes using links so much easier. To make it a really consistent feature though we have to make sure that it is DRY and simple to use for the DJA user.

So I think to find a solution how to retrieve relations will be crucial part of this feature.

My first idea doesn't work in case of `SerializerMethodRelatedField`. I haven't thought of that. But I think `SerializerMethodRelatedField` case also doesn't work well with `related_name_mapping` because it could be a complex queryset as method which we would need to duplicate on the model then as well.

From a DJA user perspective I think relations are defined on the serializer - meaning we should also take it from there.

Most likely the only way to implement this would be to instantiate the parent serializer and serialize the given model. Then ready the relation from serializer.data.

Something like this:
```python
instance = self.get_object()
serializer = self.get_serializer(instance)
serializer.data['name-of-relation']
```

I haven't tried so this is just an idea. But if it works don't you think that is the much cleaner way to make sure that relations on serializer are exposed in the api with the same name?

Anton-Shutik:
I think this is overkill. You need to serialize parent instance, just to get one related entity. What if parent serializer contains 20+ relations ? It will serialize all of them, but actually we will use only one. That might just hammer the database. 
Second point against that is that Serializer does not serialize related entity. So the code like this 
```python 
serializer.data['name-of-relation']
```
will give us just json representing **relation** itself (say, object identifiers, related urls), but not entire related entity. Not tried, but almost sure that it will exactly like that.

ps. small question
what stands for DRY ?
I guess DJA is the app itself (django rest framework json api)? 

Thanks

Anton-Shutik:
@sliverc Did some improves, could you pls take a look ? It will be easier to review on per commit basis. Let me know if this version looks like final and I will start writing tests and docs. Thanks

rest_framework_json_api/views.py
```diff
@@ -98,12 +101,87 @@ def get_queryset(self, *args, **kwargs):
...
+        return Response(serializer.data)
+
+    def get_serializer_class(self):
+        parent_serializer_class = super(RelatedMixin, self).get_serializer_class()
+
+        if 'related_field' in self.kwargs:
+            field_name = self.kwargs['related_field']
+
+            assert hasattr(parent_serializer_class, 'included_serializers')\
+                or self.related_serializers,\
```
sliverc: This looks goo to me. One more though on `related_serializers`.

I think we should move it to the serializer because of following reasons:
1. related fields are defined on the serializer not the view so to define `related_serializers` on serializer feels more natural to me
2. as `included_serializers` is also on serializer it is more consistent. User can then simply decide whether he simply wants related serializers or included serializers at the same spot.
3. Serializers have meta classing. This is out of scope for this PR but in the future we can do `import_class_from_dotted_path` during loading of classes and not on each requests. `PolymorphicModelSerializer` already does meta classing something we should introduce for `ModelSerializer` in the future as well.

What do you think?

rest_framework_json_api/views.py
```diff
@@ -98,12 +101,87 @@ def get_queryset(self, *args, **kwargs):
...
+        return parent_serializer_class
+
+    def get_related_field_name(self):
+        field_name = self.kwargs['related_field']
+        if field_name in self.related_field_mapping:
+            return self.related_field_mapping[field_name]
+        return field_name
+
+    def get_related_instance(self):
+        parent_obj = self.get_object()
```
sliverc: Nice! Really like it. I think we should now remove the  `related_field_mapping` as not needed in most cases anymore.

We can still leave the method `get_related_field_name` so if someone wants to have its own mapping or whatever logic to get related field name they can overwrite it.

sliverc:
Ahh yeah. I forgot [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) stands for don't repeat yourself.

rest_framework_json_api/views.py
Anton-Shutik: Actually I do think that related urls and "calculating" related entities is not a part of serializer itself, and should be done in the RelatedMixin. But I'm also ok to do it in the serializer since it just works.

Anton-Shutik:
@sliverc done + added tests. Will add docs later. If you have a time, feel free to write some docs

Anton-Shutik:
@sliverc added docs, updated changelog. Anything else required for merge ?

rest_framework_json_api/views.py
```diff
@@ -98,12 +101,83 @@ def get_queryset(self, *args, **kwargs):
...
+            field_name = self.kwargs['related_field']
+
+            # Try get the class from related_serializers
+            if hasattr(parent_serializer_class, 'related_serializers'):
+                class_str = parent_serializer_class.related_serializers.get(field_name, None)
+                if class_str is None:
+                    raise NotFound
+
+            elif hasattr(parent_serializer_class, 'included_serializers'):
+                class_str = parent_serializer_class.included_serializers.get(field_name, None)
```
sliverc: This doesn't seem to be covered by a test. It would be good to add one as this is also used in a success case.

https://codecov.io/gh/django-json-api/django-rest-framework-json-api/pull/451/diff#D7-140

docs/usage.md
```diff
@@ -443,6 +443,50 @@ class LineItemViewSet(viewsets.ModelViewSet):
 not render `data`. Use this in case you only need links of relationships and want to lower payload
 and increase performance.
 
+#### Related urls
```
sliverc: Do you think we should combine this documentation with https://django-rest-framework-json-api.readthedocs.io/en/stable/usage.html#related-fields ? 

Maybe we do not need to document overwriting of `get_queryset` anymore as this is obsolete. Or do you still see a use case for it?

Anton-Shutik: It is still working and can be used. I think we can remove the docs when we deprecate this stuff.

Anton-Shutik:
@sliverc done

Anton-Shutik:
@sliverc any more updates required in order to merge this ?

docs/usage.md
```diff
@@ -443,6 +443,50 @@ class LineItemViewSet(viewsets.ModelViewSet):
...
+url(r'^orders/(?P<pk>[^/.]+)/(?P<related_field>\w+)/$',
+        OrderViewSet.as_view({'get': 'retrieve_related'}),
+        name='order-related'),
+```
+Make sure that RelatedField declaration has `related_link_url_kwarg='pk'` or simply skipped (will be set by default):
+```python
+    line_items = ResourceRelatedField(
+        queryset=LineItem.objects,
+        many=True,
+        related_link_view_name='order-lineitems-list',
```
sliverc: shouldn't this be order-related?

docs/usage.md
```diff
@@ -443,6 +443,50 @@ class LineItemViewSet(viewsets.ModelViewSet):
...
+        related_link_url_kwarg='pk',
+        self_link_view_name='order_relationships'
+    )
+
+    customer = ResourceRelatedField(
+        queryset=Customer.objects,
+        related_link_view_name='order-customer-detail',
```
sliverc: shouldn't this be order-related?

docs/usage.md
```diff
@@ -443,6 +443,50 @@ class LineItemViewSet(viewsets.ModelViewSet):
...
+        self_link_view_name='order-relationships'
```
sliverc: above it's order_relationships and here it's order-relationships. Maybe example should be a bit more complete to also show parent view in url so it is clear what is added here. I think this kind of goes into the direction to merge old way and new way in the documentation.

Anton-Shutik:
@sliverc Fixed imports, updated docs. If you have better idea about the docs - feel free to update it as you wish.

sliverc:
Then let's merge this and improve the documentation when needed as we gain more experience when using it.
Thanks @Anton-Shutik for your work.

Anton-Shutik:
great! thanks! Are you going to release it on pypi ?

PR #453, opened 2018-08-05
Avoid exception when trying to include skipped relationship

stas:
# Fixes

It is a very weird error I hit. We are using the DRF-JSONAPI to build 
https://github.com/occrp/id-backend

One of the packages we use is [django-simple-activity](https://github.com/richardARPANET/django-simple-activity). Where this package provides a model with a polymorphic relationship which we renamed in our project based on the relationship type:
https://github.com/occrp/id-backend/blob/master/api_v3/serializers/action.py#L55-L62

I get a strange error [when I try to include dynamically](https://github.com/occrp/id-backend/blob/master/api_v3/tests/views/test_activities.py#L55-L78) in the payload the renamed relationships.

## Description of the Change

I tracked down the error here:
https://github.com/django-json-api/django-rest-framework-json-api/blob/master/rest_framework_json_api/renderers.py#L356

I was not able to reproduce it as part of the DRF-JSONAPI test suite, unfortunately (the most related changes set is #412). I'll be more than happy to provide some tests if you could direct me or just explain a bit the purpose of the `extract_included()` arguments... Maybe I'm doing something wrong.

## Checklist

- [x] unit-test added
- [ ] documentation updated

Thanks in advance for any help or for reviewing this!!!

### DIFF
--- a/rest_framework_json_api/renderers.py
+++ b/rest_framework_json_api/renderers.py
@@ -375,7 +375,7 @@ class JSONRenderer(renderers.JSONRenderer):
                 serializer_data = field.data
 
             if isinstance(field, relations.RelatedField):
-                if relation_instance is None:
+                if relation_instance is None or not serializer_data:
                     continue
 
                 many = field._kwargs.get('child_relation', None) is not None
--- a/AUTHORS
+++ b/AUTHORS
@@ -22,3 +22,4 @@ Yaniv Peer <yanivpeer@gmail.com>
 Mohammed Ali Zubair <mazg1493@gmail.com>
 Jason Housley <housleyjk@gmail.com>
 Beni Keller <beni@matraxi.ch>
+Stas S. <stas@nerd.ro>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ any parts of the framework not mentioned in the documentation should generally b
 
 ### Fixed
 
-* Resource related fields can be blank, serializer/renderer should skip these.
+* Avoid exception when trying to include skipped relationship
 
 ## [2.7.0] - 2019-01-14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 Note that in line with [Django REST Framework policy](http://www.django-rest-framework.org/topics/release-notes/),
 any parts of the framework not mentioned in the documentation should generally be considered private API, and may be subject to change.
 
+## [Unreleased]
+
+### Fixed
+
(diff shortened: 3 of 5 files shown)


sliverc:
Not quite clear what is happening would also need to do more investigation. I think it would be important to be able to reproduce this in a DJA test as well before merging this change as otherwise the error might occur in the future again after the next refactoring.

In terms of `extract_included` arguments just out of my head (hope got it right but might be good to add a doc string there I guess):
`fields` serializer fields of resource
`resource` the serialized resource (usually type and id)
`resource_instance` the model instance of the resource
`included_resources` list of relations which need to be included (this gets reduced as includes are processed)
`included_cache` list of already extracted relationship

#412 has only been introduced in version 2.5.0. Can you confirm that the issue also exists in version 2.4.0 or not? This way we at least know whether it is a regression or not.

stas:
@sliverc based on the changelog, it looks like it is a regression:
https://github.com/occrp/id-backend/compare/8cb3c5992047734c4d6fc...master#diff-8af41323990a2c8baf45a65159722ba5L67 (the last two files in the diff).

I will try to replicate the issue as part of a test case.
Thanks for the help so far!

sliverc:
@stas Have you been able to investigate this issue further? It would be great to see what causes this regression.

stas:
@silverc, apologies for the late reply!

I reviewed again the latest version and the issue persists.
Here's a similar use-case to mine.

I have a model with a polymorphic relationship which I need to serialize only if a specific polymorphic type of relationship is present. Using the `rest_framework.fields.SkipField()` leads to the issue.

Latest commit includes a test case.

To be honest, I'd expect that the `serializers.PolymorphicModelSerializer` would pick up the right serializer based on the object type so that I don't have to patch it. But this is another issue which is up to you if you think it's relevant to discuss.

sliverc:
Thanks for looking into this again.

I am bit confused that CI is green whereas the reproducible test should actually fail in my opinion showing what the error is. Or am I missing something here?

In terms of `PolymorphicModelSerializer`: Using the right serializer is a different question to be discussed in another issue if need to.

stas:
The fix is included in 8856e0f

rest_framework_json_api/renderers.py
```diff
@@ -375,6 +375,9 @@ def extract_included(cls, fields, resource, resource_instance, included_resource
                 serializer_data = field.data
 
             if isinstance(field, relations.RelatedField):
+                if not serializer_data:
```
sliverc: To reduce length of code I might be better to combine this if with the next if.

```
if relation_instance is None or not serializer_data:
```

stas:
@sliverc please let me know if there's anything else I can help with to get it merged. Thank you!

sliverc:
Merged. Thanks for your contribution.

PR #455, opened 2018-08-09
WIP: Filter Backends

n2ygk:
Fixes # 

## Description of the Change
* Add optional [jsonapi-style](http://jsonapi.org/format/) filter backends. See [usage docs](docs/usage.md#filter-backends)
  * query parameter validation -- raises 400 errors rather than silently ignoring "bad" parameters
  * sort - based on `rest_framework.filters.OrderingFilter`
  * keyword filter across multiple fields based on `rest_framework.filters.SearchFilter`
  * field-level filter based on `django_filters.rest_framework.DjangoFilterBackend`

This is an initial draft for your feedback/suggestions. Still no unit tests, and a couple of gaps in the implementation.

## Checklist

- [ ] unit-test added
- [x] documentation updated

### DIFF
--- a/rest_framework_json_api/backends.py
+++ b/rest_framework_json_api/backends.py
@@ -63,13 +63,15 @@ class JsonApiOrderingFilter(JsonApiFilterMixin, OrderingFilter):
 
     This implements http://jsonapi.org/format/#fetching-sorting and raises 400 if any sort field is invalid.
     """
+    ignore_bad_sort_fields = False
+
     def remove_invalid_fields(self, queryset, fields, view, request):
         """
         override remove_invalid_fields to raise a 400 exception instead of silently removing them.
         """
         valid_fields = [item[0] for item in self.get_valid_fields(queryset, view, {'request': request})]
         bad_terms = [term for term in fields if term.lstrip('-') not in valid_fields and ORDER_PATTERN.match(term)]
-        if bad_terms:
+        if bad_terms and not self.ignore_bad_sort_fields:
             raise ValidationError(
                 'invalid sort parameter{}: {}'.format(('s' if len(bad_terms) > 1 else ''), ','.join(bad_terms)))
         return super(JsonApiOrderingFilter, self).remove_invalid_fields(queryset, fields, view, request)
@@ -116,6 +118,7 @@ class JsonApiFilterFilter(JsonApiFilterMixin, DjangoFilterBackend):
         """
         fs = super(JsonApiFilterFilter, self).get_filterset(
             request, queryset, view)
+        # TODO: change to have option to silently ignore bad filters
         for k in self.filter_keys:
             if k not in fs.filters:
                 raise ValidationError("invalid filter[{}]".format(k))
--- a/rest_framework_json_api/backends.py
+++ b/rest_framework_json_api/backends.py
@@ -1,34 +1,34 @@
-from django.db.models.sql.constants import ORDER_PATTERN
-from django_filters.rest_framework import DjangoFilterBackend
 from rest_framework.exceptions import ValidationError
-from rest_framework.filters import (BaseFilterBackend, OrderingFilter,
-                                    SearchFilter)
+from rest_framework.filters import BaseFilterBackend, OrderingFilter, SearchFilter
 from rest_framework.settings import api_settings
 
+from django_filters.rest_framework import DjangoFilterBackend
+from rest_framework_json_api.utils import format_value
 
-class JsonApiFilterMixin(object):
+
+class JSONAPIFilterMixin(object):
     """
     class to share data among filtering backends
     """
-    jsonapi_query_keywords = ('sort', 'filter', 'fields', 'page', 'include')
+    # search_param is used both in SearchFilter and JSONAPIFilterFilter
     search_param = api_settings.SEARCH_PARAM
-    ordering_param = 'sort'
 
     def __init__(self):
         self.filter_keys = []
 
 
-class JsonApiQueryValidationFilter(JsonApiFilterMixin, BaseFilterBackend):
+class JSONAPIQueryValidationFilter(JSONAPIFilterMixin, BaseFilterBackend):
     """
-    A backend filter that validates query parameters for jsonapi spec conformance and raises a 400 error
-    rather than silently ignoring unknown parameters or incorrect usage.
+    A backend filter that validates query parameters for jsonapi spec conformance and raises a 400
+    error rather than silently ignoring unknown parameters or incorrect usage.
 
     set `allow_duplicate_filters = True` if you are OK with the same filter being repeated.
 
-    TODO: For jsonapi error object compliance, must set jsonapi errors "parameter" for the ValidationError.
-          This requires extending DRF/DJA Exceptions.
+    TODO: For jsonapi error object compliance, must set jsonapi errors "parameter" for the
+          ValidationError. This requires extending DRF/DJA Exceptions.
     """
-    allow_duplicated_filters = False
+    jsonapi_query_keywords = ('sort', 'filter', 'fields', 'page', 'include')
+    jsonapi_allow_duplicated_filters = False
 
     def validate_query_params(self, request):
         """
@@ -47,8 +47,8 @@ class JsonApiQueryValidationFilter(JsonApiFilterMixin, BaseFilterBackend):
             if keyword not in self.jsonapi_query_keywords:
                 raise ValidationError(
                     'invalid query parameter: {}'.format(keyword))
-            if not self.allow_duplicated_filters and len(
-                    request.query_params.getlist(qp)) > 1:
+            if not self.jsonapi_allow_duplicated_filters \
+                    and len(request.query_params.getlist(qp)) > 1:
                 raise ValidationError(
                     'repeated query parameter not allowed: {}'.format(qp))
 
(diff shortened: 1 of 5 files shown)


rest_framework_json_api/backends.py
```diff
@@ -0,0 +1,148 @@
...
+    jsonapi_query_keywords = ('sort', 'filter', 'fields', 'page', 'include')
+    search_param = api_settings.SEARCH_PARAM
+    ordering_param = 'sort'
+
+    def __init__(self):
+        self.filter_keys = []
+
+
+class JsonApiQueryValidationFilter(JsonApiFilterMixin, BaseFilterBackend):
+    """
```
sliverc: Not quite sure how to get this right. Especially when an API designer wants to define additional query params with different naming per view it would also need to be configurable.

All would be possible to do somehow but I kind of think it makes the life more difficult of an api designer and the gain is very little in the default case.

Wouldn't it be good enough just to verify the params in the other filter backend classes whether they exists and leave this filter class out for now?

From a maintenance perspective I would prefer not to provide this for now.

rest_framework_json_api/backends.py
```diff
@@ -0,0 +1,148 @@
...
+        return queryset
+
+
+class JsonApiOrderingFilter(JsonApiFilterMixin, OrderingFilter):
+    """
+    The standard rest_framework.filters.OrderingFilter works mostly fine as is, but with .ordering_param = 'sort'.
+
+    This implements http://jsonapi.org/format/#fetching-sorting and raises 400 if any sort field is invalid.
+    """
+    ignore_bad_sort_fields = False
```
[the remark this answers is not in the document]
n2ygk: Serves me right for trying to anticipate your review;-) I originally left it out:-)

rest_framework_json_api/backends.py
```diff
@@ -0,0 +1,148 @@
...
+        """
+        valid_fields = [item[0] for item in self.get_valid_fields(queryset, view, {'request': request})]
+        bad_terms = [term for term in fields if term.lstrip('-') not in valid_fields and ORDER_PATTERN.match(term)]
+        if bad_terms and not self.ignore_bad_sort_fields:
+            raise ValidationError(
+                'invalid sort parameter{}: {}'.format(('s' if len(bad_terms) > 1 else ''), ','.join(bad_terms)))
+        return super(JsonApiOrderingFilter, self).remove_invalid_fields(queryset, fields, view, request)
+
+
+class JsonApiSearchFilter(JsonApiFilterMixin, SearchFilter):
```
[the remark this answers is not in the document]
n2ygk: If we just include `JSONAPISearchFilter` in `DEFAULT_FILTER_BACKENDS` or the `filter_backends` attribute then the default value of `filter[search]`, `filter[all]` or whatever is predefined. Otherwise, including plain `SearchFilter` will require one to also set settings.JSON_API_SEARCH_PARAM or `search_param` attribute in order to comply with the requirement to use `filter`. So it's less typing to just use the JSONAPI-flavored class even if it doesn't do much.

Regarding the name of the filter, I like `all` (adverb) because it implies search *all* the (searchable) fields whereas `search` feels more like a verb usage to me which is anti-REST. But you can use one and I can use the other as it's configurable.

rest_framework_json_api/backends.py
```diff
@@ -0,0 +1,148 @@
...
+    """
+    allow_duplicated_filters = False
+
+    def validate_query_params(self, request):
+        """
+        Validate that query params are in the list of valid `jsonapi_query_keywords`
+        Raises ValidationError if not.
+        """
+        for qp in request.query_params.keys():
+            bracket = qp.find('[')
```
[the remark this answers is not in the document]
n2ygk: I was trying to care about performance as this code executes for every request. Also the match would be something more complicated since it's searching for all keywords, whether or not they have brackets (e.g. `filter[foo], sort, ...`). It seems a regex is overkill and detracts from readability.

rest_framework_json_api/backends.py
n2ygk: It seems inappropriate to verify the params in other classes as they only "own" their own params.  Verification is already happening there. For instance, `filter[blah]` checks to see if `blah` is a defined filter.
Some of the params are not even backend filters, like `include` which is elsewhere in the serializer code.

The problem this class addresses is that a client would inadvertently misspell a param and then (if they don't notice that) wonder why the action they requested isn't happening. I'm all for making it easier on the client developer to notice these kinds of errors. If they want to "break" JSON:API and add additional parameters they can:
1. Not include this backend filter at all and the default silent behavior of not complaining about params that are not recognized is achieved -- which was kind of the reason for the `ignore_bad_sort_fields` as well; If you really want your API to be silent to all query parameter errors, this is needed.
2. Extend this backend filter to add additional params to the list by changing the `jsonapi_query_keywords` attribute (or add a `JSON_API_QUERY_KEYWORDS` to settings although I've currently left that out to make it difficult to break the standard.)

Needless to say I feel pretty strongly that this is a useful backend filter to help achieve [JSON:API compliance for query parameter handling](http://jsonapi.org/format/#query-parameters):

"If a server encounters a query parameter that does not follow the naming conventions above, and the server does not know how to process it as a query parameter from this specification, it **MUST** return `400 Bad Request`."

rest_framework_json_api/backends.py
```diff
@@ -0,0 +1,148 @@
...
+        Turns filter[<field>]=<value> into <field>=<value> which is what DjangoFilterBackend expects
+        """
+        self.filter_keys = []
+        # rewrite `filter[field]` query parameters to make DjangoFilterBackend work.
+        data = request.query_params.copy()
+        for qp, val in data.items():
+            if qp[:
+                  7] == 'filter[' and qp[-1] == ']' and qp != self.search_param:
+                key = qp[7:-1].replace(
+                    '.', '__'
```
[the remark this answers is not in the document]
n2ygk: Ugh, I'm going to stop using yapf. It messed up the line breaks. It should be this:
```python
if qp[:7] == 'filter[' and qp[-1] == ']' and qp != self.search_param:
```

rest_framework_json_api/backends.py
n2ygk: I didn't address the different naming per view part. I believe the attribute is actually defined for the view. It's certainly declared that way. I'm guessing `backend_filters` classes are mixed in with the view class? I need to write unit tests for these permutations to confirm.

rest_framework_json_api/backends.py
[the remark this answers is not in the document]
sliverc: I kind of think `filter[all]` is misleading as it implies that it takes into account all fields for filtering. This is actually not true as it only search in the defined `search_fields`. `filter[search]` might also be a bit odd though I agree.

Even though everyone can define the name as they like if we add it to DJA we have to come up with a recommendation. But this doesn't seem to be easy as the json api is silent on it and we are just two users and have two different opinions already :smile: 

While thinking it seems that adding a `JSONAPISearchFilter` to DJA doesn't add any value as someone can already use the DRF `SearchFilter` today without this change and define whatever parameter he likes.

Also in DJA we don't have to reimplement all features which are in DRF but only implement what is different according to the json api specification. 

For this reasons I think we shouldn't provide a `JSONAPISearchFilter`. We might add a reference to the DRF documentation that there are other filter backends though.

rest_framework_json_api/backends.py
sliverc: I agree that this would make DJA better comply with JSON API spec which is one of ours goals of DJA. 

I think though there is simple no hurry to implement this in DJA and I think we shouldn't rush into implementing this without having a proper understanding how to do it well.
Because once its landed in DJA and we will find that it is not a good fit we will have to go through a deprecation cycle to change it which makes maintaining DJA harder and the code more complex.

Another reason why I think this is hard to implement is that query params can be accessed wherever request is accessible. Query params are not only used in filter backends but also in other parts e.g. pagination is one example but I am sure there are others.
Also it kind of feels odd to add validation into a filter backend which should be there for filtering.
Maybe a middleware might be the better spot but I am not sure and would try to experiment with it first to get a clear understanding.

We could think of an experimental area for this. We haven't done this in the past and would need to rethink how to structure it though.

So why do we not just try it out in our projects (I can also think of one of our projects where it could be helpful)? Once we have it running in production and it is solid we can still consider adding it to DJA. What do you think?

n2ygk: I'm going to finish up this PR, leaving it marked WIP, and then maybe we can make an experimental branch and merge it there?  We can always rebase later. I have a working example and test case in my local code but wanted to merge in to the example and test cases here so others can try it and provide feedback.

I would argue that **filtering** for valid query parameters is a legit use of a filter;-) But this could be done as a view mixin too. I think middleware is too early in the stack as we haven't even gotten to DRF at that point, whereas backend filters are a DRF layer.

rest_framework_json_api/backends.py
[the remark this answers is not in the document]
n2ygk: Let's call it `filter[all-the-things]` ;-)

I agree with your reasoning for calling it something that doesn't imply every field. How about `filter[fields]` or `filter[search_fields]` as the default?

I'm OK with documenting to just use `SearchFilter` and set `search_param`. I just needed a way to make sure when validating filter field names that the one we use for keyword searching doesn't come up as an error. I can just as easily change the code to look for DRF's settings.REST_FRAMEWORK['SEARCH_PARAM'].

I'll drop the code and document example usage of SearchFilter.

n2ygk:
closing (for now)

rest_framework_json_api/backends.py
n2ygk: @sliverc FYI - I've closed the PR while I keep working on this. Do we want to create an experiment branch so others can test/contribute?

BTW, using DjangoFilterBackend adds a new dependency for `django-filter`. That could be a problem. It's not core Django or DRF but is the top-rated filtering package and the one reference from DRF docs.

sliverc: In terms of `django-filter`. We have the same issue with `django-polymorphic`. This is marked as optional dependency, we can do the same with `django-filter`. We just have to make sure that DJA can still be used when this dependency is not installed - meaning we have to be careful with imports.

Actually `django-filter` has the exact same issue just the other way around as it is also usable without Django REST Framework. All rest_framework related classes are part of the `rest_framework` module. We could create a `django_filters` module in our case.

In terms of how to move forward:
I am an advocate of small PRs as they are simpler to review and there quicker to get merged. 

What I would suggest how we move forward:
1. Create a PR for `JSONAPIOrderingFilter` - this is a fairly straight forward change and can be merged easily
2. Create a PR for `JSONAPIDjangoFilterBackend` with its own module approach
3. Create a PR for `JsonApiQueryValidationFilter` where we can experiment how this works best. (I will comment on the experimental discussion you started later on)

How does this sound?

n2ygk: @sliverc you wrote:

> In terms of django-filter. We have the same issue with django-polymorphic. This is marked as optional dependency, we can do the same with django-filter. We just have to make sure that DJA can still be used when this dependency is not installed - meaning we have to be careful with imports.

Actually there are no imports for django-polymorphic. I don't quite understand how this code works: https://github.com/django-json-api/django-rest-framework-json-api/blob/a3205367c8050c978c636cb6b7867b5351c3b2d5/rest_framework_json_api/serializers.py#L215 w.r.t. django-polymorphic.

> Actually django-filter has the exact same issue just the other way around as it is also usable without Django REST Framework. All rest_framework related classes are part of the rest_framework module. We could create a django_filters module in our case.

I don't quite understand what you mean by creating a `django_filters` module. Are you saying submit this as a PR to django-filter? As it stands right now, I'll either have to wrap the `JSONAPIDjangoFilter` implementation with a try/except of `from django_filters.rest_framework import DjangoFilterBackend` or just make `django-filter` a dependency for DJA.  Please advise.

sliverc: No, what I mean is to create a folder django_filters in DJA and add a __init__,py and backends.py to implement a custom backend. (Hint: Simplified each python file is called a module but more details at https://docs.python.org/3.6/tutorial/modules.html)

If a user of DJA doesn't import this newly created `django_filters` module or uses it in settings then he doesn't have to install django_filters.

PR #459, opened 2018-08-20
JSONAPIOrderingFilter

### LINKED ISSUE #432: spec compliance: sparse fieldsets, sorting & filtering query parameters

n2ygk:
Fixes #432 

## Description of the Change
Partially addresses #432 by adding `JSONAPIOrderingFilter` backend. Broken up into smaller PR's per @sliverc.

## Checklist

- [x] unit-test added
- [x] documentation updated

### DIFF
--- a/example/fixtures/blogentry.yaml
+++ b/example/fixtures/blogentry.yaml
@@ -1,0 +1,60 @@
+- model: example.blog
+  pk: 1
+  fields: {name: ANTB, tagline: ANTHROPOLOGY (BARNARD), created_at: '2018-08-20', modified_at: '2018-08-20'}
+- model: example.blog
+  pk: 2
+  fields: {name: CLSB, tagline: CLASSICS (BARNARD), created_at: '2018-08-20', modified_at: '2018-08-20'}
+- model: example.blog
+  pk: 3
+  fields: {name: AMSB, tagline: AMERICAN STUDIES (BARNARD), created_at: '2018-08-20', modified_at: '2018-08-20'}
+- model: example.blog
+  pk: 4
+  fields: {name: CHMB, tagline: CHEMISTRY (BARNARD), created_at: '2018-08-20', modified_at: '2018-08-20'}
+- model: example.blog
+  pk: 5
+  fields: {name: ARHB, tagline: ART HISTORY (BARNARD), created_at: '2018-08-20', modified_at: '2018-08-20'}
+- model: example.blog
+  pk: 6
+  fields: {name: ITLB, tagline: ITALIAN (BARNARD), created_at: '2018-08-20', modified_at: '2018-08-20'}
+- model: example.blog
+  pk: 7
+  fields: {name: BIOB, tagline: BIOLOGICAL SCIENCES (BARNARD), created_at: '2018-08-20', modified_at: '2018-08-20'}
+- model: example.entry
+  pk: 1
+  fields: {blog: 1, headline: ANTH1009V, body_text: INTRO TO LANGUAGE & CULTURE, created_at: '2018-08-20', modified_at: '2018-08-20'}
+- model: example.entry
+  pk: 2
+  fields: {blog: 2, headline: CLCV2442V, body_text: EGYPT IN CLASSICAL WORLD-DISC, created_at: '2018-08-20', modified_at: '2018-08-20'}
+- model: example.entry
+  pk: 3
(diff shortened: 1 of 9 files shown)


example/fixtures/blogentry.yaml
```diff
@@ -0,0 +1,60 @@
+- model: example.blog
```
sliverc: let's also do this in JSON as in drf_example

example/tests/test_backends.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def setUp(self):
+        self.entries = Entry.objects.all()
+        self.blogs = Blog.objects.all()
+
+    def test_sort(self):
+        """
+        test sort
+        """
+        response = self.client.get(ENTRIES + '?sort=headline')
```
sliverc: parameters can be passed on as data e.g. in this case data={'sort': 'headline'}. This way there is no need for string concats

example/tests/test_backends.py
```diff
@@ -0,0 +1,56 @@
...
+        self.assertEqual(response.status_code, 200,
+                         msg=response.content.decode("utf-8"))
+        j = json.loads(response.content.decode("utf-8"))
```
sliverc: to get json result from response you should be able to simply do `result = response.json()` I would also prefer not to use single letter variables as python is not strongly typed so the name should tell what it which makes the code more readable (e.g. in this example result instead j or similar)

The same counts to the tests below.

example/tests/test_backends.py
```diff
@@ -0,0 +1,56 @@
+import json
+
+from rest_framework.test import APITestCase
+
+from ..models import Blog, Entry
+
+ENTRIES = "/nopage-entries"
```
sliverc: we shouldn't build routes manually but instead use [django reverse resolver](https://docs.djangoproject.com/en/2.1/ref/urlresolvers/#reverse)

In this case this would be `reverse('nopage-entry')`. It will be then a full path of the testserver url.

This needs to be done in setUp or what I usually do in each single test as it is more implicit.

example/tests/test_backends.py
n2ygk: I've seen cases where `response.json()` doesn't work. Taking a look if this is one of them.

example/tests/test_backends.py
n2ygk:
```
django.urls.exceptions.NoReverseMatch: Reverse for 'nopage-entry' not found. 'nopage-entry' is not a valid view function or pattern name.
```

n2ygk: maybe `nopage-entry-list`

example/tests/test_backends.py
n2ygk: but this is not one of them.

sliverc:
I am bit confused that you merged this before I had the chance to look at the changes you did again.

I think it is important that we only merge once changes are approved.

Reasons for this:
1. Requested changes need to be reviewed again whether there are done as expected.
2. sometimes because of complexity of code errors might only be found at a second glance

In this case the latter is actually the case. Just when looking at the code I have the feeling that when we pass on a camel cased variable e.g. blogEntry in case `blog_entry` sort field exists no error would be throw but the parameter be removed and no sorting will happen because parent class with `remove_invalid_fields` is called.

It would be great if you could verify this (best in a test) and open up another PR.

Also another thing I noticed is I think we should stick to the same naming of modules as Django REST Framework does meaning the module should be named `filters.py` instead of `backends.py`. This seems to be more consistent to me.

n2ygk:
Sorry, I thought you indicated the changes were sufficient. Won’t do it
again.

n2ygk:
@sliverc wrote:
> In this case the latter is actually the case. Just when looking at the code I have the feeling that when we pass on a camel cased variable e.g. blogEntry in case blog_entry sort field exists no error would be throw but the parameter be removed and no sorting will happen because parent class with remove_invalid_fields is called.

I override `remove_invalid_fields` and use `format_value` here https://github.com/django-json-api/django-rest-framework-json-api/blob/e6290af5c34bf1760cb8c79fb8648b58e9336685/rest_framework_json_api/backends.py#L29

However, I think I see a fundamental problem with the information-losing reversibility of camel case, etc. Namely, if the model uses camel case rather than the "preferred" underscore for field names, the `format_value(...'underscore')` will do the wrong thing. I guess perhaps we just need to make the documentation a bit stronger about that requirement?

I'll take a look at more explicit testing and submit a PR shortly.

n2ygk:
Of course you were correct, the sort is broken. Fixing.

PR #461, opened 2018-08-21
Do not lazy load forward relations

### LINKED ISSUE #337: Query explosion even on simple models with no included serializers

timcbaoth:
If the forward relations are included, the renderer will break, so this optimization only works if it is not included.

## Description of the Change
Fixes a copy paste error, which resulted in an optimization not being applied.

## Checklist

- [x] unit-test added
- [ ] documentation updated


sliverc:
Could you elaborate what you mean with `the rendererer will break`? I think best would be to reproduce this in a test case then I think things will be clearer and we make sure this error won't happen again in the future.

timcbaoth:
A test case was added.

example/tests/test_serializers.py
```diff
@@ -30,6 +32,32 @@ def setUp(self):
...
+        MockRequest = namedtuple('Request', ['query_params'])
+        request_without_includes = MockRequest({})
+
+        class BlogSerializer(ModelSerializer):
+            class Meta:
+                model = Blog
+                fields = '__all__'
+
+            def to_representation(self, instance):
+                raise Exception('to_representation of BlogSerializer was called')
```
sliverc: the disadvantage of this way of testing is that this code will actually never be covered. A bit odd when test code is not covered and the long term goal of DJA would be to have 100% coverage.

So I would prefer if the already existing serializers in example/serializers.py are used for this. `to_representation` would need to be mocked and [assert_not_called](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_not_called) could be used to check that to_representation has not been called.

example/tests/test_serializers.py
```diff
@@ -30,6 +32,32 @@ def setUp(self):
...
+            class Meta:
+                model = Entry
+                fields = '__all__'
+
+            included_serializers = {
+                'blog': BlogSerializer,
+            }
+
+        serializer = EntrySerializer(context={'request': request_without_includes})
+        serializer.to_representation(self.entry)
```
sliverc: I think it would also be good to assert the result.

rest_framework_json_api/serializers.py
```diff
@@ -182,35 +182,41 @@ def to_representation(self, instance):
...
-            check_for_none = attribute.pk if isinstance(attribute, PKOnlyObject) else attribute
-            if check_for_none is None:
-                ret[field.field_name] = None
-            else:
-                ret[field.field_name] = field.to_representation(attribute)
-
         return ret
 
+    def _get_field_representation(self, field, instance):
+        request = self.context.get('request', None)
```
sliverc: None is the default return value when using `get`. So seems to be a bit cleaner to use `self.context.get('request')`.

:+1: to extract this logic into its own method.

example/tests/test_serializers.py
timcbaoth: I agree. I added another test to validate the result.

example/tests/test_serializers.py
timcbaoth: I didn't think about it in this way, that you also want full coverage on test code itself. Interesting.

Anyway it is nicer to have a proper assert, so I changed it according to your suggestion.

example/tests/test_serializers.py
```diff
@@ -30,6 +32,32 @@ def setUp(self):
                 Author.objects.create(name=name, email='{}@example.org'.format(name))
             )
 
+    def test_forward_relationship_not_loaded_when_not_included(self):
+        MockRequest = namedtuple('Request', ['query_params'])
```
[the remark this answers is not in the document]
timcbaoth: Thanks!

example/tests/test_serializers.py
```diff
@@ -30,6 +47,59 @@ def setUp(self):
...
+
+        request_without_includes = Request(request_factory.get('/'))
+        serializer = EntrySerializer(context={'request': request_without_includes})
+        result = serializer.to_representation(self.entry)
+
+        # Remove non deterministic fields
+        result.pop('created_at')
+        result.pop('modified_at')
+
+        expected = OrderedDict(
```
sliverc: ordering is not relevant in an assertion so I think the code would be more readable to use normal dict.

timcbaoth: Makes sense. I will also change it to assertDictEqual.

PR #466, opened 2018-08-24
django_filters.DjangoFilterBackend

### LINKED ISSUE #432: spec compliance: sparse fieldsets, sorting & filtering query parameters

n2ygk:
Fixes #432 

## Description of the Change
Implements `django_filters.DjangoFilterBackend` a Django ORM-style JSON:API `filter[]` implementation. See docs/usage.md for details.

## Checklist

- [x] unit-test added
- [x] documentation updated


n2ygk:
@sliverc I think this is *really* ready for your review now.

rest_framework_json_api/filters/filter.py
```diff
@@ -0,0 +1,135 @@
...
+import warnings
+
+from rest_framework.exceptions import ValidationError
+from rest_framework.settings import api_settings
+
+from rest_framework_json_api.utils import format_value
+
+# django-filter is an optional package. Generate a dummy class if it's missing.
+try:
+    from django_filters.rest_framework import DjangoFilterBackend
```
sliverc: Instead of try except here let's move this into its own directory/module following the same structure as django filter does.

In this example `django_filters/backends.py` would be a suitable place to put it (potentially add a import in `django_filters/__init__.py`. This way `try` `except` is not needed as when a user of DJA doesn't import this module no error will occur even when django-filters is not installed.

django-filters does it the same way with `rest_framework` and I have seen this as well in other library such as `graphene-django`.

rest_framework_json_api/filters/filter.py
```diff
@@ -0,0 +1,135 @@
...
+        # TODO: find a better way to deal with search_param.
+        search_param = api_settings.SEARCH_PARAM
+
+        # Make this regex check for 'filter' as well as 'filter[...]'
+        # Leave other incorrect usages of 'filter' to JSONAPIQueryValidationFilter.
+        # See http://jsonapi.org/format/#document-member-names for allowed characters
+        # and http://jsonapi.org/format/#document-member-names-reserved-characters for reserved
+        # characters (for use in paths, lists or as delimiters).
+        # regex `\w` matches [a-zA-Z0-9_].
+        # TODO: U+0080 and above allowed but not recommended. Leave them out for now. Fix later?
```
sliverc: I think this is fine for now to be more restrictive. Can always be relaxed at a later point.

rest_framework_json_api/filters/filter.py
```diff
@@ -0,0 +1,135 @@
...
+            warnings.warn("must install django-filter package to use JSONAPIDjangoFilter")
+
+        def filter_queryset(self, request, queryset, view):
+            """
+            do nothing
+            """
+            return queryset
+
+else:
+    class JSONAPIDjangoFilter(DjangoFilterBackend):
```
sliverc: I think we should follow the naming of upstream this way it is clear what we overwrite. In this case it would be `JSONAPIDjangoFilterBackend`.

rest_framework_json_api/filters/filter.py
```diff
@@ -0,0 +1,135 @@
...
+            """
+            Sometimes there's no filterset_class defined yet the client still
+            requests a filter. Make sure they see an error too. This means
+            we have to get_filterset_kwargs() even if there's no filterset_class.
+
+            TODO: .base_filters vs. .filters attr (not always present)
+            """
+            filterset_class = self.get_filterset_class(view, queryset)
+            kwargs = self.get_filterset_kwargs(request, queryset, view)
+            self.validate_filter(self.filter_keys, filterset_class)
```
sliverc: we should keep this filter backend as a singleton and not introduce state.

What about `get_filterset_kwargs` returns filters as well? In `get_filterset` we can then do `kwargs.pop('filters')` so `filter_class(**kwargs) will still work?

rest_framework_json_api/filters/filter.py
```diff
@@ -0,0 +1,135 @@
...
+                    self.filter_keys.append(key)
+                    del data[qp]
+            return {
+                'data': data,
+                'queryset': queryset,
+                'request': request,
+            }
+
+        def filter_queryset(self, request, queryset, view):
+            """
```
sliverc: hmm.... django-filters 2.0 does quiet a bit more here than 1.1. Not sure what errors this could cause.

Question is should we rather say we only support django-filters 2.0? This will also mean Python 2.7 users can not use it. But django-filters has done the same so we would align with upstream. And it makes our life certainly easier and we do not introduce potential errors with changes to support backwards compatibility.

What do you think?

requirements-development.txt
```diff
@@ -14,4 +14,4 @@ Sphinx
 sphinx_rtd_theme
 tox
 twine
-
+django-filter>=2.0
```
sliverc: let's add this to setup.py as well as test dependency (same way done with django-polymorphic)

rest_framework_json_api/filters/filter.py
```diff
@@ -0,0 +1,135 @@
...
+        # See http://jsonapi.org/format/#document-member-names for allowed characters
+        # and http://jsonapi.org/format/#document-member-names-reserved-characters for reserved
+        # characters (for use in paths, lists or as delimiters).
+        # regex `\w` matches [a-zA-Z0-9_].
+        # TODO: U+0080 and above allowed but not recommended. Leave them out for now. Fix later?
+        #       Also, ' ' (space) is allowed within a member name but not recommended.
+        filter_regex = re.compile(r'^filter(?P<ldelim>\[?)(?P<assoc>[\w\.\-]*)(?P<rdelim>\]?$)')
+
+        def validate_filter(self, keys, filterset_class):
+            for k in keys:
```
sliverc: for validation whether a filter is used shouldn't we not also use the form?

See https://github.com/carltongibson/django-filter/blob/2.0.0/django_filters/filterset.py#L223

rest_framework_json_api/filters/filter.py
```diff
@@ -0,0 +1,135 @@
...
+            # rewrite filter[field] query params to make DjangoFilterBackend work.
+            data = request.query_params.copy()
+            for qp, val in data.items():
+                m = self.filter_regex.match(qp)
+                if m and (not m.groupdict()['assoc'] or
+                          m.groupdict()['ldelim'] != '[' or m.groupdict()['rdelim'] != ']'):
+                    raise ValidationError("invalid filter: {}".format(qp))
+                if m and qp != self.search_param:
+                    if not val:
+                        raise ValidationError("missing {} test value".format(qp))
```
sliverc: not so sure on this validation error. What about when I want to filter for all values with blank string?

requirements-development.txt
n2ygk: I couldn't add the dependency to setup.py due to the requirement for 1.1 for Py27 vs. 2.0 for Py3. I don't think there's a way to conditionalize it there. That's why it's in tox.ini. django-polymorphic doesn't require a back-level version for py27.

rest_framework_json_api/filters/filter.py
n2ygk: See my more general question about upstream naming in https://github.com/django-json-api/django-rest-framework-json-api/issues/467#issuecomment-416995999

rest_framework_json_api/filters/filter.py
n2ygk: Oh you mean instead of carrying around `self.filter_keys`? Yeah, cool.

rest_framework_json_api/filters/filter.py
n2ygk: I think as long as upstream DRF and Django support py27 we should as well. django-filter has not completely dropped py27 yet; they've documented it [here in the README](https://github.com/carltongibson/django-filter/blame/master/README.rst#L27): 
> If you need to support Python 2.7 use the version 1.1 release.

All the unit tests pass for both versions. That tells me there might be more advanced usage in 2.0 that I'm not testing for, but it seems pretty solid given the number of tests.

rest_framework_json_api/filters/filter.py
n2ygk: See [test_filter_empty](https://github.com/django-json-api/django-rest-framework-json-api/pull/466/files#diff-313750acd6d91887834a573075e6dcb8R155) which uses `r'^$'` or you an add a custom filter class as described in [filtering by empty values](https://django-filter.readthedocs.io/en/master/guide/tips.html?highlight=empty#filtering-by-empty-values) and [filtering by an empty string](https://django-filter.readthedocs.io/en/master/guide/tips.html?highlight=empty#filtering-by-an-empty-string) in the django-filter docs.

In django-filter "_empty values are interpreted as a skipped filter._" I thought it better to complain than to return an incorrect result as JSON:API in general follows a pattern of returning errors rather than silently doing the wrong thing. I can see exactly this use case of wanting to filter for an empty string and instead getting no filter at all.

rest_framework_json_api/filters/filter.py
n2ygk: Thanks. Will look at the examples, although this sounds like I'm losing my consistent naming of all JSONAPI filters being importable from `.filters`.

rest_framework_json_api/filters/filter.py
n2ygk: Somehow I thought he dropped `filter_queryset` between 1.1 and 2.0 so I was overriding the 1.1 version of it. Now I see it's there in 2.0 as well, but not the one you reference but rather this one: https://github.com/carltongibson/django-filter/blob/2.0.0/django_filters/rest_framework/backends.py#L89-L96

I didn't need to override it in 2.0 because it just calls `get_filterset` which I did override. But in 1.1 there's no call to get_filterset: https://github.com/carltongibson/django-filter/blob/1.1.0/django_filters/rest_framework/backends.py#L50-L56

So I'll have to fix this to make sure it doesn't break 2.0! Probably just check the [VERSION]( https://github.com/carltongibson/django-filter/blob/1.1.0/django_filters/__init__.py#L34) and `return super(...)` if > 1.1.

n2ygk:
@sliverc thanks for the review! I believe I've adequately resolved most items. Please feel free to unresolve if you disagree.

I have left one open item which has to do with naming style. I think we need to clarify that further.

rest_framework_json_api/filters/filter.py
n2ygk: I guess pushing outdated this open review item. That was not my intention:-(

n2ygk: Per https://github.com/django-json-api/django-rest-framework-json-api/issues/467#issuecomment-417338257 I've renamed to `DjangoFilterBackend` in 51b99464a7061f0c542c5eca42dfcf9fffc25467. This follows the upstream naming and is distinguished from it because it is imported from `rest_framework_json_api_filters`.

n2ygk:
@sliverc If you agree with https://github.com/django-json-api/django-rest-framework-json-api/issues/467#issuecomment-417338257 then I *think* this is ready for final review and merge.

If you disagree I can revert 51b9946 and we can work it out.

rest_framework_json_api/filters/filter.py
n2ygk: I'm going to resolve this based on our discussion about naming in #471 
@sliverc let me know if you are OK and please merge.

rest_framework_json_api/filters/filter.py
sliverc: I see your point. :+1:

rest_framework_json_api/filters/__init__.py
```diff
@@ -0,0 +1,6 @@
+import pkgutil
+from .sort import JSONAPIOrderingFilter  # noqa: F401
+# If django-filter is not installed, no-op.
+if pkgutil.find_loader('django_filters') is not None:
+    from .django_filter import DjangoFilterBackend  # noqa: F401
+del pkgutil
```
sliverc: Do you know any popular Django/DRF library which supports 3rd party libraries this way?

All the ones I know create a separate module/folder for 3rd party dependencies. This actually includes Django Filters (separate folder for rest_framework) or also [graphene-django](https://github.com/graphql-python/graphene-django).

I would really prefer it when we also create its own module for 3rdParty application where necessary so in this case `django_filters`.

rest_framework_json_api/filters/filter.py
sliverc: :+1:

requirements-development.txt
sliverc: :+1:  Makes sense. Let's leave this out then.

rest_framework_json_api/filters/__init__.py
n2ygk: OK, I can move it.

sliverc: Interesting, that's where you found this code bit...

When looking at the discussion why this was introduced is that following code will work:
```python
import django_filters
django_filters.rest_framework.DjangoFilterBackend()
```

Instead if this import mechanism were not there following would need to be done to use it:
```python
import django_filters.rest_framework
django_filters.rest_framework.DjangoFilterBackend()
# or
from django_filters import rest_framework
rest_framework.DjangoFilterBackend
```

This is an API design decision by django filters that one import statement is enough to auto import all modules. We at this point do not have such an auto import mechanism in DJA and I would prefer to leave it this way.

Overall Django Filters still has `rest_framework` specific classes in its own module. As outlined above we should do the same as this way it is clear by the code that in this directory a dependency is needed to be installed.

And as we do not have the same API decision as django filters we also do not need `pkgutil`.

n2ygk: @sliverc Take a look and see if this is OK. I made a `rest_framework_json_api.django_filters` next to `rest_framework_json_api.filters`. The former contains `DjangoFilterBackend` in `backends.py` and the latter current contains `OrderingFilter` but will add `QueryValidationFilter` (based off standard DRF filters) in a future PR.

PR #469, opened 2018-08-31
revert JSONAPI prefix to JsonApi for paginators.

### LINKED ISSUE #471: naming style: JSONAPI prefix or not?
[This started as a misplaced comment thread in #467. Let's move it here.]

I have a naming question: Do we prefix everything with `JSONAPI` or override existing names in DRF? We now have both styles:
- DJA serializers override DRF's `HyperlinkedModelSerializer` and so on. The advantage is code written for DRF can upgrade to DJA by simply changing an import. The disadvantage is that it's easy to "forget" to change the import from `rest_framework` to `rest_framework_json_api`.
- DJA paginators are named `JSONAPIPageNumberPagination` instead of overriding DRF's `PageNumberPagination`. The advantage is that it's very obvious that this is DJA-specific. The disadvantage is that instead of changing one import, you have to change each class use instance.

The approach here could be either:
- [x] DJA `DefaultRouter` overriding DRF `DefaultRouter`. 
- [ ] DJA `JSONAPIDefaultRouter`

---

@sliverc replied:
> Yeah I know I have also noticed. As we had to make backwards incompatible changes for pagination we started to add the JSONAPI prefix to work around this. There were also other cases in the past where this was done.

> Actually I would prefer not to use `JSONAPI` prefix as it is already clear by the rest_framework_json_api module that it is imported from DJA. As you outlined this way someone can also easily switch between DRF and DJA version by simply changing the import.

> However this won't be easy to accomplish to go that direction to make it consistent and backwards compatible. Potentially we have to break this at some point.

> But for new things (actually haven't thought about that when reviewing the filter backends) we can easily leave out the `JSONAPI` prefix.

> What is your preference?

---

I replied:

>> As we had to make backwards incompatible changes for pagination

> I don't think there was actually a backwards compatibility issue here as if one wants to be backward compatible they would just use `rest_framework.pagination.PageNumberPagination`. (Likewise for LimitOffset.)

>> Potentially we have to break this at some point.

> I say sooner than later. In the last release (2.5.0), we used the prefix `JsonApi` for paginators which is already deprecated by #463. The backend filters are not yet in a released version so have never been "seen" by their JSONAPI-prefixed names.

> The only other instance I was able to find was `rest_framework_json_api.metadata.JSONAPIMetadata` which extends `rest_framework.metadata.SimpleMetadata`; This could easily be renamed (with deprecation) to `rest_framework_json_api.metadata.SimpleMetadata` which would more clearly show the inheritance as well.

> The only other prefix I saw like this is `JSONAPIMeta` serializer inner class which I think is required to stay that way to distinguish it from the `Meta` inner class. 

>> What is your preference?

> My preference is to put in the effort now:
> - [x] drop the JSONAPI prefix from #466 before merging it. We have an open question (https://github.com/django-json-api/django-rest-framework-json-api/pull/466#discussion_r213601201) around naming it as well.
> - [ ] new PR to back out the prefix changes done in #463 and eliminate the JsonApi prefix and add deprecation warnings for `page` vs. `page[number]` and `page_size` vs. `page[size]`. Or just document this as a breaking change: `page` and `page_size` were never correct JSON:API usage.
> - [ ] new PR to rename/deprecate `JSONAPIMetadata`.

> Do you agree?

---

@sliverc wrote:

> There is also `JSONAPISettings` which we could rename to `APISettings`.

> Also I have noticed that we are kind of getting into the habit to misuse some issues to discuss things which might be related but not actually the same issue.

> I think for outsiders to better follow and to better track issues (e.g. a PR should solve one issue and not a comment of an issue) it would be better to create new issues for this (in this case a new issue where discussion happens on JSONAPI prefix).

> I will also follow up on a issue for versioning and deprecation policy.

n2ygk:
Fixes #471

## Description of the Change

Reverts JSONAPI prefix to JsonApi prefix on paginators as in release 2.5.0. Will replace without
the JsonApi prefix in release 3.0. Deprecation warnings added now.

## Checklist

- [x] unit-test added
- [x] documentation updated (including a few missed updates for other stuff in README.rst)


sliverc:
Let's try to follow common practices concerning deprecation and versioning.

I haven't had the time to document this but I think as DJA we should follow [Semver](https://semver.org/) and implement a similar deprecation policy like [Ember](https://www.emberjs.com/) (which is most likely the most used JS framework with JSON API spec).

This way we have a simple migration guide for our users. We can simply say, upgrade to the latest minor version of a series (e.g. 2.6.0) and solve all deprecation warnings. In version 3.0.0 we simply remove all deprecation warnings and do nothing else (real changes follow then in a version 3.1.0 etc.).

What this mean for our pagination:
* We keep the naming of `JSONAPIPageNumberPagination` till version 2.6.0
* In version 3.1.0 (or we could think of doing it in 3.0.0 as well) we make `JSONAPIPageNumberPagination` deprecated and change logic of `PageNumberPagination` to `JSONAPIPageNumberPagination` is today

This way we do not have to introduce a new arbitrary setting and have a simple migration guide for our DJA users.

What do you think?

n2ygk:
I was assuming we were following semver.

Since 2.5.0 released `JsonApi` as the prefix,  I think that should be what remains rather than an intermediate prefix of `JSONAPI`, which has not made it to a release yet. This involves reverting cfe89ea from master I believe.

Restating what you said, In addition we should add deprecation warnings now (presumably for a forthcoming 2.6.0 release) that indicate that in 3.0 or 3.1:
- PageNumberPagination will change the default from `page` and `page_size` to `page[number]` and `page[size]`, effectively doing what JsonApiPageNumberPagination does now.
- LimitOffsetPagination will change the default `max_limit` from `None` to `100`. Ditto.
- JsonApiPageNumberPagination will go away, replaced by `PageNumberPagination`.
- JsonApiLimitOffsetPagination will go away, replaced by `LimitOffsetPagination`.

I can revert the `JSON_API_STANDARD_PAGINATION` and so on in this PR that do more than the above.

Do you agree?

sliverc:
This sounds good. :+1:

rest_framework_json_api/pagination.py
```diff
@@ -100,63 +116,34 @@ def get_paginated_response(self, data):
...
 
     def __init__(self):
-        warnings.warn(
-            'PageNumberPagination is deprecated. Use JSONAPIPageNumberPagination '
-            'or create custom pagination. See '
-            'https://django-rest-framework-json-api.readthedocs.io/en/stable/usage.html#pagination',
-            DeprecationWarning)
+        if self.page_query_param == 'page' or self.page_size_query_param == 'page_size':
+            warnings.warn(
+                'PageNumberPagination will change in release 3.0 to default to use page[number] and'
```
sliverc: For deprecation warnings two things which I think are important:
1. Warning should be fixable at point of occurrence (so user should not wait till a new release to be able to fix a deprecation warning)
2. Warning should document how to fix them

To make it the case here we should explain that user should use custom pagination classes and they should define `page_query_param` and `page_size_query_param`.

For the deprecation warning it should only occur if the `type(self) == PageNumberPagination` or `type(self)`.__dict__ doesn't include `page_query_param` and `page_size_query_param`. (this codes bits are just out of my head and are not tested but I hope they explain what I mean.)

rest_framework_json_api/pagination.py
```diff
@@ -100,63 +116,34 @@ def get_paginated_response(self, data):
...
     max_limit = None
 
     def __init__(self):
-        warnings.warn(
-            'LimitOffsetPagination is deprecated. Use JSONAPILimitOffsetPagination '
-            'or create custom pagination. See '
-            'https://django-rest-framework-json-api.readthedocs.io/en/stable/usage.html#pagination',
-            DeprecationWarning)
+        if self.max_limit is None:
+            warnings.warn(
```
sliverc: same as with warning in PageNumberPagination

rest_framework_json_api/pagination.py
n2ygk: Yes, thanks. I understand and will make it work.

n2ygk:
@sliverc I think I've done the PendingDeprecationWarnings correctly using `self.__dict__` and also tried to document how to prevent getting caught by the deprecations.

sliverc:
Thanks. The deprecation warnings look good now.

I tried to put myself into the shoes of an user and have a feeling they won't understand why they have to implement custom classes. We would certainly not recommend for them to do it unless they use the old parameters `page` and `page_number`.

To simplify the upgrade procedure for an existing user I think we should at this point not deprecate `JsonApiPageNumberPagination`. This way the documentation can state to move to `JsonApiPageNumberPagination`. Only if someone wants to keep the old parameters do they need to implement their custom pagination.

In the next release we will then deprecate `JsonApiPageNumberPagination` and say to move to `PageNumberPagination`.

Much cleaner for a user to understand what is happening.

Also how to solve deprecation needs to be part of the CHANGELOG and not hidden in the documentation.

I haven't written any detailed review because I think there might be a bit too many back and forth. Is it OK if I push to this PR? Would be more efficient and then you can look whether you agree with the changes.

n2ygk:
Yes. Please do.

sliverc:
@n2ygk I have adjusted the code and clarified changelog how to migrate. Does it look ok to you?

Do you see anything else which we need to do before releasing 2.6.0? Otherwise would I suggest we do a release and just short after follow up with a 3.0.0 release where we remove all deprecation warnings and deprecate JsonApiPageNumberPagination and JsonApiLimitOffsetPagination + changing defaults of PageNumberPaginiation and LimitOffsetPagination.

P.S.
We should squash merge this PR to keep a clear history in master

n2ygk:
@sliverc This looks good and ready for you to squash merge.

Before a 2.6.0 release, please let me submit two more PRs:
1) QueryParameterValidation filter backend which is a companion to the others. This just enforces that only JSON:API-defined query parameters are allowed.
2) Documentation/examples update to show use of `rest_framework.filters.SearchFilter` with `api_settings.SEARCH_PARAM` = `filter[search]`
That should fully round out all the query param-based stuff.

I will want to get to some new exception classes that properly format the errors object but that can wait.

sliverc:
Thanks. Merged.

I would prefer that we don't wait for a release concerning point 1. This can be added as fully backwards compatible change at any time later one. Point 2 makes certainly sense to document before a release.

PR #474, opened 2018-09-10
Move to keep a changelog format

### LINKED ISSUE #470: Introduce versioning and deprecation policy
Currently this is not documented so it is not clear what versioning and deprecation policy DJA follows.

This needs to be documented. I would suggest we follow common standards, using [Semver](https://semver.org/) and implement a similar deprecation policy like [Ember](https://www.emberjs.com/) (which is most likely the most used JS framework with JSON API spec).

This way we have a simple migration guide for our users. We can simply say, upgrade to the latest minor version of a series (e.g. 2.6.0) and solve all deprecation warnings. In version 3.0.0 we simply remove all deprecation warnings and do nothing else (real changes follow then in a version 3.1.0 etc.).

Related to #467 #469

sliverc:
Fixes #470

Includes documentation of versioning and deprecation policy.

This change will also help to better document deprecation changes like currently proposed in https://github.com/django-json-api/django-rest-framework-json-api/pull/469

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,11 +21,6 @@ any parts of the framework not mentioned in the documentation should generally b
 
 * Replaced binary `drf_example` sqlite3 db with a [fixture](example/fixtures/drf_example.json). See [getting started](docs/getting-started.md#running-the-example-app).
 
-### Deprecated
-
-* For naming consistency, renamed new `JsonApi`-prefix pagination classes to `JSONAPI`-prefix.
-  * Deprecates `JsonApiPageNumberPagination` and `JsonApiLimitOffsetPagination`
-
 ### Fixed
 
 * Performance improvement when rendering relationships with `ModelSerializer`


CHANGELOG.md
```diff
@@ -1,94 +1,155 @@
-[unreleased]
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+Note that in line with [Django REST Framework policy](http://www.django-rest-framework.org/topics/release-notes/),
```
n2ygk: One concern I have is that DRF does not follow semver:

> Medium version numbers (0.x.0) may include API changes, in line with the deprecation policy. You should read the release notes carefully before upgrading between medium point releases.

Where semver requires only new backwards-compatible changes.

As DJA is an extension of DRF, should we follow DRF's versioning or be more explicit that we instead follow semver?

CHANGELOG.md
```diff
@@ -1,94 +1,155 @@
...
 * Add related urls support. See [usage docs](docs/usage.md#related-urls)
-* Replaced binary `drf_example` sqlite3 db with a [fixture](example/fixtures/drf_example.yaml). See [getting started](docs/getting-started.md#running-the-example-app).
+* Add optional [jsonapi-style](http://jsonapi.org/format/) filter backends. See [usage docs](docs/usage.md#filter-backends)
+
+### Changed
+
+* Replaced binary `drf_example` sqlite3 db with a [fixture](example/fixtures/drf_example.json). See [getting started](docs/getting-started.md#running-the-example-app).
+
+### Deprecated
+
```
n2ygk: I believe we've decided to yank this PR before a release ever happened, so I don't think it needs to be listed as deprecated.

CHANGELOG.md
sliverc: let's use semver which is commonly accepted.

CHANGELOG.md
sliverc: Fair enough. Removed.

PR #476, opened 2018-09-13
Document how to use `rest_framework.filters.SearchFilter`

### LINKED ISSUE #432: spec compliance: sparse fieldsets, sorting & filtering query parameters

n2ygk:
Fixes #432 

## Description of the Change
Documents how to use `rest_framework.filters.SearchFilter` along with the other filter backends.

Includes updates to the example app to demonstrate usage and test cases.

CHANGELOG.md remains unchanged because this is included in: "Add optional [jsonapi-style](http://jsonapi.org/format/) filter backends"

I also documented some `filterset_fields` to complement the `search_fields` in the usage document showing how to configure filter backends.

## Checklist

- [x] unit-test added
- [x] documentation updated

### DIFF
--- a/example/tests/test_filters.py
+++ b/example/tests/test_filters.py
@@ -338,3 +338,33 @@ class DJATestFilters(APITestCase):
         dja_response = response.json()
         self.assertEqual(dja_response['errors'][0]['detail'],
                          "missing filter[headline] test value")
+
+    def test_search_keywords(self):
+        """
+        test for `filter[search]=keyword1...` (keyword1 [AND keyword2...])
+        """
+        for keywords in ("research", "chemistry", "nonesuch",
+                         "research seminar", "research nonesuch",
+                         "barnard classic"):
+            response = self.client.get(self.url, data={'filter[search]': keywords})
+            self.assertEqual(response.status_code, 200, msg=response.content.decode("utf-8"))
+            dja_response = response.json()
+            # see the search_fields defined in views.py.
+            a = {}
+            b = {}
+            c = {}
+            d = {}
+            keys = keywords.split()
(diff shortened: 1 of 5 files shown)


example/tests/test_filters.py
```diff
@@ -338,3 +338,33 @@ def test_filter_missing_rvalue_equal(self):
         dja_response = response.json()
         self.assertEqual(dja_response['errors'][0]['detail'],
                          "missing filter[headline] test value")
+
+    def test_search_keywords(self):
+        """
+        test for `filter[search]=keyword1...` (keyword1 [AND keyword2...])
+        """
+        for keywords in ("research", "chemistry", "nonesuch",
```
sliverc: This test is a bit difficult to comprehend. It would be much easier to write it with [pytest parametrize](https://docs.pytest.org/en/latest/parametrize.html).

I ask myself though do we really need a test for SearchFilter? It is DRF functionality and as such it is already tested in DRF itself and we just do same work again. Hence I think we could remove this test.

n2ygk:
@sliverc My reason for the test is because it is testing(demonstrating) the
example app’s usage.

sliverc:
@n2ygk if this is the case then the test should be simplified where only one keyword is searched for and a simple `assert result.json() == expected` will do. This will show how it works and is readable by someone who wants to see how search filter is used.

n2ygk:
@sliverc Sorry, I didn't notice your followup comment until just now... Please take a look at the test case again before we kill it:-(.  I am also happy to add the result.json() assertion with static test values but still think it's valuable to demonstrate the multi-keyword searching across ORM-related fields.

PR #477, opened 2018-09-13
renamed JSONAPIOrderingFilter to OrderingFilter

### LINKED ISSUE #471: naming style: JSONAPI prefix or not?

n2ygk:
Fixes #471

## Description of the Change
Remove "JSONAPI" prefix from `OrderingFilter` before it is released.

No CHANGELOG.md changes needed as it's covered by "Add optional [jsonapi-style](http://jsonapi.org/format/) filter backends. See [usage docs](docs/usage.md#filter-backends)"

## Checklist

- [x] documentation updated

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -173,7 +173,7 @@ override ``settings.REST_FRAMEWORK``
         ),
         'DEFAULT_METADATA_CLASS': 'rest_framework_json_api.metadata.JSONAPIMetadata',
         'DEFAULT_FILTER_BACKENDS': (
-            'rest_framework_json_api.filters.JSONAPIOrderingFilter',
+            'rest_framework_json_api.filters.OrderingFilter',
             'rest_framework_json_api.django_filters.DjangoFilterBackend',
         ),
         'TEST_REQUEST_RENDERER_CLASSES': (
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -33,7 +33,7 @@ REST_FRAMEWORK = {
     ),
     'DEFAULT_METADATA_CLASS': 'rest_framework_json_api.metadata.JSONAPIMetadata',
     'DEFAULT_FILTER_BACKENDS': (
-        'rest_framework_json_api.filters.JSONAPIOrderingFilter',
+        'rest_framework_json_api.filters.OrderingFilter',
         'rest_framework_json_api.django_filters.DjangoFilterBackend',
     ),
     'TEST_REQUEST_RENDERER_CLASSES': (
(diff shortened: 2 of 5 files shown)


sliverc:
Well spotted. I have created a milestone 2.6.0 so if you find any other issue which should be fixed before release 2.6.0 you can assign milestone to it.

PR #481, opened 2018-09-17
query validation filter

n2ygk:
## Description of the Change

Rounds out the suite of JSON:API backend filters with one that checks for mandatory query parameter names and returns a 400 if one that doesn't match is seen, or if a param like `sort` is repeated. This makes for a more positive client developer experience in that incorrect query params (typos, etc.) are reported as errors rather than being silently ignored.

## Checklist

- [x] unit-test added
- [x] documentation updated

Changelog is already covered by "Add optional [jsonapi-style](http://jsonapi.org/format/) filter backends. See [usage docs](docs/usage.md#filter-backends)" from prior filters.

Commit history is pretty ugly because this was authored a while ago and then a bunch of new master commits were merged. This should be squashed when merged.


rest_framework_json_api/filters/queryvalidation.py
```diff
@@ -0,0 +1,39 @@
...
+
+class QueryValidationFilter(BaseFilterBackend):
+    """
+    A backend filter that performs strict validation of query parameters for
+    jsonapi spec conformance and raises a 400 error if non-conforming usage is
+    found.
+
+    If you want to add some additional non-standard query parameters,
+    simply override `.query_regex` adding the new parameters but, "with the additional
+    requirement that they MUST contain contain at least one non a-z character (U+0061 to U+007A).
```
sliverc: instead of repeating parts of the specification I prefer we only reference the jsonapi spec by link here. If someone wants to break the spec it is really up to them anyway.

rest_framework_json_api/filters/queryvalidation.py
```diff
@@ -0,0 +1,39 @@
+import re
+
+from rest_framework.exceptions import ValidationError
+from rest_framework.filters import BaseFilterBackend
+
+
+class QueryValidationFilter(BaseFilterBackend):
```
sliverc: So it is not confused with query set validation I prefer we are a bit more explicit with the name.

e.g. QueryParameterValidationFilterBackend

docs/usage.md
```diff
@@ -104,9 +105,32 @@ class MyLimitPagination(JsonApiLimitOffsetPagination):
 
 ### Filter Backends
 
-Following are descriptions for two JSON:API-specific filter backends and documentation on suggested usage
+Following are descriptions for three JSON:API-specific filter backends and documentation on suggested usage
```
sliverc: just a detail but referencing the number of filters here sounds a bit like a overkill and we might forget changing it in the future when we add more.

Why not just:
`Following are descriptions of JSON:API-specific filter backends and documentation on suggested usage`

n2ygk:
@sliverc Thanks for letting me sneak this into 2.6.0!

PR #483, opened 2018-09-20
Release 2.6.0

sliverc:
Additional some small clarification of changelog.

@n2ygk 
I think we are ready for a new release. In case something is missing let me know.


n2ygk:
@sliverc PS: I assume you'll push the tag and twine, etc.?

PR #488, opened 2018-10-08
Don't render empty objects on relationships

robertobarreda:
## Description of the Change

REGRESION (version 2.6)

In previous versions, OneToOneFields without Hyperlink settings in its Serializer doesn't render anything.

In version 2.6, an empty JSON object is rendered.

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/rest_framework_json_api/renderers.py
+++ b/rest_framework_json_api/renderers.py
@@ -143,7 +143,7 @@ class JSONRenderer(renderers.JSONRenderer):
             if isinstance(field, HyperlinkedMixin):
                 field_links = field.get_links(resource_instance, field.related_link_lookup_field)
                 relation_data.update({'links': field_links} if field_links else dict())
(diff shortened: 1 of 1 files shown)


sliverc:
Thanks for your contribution.

The specification actually states to return `null for empty to-one relationships.`
See http://jsonapi.org/format/#document-resource-object-linkage

So I think this means this is actually a bug rather than a regression - the bug is just different in previous versions where relationship is not rendered at all and now where it is rendered as empty dict.

So I would prefer when the PR could be adjusted to return null instead of skipping it. A test would also need to be added so we can make sure that this bug doesn't occur again in the future.

robertobarreda:
IMHO I'd pass this fix (be consistent with the previous behaviour) and do another PR to fix the rendering according to the spec.

sliverc:
Closing in favor of #526

PR #490, opened 2018-10-09
Added rendering tests using default drf serializers

Alig1493:
Fixes #

## Checklist

- [x] unit-test added
- [ ] documentation updated


Alig1493:
@sliverc Here's the initial progress.

sliverc:
Thanks @Alig1493 This looks like a good start.

The test you are covering here is based on rendering. I think it will also be important to test parsing with DRF serializers resp. the full request/response cycle. This could be something like `test_model_viewsets.py` but with DRF classes.

Alig1493:
Alright I'll go take a look.

Alig1493:
@sliverc I had some issues regarding testing the response of the default drf serializer urls:
The `test_model_viewsets.py` file tested urls that called the viewsets which had json-api serializers.
Does this mean I'll be making my own custom viewsets and register it with the urls as new endpoints and test those new endpoints?

I took the approach of making new serializers for the Blog model using the default serializer and adding it as a viewset in urls. I didn't make new files for these and added the code to the existing code base. Should I push into the PR for you to check?

Thanks.

Alig1493:
@sliverc pushed.

example/serializers.py
```diff
@@ -50,6 +58,25 @@ class Meta:
         meta_fields = ('copyright',)
 
 
+# DRF default serializer to test default DRF functionalities
+class BlogDRFSerializer(serializers.ModelSerializer):
```
sliverc: shouldn't this be derived from drf_serilazers.ModelSerializer as well? Also best move the comment into the class a block comment.

example/serializers.py
```diff
@@ -27,6 +28,13 @@ class Meta:
         fields = ('tag',)
 
 
+# DRF default serializer to test default DRF functionalities
+class TaggedItemDRFSerializer(drf_serilazers.ModelSerializer):
```
sliverc: I prefer comments to be block comments of the class itself using `""" """`

example/tests/unit/test_default_drf_serializers.py
```diff
@@ -73,3 +76,24 @@ def test_render_format_keys(settings):
 
     result = json.loads(rendered.decode())
     assert result['data']['attributes']['json-field'] == {'json-key': 'JsonValue'}
+
+
+class TestDRFBlogViewSet(TestBlogViewSet):
```
sliverc: as all other tests are written in pytest in this module I would write these tests the same way. This way we do not have a dependency on TestBlogViewSet (which basically only creates an entry during setup phase anyway).

example/tests/unit/test_default_drf_serializers.py
```diff
@@ -73,3 +76,24 @@ def test_render_format_keys(settings):
...
+                'id': '{}'.format(self.blog.id),
+                'links': {'self': 'http://testserver/blogs/{}'.format(self.blog.id)},
+                'meta': {'copyright': 2018},
+                'relationships': {'tags': {'data': []}},
+                'type': 'blogs'
+            },
+            'meta': {'apiDocs': '/docs/api/blogs'}
+        }
+        got = resp.json()
+        print(got)
```
sliverc: left over print statement? I guess should be removed.

example/tests/unit/test_default_drf_serializers.py
```diff
@@ -0,0 +1,99 @@
+import json
+
+from django.urls import reverse
```
sliverc: Imports are in invalid order according to isort.

See https://travis-ci.org/django-json-api/django-rest-framework-json-api/jobs/440153239#L936

example/tests/unit/test_default_drf_serializers.py
```diff
@@ -0,0 +1,125 @@
...
+    rendered = render_dummy_test_serialized_view(DummyTestViewSet)
+
+    result = json.loads(rendered.decode())
+    assert result['data']['attributes']['json-field'] == {'json-key': 'JsonValue'}
+
+
+@pytest.mark.django_db
+def test_get_object_gives_correct_blog(client, blog, entry):
+
+    url = reverse('drf-entry-blog', kwargs={'entry_pk': entry.id})
```
Alig1493: @sliverc I have used reverse here.

example/tests/unit/test_default_drf_serializers.py
```diff
@@ -0,0 +1,125 @@
...
+    assert result['data']['attributes']['json-field'] == {'JsonKey': 'JsonValue'}
+
+
+def test_render_format_keys(settings):
+    """Test that json field value keys are formated."""
+    delattr(settings, 'JSON_API_FORMAT_FILED_NAMES')
+    settings.JSON_API_FORMAT_KEYS = 'dasherize'
+    rendered = render_dummy_test_serialized_view(DummyTestViewSet)
+
+    result = json.loads(rendered.decode())
```
Alig1493: @sliverc as for json here's the usage

Alig1493:
@sliverc apologies for taking so long to get back to you I became occupied with something else for some time. One thing that I was wondering is that if it would be possible for you to guide me as to how to make put/patch requests using modelviewsets. I've mostly used generic class based views, although that's no excuse. My initial attempt in trying to patch the `drf-entry-blog` url endpoint of mine keeps giving me method not allowed error.

Alig1493:
@sliverc Update: I've figured out how to patch using routers and model viewsets. Hopefully you'll be able to review it soon. My remaining question is that when posting data should I be only creating a Blog object, or will I have to link the blog object to the entry object to which it's related.

example/tests/conftest.py
```diff
@@ -62,3 +62,13 @@ def single_art_project(art_project_factory):
 @pytest.fixture
 def client():
     return APIClient()
+
+
+@pytest.fixture
+def blog():
```
sliverc: This is actually not needed as this is automatically done with pytest-factoryboy where this factory is [registered](https://github.com/django-json-api/django-rest-framework-json-api/pull/490/files#diff-b9e1915dc58c0e683db3b4b7a7a7c365R18).

example/tests/unit/test_default_drf_serializers.py
```diff
@@ -0,0 +1,188 @@
...
+        },
+    }
+
+    resp = client.post(url, request_data)
+
+    # look for created blog in database
+    blog = Blog.objects.filter(name=name)
+
+    # check if blog exists in database
+    assert blog.exists()
```
sliverc: I guess this asserts is not needed as the next asserts checks already whether it exists by counting it.

example/tests/unit/test_default_drf_serializers.py
```diff
@@ -0,0 +1,188 @@
...
+    assert blog.count() == 1
+
+    # get created blog from database
+    blog = blog[0]
```
sliverc: `blog.first()` might be more elegant.

Alig1493: OF COURSE. Completely missed that train of thought.

example/tests/unit/test_default_drf_serializers.py
Alig1493: Good point.

example/tests/conftest.py
Alig1493: Oh that's my bad on not seeing the registered factories.

Alig1493:
@sliverc I have made changes as per your review. Please have a look.

sliverc:
Looks good.

It needs to be rebased to current master and CI isort issue https://travis-ci.org/django-json-api/django-rest-framework-json-api/jobs/441354640 fixed and the in should be ready.

Alig1493:
Isort is a package that sorts imports alphabetically. So I just have to fix my order of imports?

Alig1493:
@sliverc Travis CI seems to be stuck. Could you please give more insight on this issue? :s

Alig1493:
@sliverc it seems all's well for now.

Alig1493:
I'd love to extend my help if possible. I'll take a look and ask if there's any questions.

PR #500, opened 2018-10-23
Format field names for metadata (OPTIONS request)

mstdokumaci:
## Description of the Change
Field names are represented wrong in OPTIONS call if you use `JSON_API_FORMAT_FIELD_NAMES` in settings. This patch fixes metadata.

## Checklist

- [ ] unit-test added
- [ ] documentation updated


rest_framework_json_api/metadata.py
```diff
@@ -83,7 +84,7 @@ def get_serializer_info(self, serializer):
         serializer.fields.pop(api_settings.URL_FIELD_NAME, None)
 
         return OrderedDict([
-            (field_name, self.get_field_info(field))
+            (format_value(field_name, json_api_settings.FORMAT_FIELD_NAMES), self.get_field_info(field))
```
sliverc: Simply use `format_value` without assigning a format type. This way it will choose the right one (also in case some still uses FORMAT_KEYS)

```suggestion
            (format_value(field_name), self.get_field_info(field))
```

sliverc:
Closing as implemented in #713

PR #503, opened 2018-10-24
Use pyupio to manage development dependencies

sliverc:
## Description of the Change

This way CI doesn't suddenly break when a dependency has updated (e.g. #501) but we still keep up-to-date.

pyupio will open a PR with the updated dependencies once they are released. 

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/.pyup.yml
+++ b/.pyup.yml
@@ -1,0 +1,5 @@
+search: False
+requirements:
+  - requirements-development.txt:
+      update: all
+      pin: True


sliverc:
@jerel I have requested permission on org django-json-api so pyup has access. Could you grant pyupio the needed access?

jerel:
@sliverc I granted it shortly after it came in. Should be working now.

sliverc:
@n2ygk 
Rebased to master and CI is green now.

Issue #496, opened 2018-10-15
Missing permission checks in RelatedMixin views


lcary:
No object level or model level permissions appear to be checked for related models in `RelatedMixin` views.

Expected behavior: The related fields should be empty in a response object if the user doesn't have view permission to the related field. Alternatively, if a response must render the related object, a permissions error should be raised.

Observed behavior: Currently, if the user has permission for a model but not related models (e.g. via a foreign key relation), they are still able to view the related object if a `RelatedMixin` view's serializer has a `ResourceRelatedField`. This seems to merit at least some documentation warning users about the potential for leaking data, but preferably this issue could be fixed by enabling permissions checking.

Reproducing the issue requires a project that checks that a user has read (or "view") permissions for models they try to access (e.g. using extended versions of `DjangoObjectPermissions` and `DjangoModelPermissions` classes mentioned in the [django-rest-framework permissions documentation](https://www.django-rest-framework.org/api-guide/permissions/)). In such a project, for any `RelatedMixin` view that uses a serializer with a `ResourceRelatedField`, grant the user permission to the serializer's model object, but do not grant the user access to the related model. If the user requests the object, they will still be able to access the related field. I'm happy to add a test case for this issue.

Perhaps object level or even model level permissions checking is out of scope for the django-rest-framework-json-api project, but it seems like both should be supported. After all, django-rest-framework supports and advertises both via the `DjangoObjectPermissions` and `DjangoModelPermissions` classes. In a project I'm working on, we use django-guardian with a similar class to check object-level permissions. Perhaps I can override `RelatedMixin.get_related_instance()` in the meantime as a stopgap solution, but a solution via configuration flag or a default in this library would be preferable.

Best,
Luc

sliverc:
Fair point. I guess `self.check_object_permissions(self.request, obj)` should be run in `get_related_instance` before it returns instance. I guess this would solve your issue?

I think a hint in the documentation should not that `permission_classes` on views used with relation links should be aware of different object types. This should already be the case with `DjangoObjectPermissions` and `DjangoModelPermissions` classes.

A PR would be welcome.

n2ygk:
@sliverc I'll take this and add to my open PR (#492) for now as this is touching the same RelatedMixin code (and @lcary and I are coworkers;-). Once we figure it out we can split out into separate PRs if warranted. Does that make sense?

n2ygk:
[...]

PR #504, opened 2018-10-25
Added related fields filtering

Anton-Shutik:
Fixes #
This simple PR adds fitlering for related querysets. Let's say we have an url like /api/order/1/items/  and I want for filter the items by `available=true`. There is no way to do it now.
With this PR we can do this by adding filter backend to the OrderviewSet:

```python
class OrderViewSet(ModelViewset):
    related_filter_backends = {'items': [IsAvailableFilter]}
```
That's it.
Filtering work exactly same way as standard DRF's filtering.

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/rest_framework_json_api/views.py
+++ b/rest_framework_json_api/views.py
@@ -108,6 +108,8 @@ class RelatedMixin(object):
     This mixin handles all related entities, whose Serializers are declared in "related_serializers"
     """
 
+    related_filter_backends = {}
+
     def retrieve_related(self, request, *args, **kwargs):
         serializer_kwargs = {}
         instance = self.get_related_instance()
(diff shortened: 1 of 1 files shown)


n2ygk:
This might help address #496

Anton-Shutik:
@n2ygk I think something like  [that](https://github.com/django-json-api/django-rest-framework-json-api/pull/505) will help better . Didn't tested, just an idea

Anton-Shutik:
For this PR it just allows to filter related querysets. 
What do you think about it ? If it looks good for you I can added some tests and update the docs.
Thanks

n2ygk:
@Anton-Shutik Actually this might be better than returning an error as you do in #505 if the goal is to reduce the related queryset to those allowed to be returned. Basically the `related_filter_backends` is a permission filter that only returns items that are permitted to be returned.

The goal I think is to silently not return related data when not permitted (make it invisible) rather than have the request fail. Does that make sense @lcary?

Anton-Shutik:
@n2ygk `related_filter_backends` is just **for filtering**. It should not check any permissions or something. It is made same way like DRF does. Filters and permissions are separate things. Let's do not mix it. If you want to hide some items from current customer - add a `IsNotHiddenItemFilter` and it will just filter out all items that the customer shouldn't see. It you want to check permissions for a given endpoint (or resource) add a `IsStuffUser`(for example) and that will return response only for stuff users. That's it.

Permissions check is a permission check. It should not do any filtering at all. And filtering is just filtering it should not check anything. That's why I've made 2 separate PRs for that.

sliverc:
Closing this for now as the discussion in #496 continues how the issue can be solved.

PR #526, opened 2018-12-04
Do not skip empty one-to-one relationships

sliverc:
Fixes #488

## Description of the Change

In case of a reverse one-to-one relationships `get_relation_instance` would swallow `DoesNotExist` exception leading renderer to skip field. Such field instead should be rendered as `null` and never be skipped.

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ any parts of the framework not mentioned in the documentation should generally b
 * Avoid patch on `RelationshipView` deleting relationship instance when constraint would allow null ([#242](https://github.com/django-json-api/django-rest-framework-json-api/issues/242))
 * Avoid error with related urls when retrieving relationship which is referenced as `ForeignKey` on parent
 * Do not render `write_only` relations
+* Do not skip empty one-to-one relationships
 
 
 ## [2.6.0] - 2018-09-20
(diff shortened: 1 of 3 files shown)


sliverc:
@n2ygk 
Changing renderer always has some risks, but making it better conform to JSON api there is no way around it :smile: 

However good testing is always a good thing. I have therefore test run this in the projects I am using DJA and it turned out green. If you have any projects to test run it in that would be great too, to lower the barrier that there is a regression in this PR.

n2ygk:
@sliverc Should I look at this now or wait until the test failures are resolved?

n2ygk:
@sliverc: I'm confused as this was already working per the spec for me. I've compared

`git+https://github.com/django-json-api/django-rest-framework-json-api.git@8c075d0` vs
`git+https://github.com/sliverc/django-rest-framework-json-api.git@null_relationships`

And both return an empty list for a to-many relationships and null for a to-one.

Is this only for `HyperlinkedIdentityField`? I don't use that in my code.

sliverc:
@n2ygk 
It's just in corner cases where a relationship is not rendered but skipped. It is basically only when `utils.get_relation_instance` cannot resolve to a relation instance.
The only known case where `utils.get_relation_instance` cannot resolve is with a reverse `OneToOneField` leading to a `DoesNotExist` exception.

When thinking about it again it is properly better just to address this `OneToOneField` issue which is known and do not touch the other only hypothetical errors which might actually never happen.

Hence I have simplified this PR and clarified it in comment and changelog. Best have a look again.

PR #529, opened 2018-12-06
Allow HyperlinkedRelatedField to be used with related urls

### LINKED ISSUE #521: HyperlinkedRelatedField is incompatible with HyperlinkedModelSerializer
If HyperlinkedRelatedField raises SkipField, HyperlinkedModelSerializer does not catch it, leading to a 500 Server Error. Is this a feature or a bug? If a feature, where is it documented?

housleyjk:
Fixes #521

## Description of the Change
The details in #521 are not clear, but I ran into a very similar `SkipField` issue when trying to use HyperlinkedRelatedFields and the RelatedMixin. The HyperlinkedRelatedField generates the correct url for the related link (e.g. /authors/1/entries), but visiting the url gives a 500 error because of an unhandled `SkipField` exception. In other words, without these changes, it is possible to give a related link in the payload, but that link won't actually work.

## Checklist

- [X] unit-test added (I couldn't get the unit tests to all work prior to making these changes, but I was able to trigger and fix this issue with the existing `test_retrieve_related_many()` test. So this code is covered by unit tests.)
- [X] NO documentation changes necessary, bug fix
- [X] changelog entry added to `CHANGELOG.md`
- [X] author name in `AUTHORS`

### DIFF
--- a/example/serializers.py
+++ b/example/serializers.py
@@ -206,7 +206,7 @@ class AuthorSerializer(serializers.ModelSerializer):
         self_link_view_name='author-relationships',
         queryset=AuthorBio.objects,
     )
-    entries = relations.ResourceRelatedField(
+    entries = relations.HyperlinkedRelatedField(
         related_link_view_name='author-related',
         self_link_view_name='author-relationships',
         queryset=Entry.objects,
--- a/example/serializers.py
+++ b/example/serializers.py
@@ -206,7 +206,7 @@ class AuthorSerializer(serializers.ModelSerializer):
         self_link_view_name='author-relationships',
         queryset=AuthorBio.objects,
     )
-    entries = relations.HyperlinkedRelatedField(
+    entries = relations.ResourceRelatedField(
         related_link_view_name='author-related',
         self_link_view_name='author-relationships',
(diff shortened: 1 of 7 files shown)


example/serializers.py
```diff
@@ -206,7 +206,7 @@ class AuthorSerializer(serializers.ModelSerializer):
         self_link_view_name='author-relationships',
         queryset=AuthorBio.objects,
     )
-    entries = relations.ResourceRelatedField(
+    entries = relations.HyperlinkedRelatedField(
```
sliverc: Not so ideal to change from `ResourceRelatedField` to `HyperlinkedRelatedField` as now the one but not the other will be tested.

Hence I think it would be better to add a new  `HyperlinkedRelatedField`. Maybe comments could be used?

housleyjk:
Ok, I've improved the testing for this feature as requested and done my best to verify that it is correct. I hope you will forgive me renaming `comment_set` -> `comments`. I thought it was more consistent with the related name `entries` already in use.

sliverc:
Thanks a lot @housleyjk I totally agree renaming `comments` to `comment_set` is much cleaner.

PR #535, opened 2018-12-28
Allow `format` through GET validation filtering

serenecloud:
## Issue

The guides on https://django-rest-framework-json-api.readthedocs.io/en/stable/usage.html#configuration recommend including the `rest_framework_json_api.filters.QueryParameterValidationFilter` in initial configuration which is good advice, but by default, the Django REST Framework UI provides a dropdown which allows you to pick `format=api` or `format=vnd.api+json`

screenshot at 2018-12-28 15-10-51

Without allowing the `format` parameter through you get the following error:

```
HTTP 400 Bad Request
Allow: GET, POST, HEAD, OPTIONS
Content-Type: application/vnd.api+json
Vary: Accept

{
    "errors": [
        {
            "detail": "invalid query parameter: format",
            "source": {
                "pointer": "/data"
            },
            "status": "400"
        }
    ]
}
```
## Description of the Change

This change adds `format` to `query_regex`

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/rest_framework_json_api/filters.py
+++ b/rest_framework_json_api/filters.py
@@ -69,7 +69,7 @@ class QueryParameterValidationFilter(BaseFilterBackend):
     the rules at http://jsonapi.org/format/#query-parameters.
     """
     #: compiled regex that matches the allowed http://jsonapi.org/format/#query-parameters:
-    #: `sort` and `include` stand alone; `filter`, `fields`, and `page` have []'s
+    #: `sort`, `include` and `format` stand alone; `filter`, `fields`, and `page` have []'s
     query_regex = re.compile(r'^(sort|include|format)$|^(filter|fields|page)(\[[\w\.\-]+\])?$')
 
     def validate_query_params(self, request):
--- a/rest_framework_json_api/filters.py
+++ b/rest_framework_json_api/filters.py
@@ -70,7 +70,7 @@ class QueryParameterValidationFilter(BaseFilterBackend):
     """
     #: compiled regex that matches the allowed http://jsonapi.org/format/#query-parameters:
     #: `sort` and `include` stand alone; `filter`, `fields`, and `page` have []'s
-    query_regex = re.compile(r'^(sort|include)$|^(filter|fields|page)(\[[\w\.\-]+\])?$')
+    query_regex = re.compile(r'^(sort|include|format)$|^(filter|fields|page)(\[[\w\.\-]+\])?$')
 
     def validate_query_params(self, request):
         """


sliverc:
Not sure how to understand https://jsonapi.org/format/#query-parameters whether it allows additional query parameters or not.

Another way would be to change [URL_FORMAT_OVERRIDE](https://www.django-rest-framework.org/api-guide/settings/#url_format_override).

@n2ygk What do you think?

sliverc:
JSON API version 1.1 Release candidate gets a bit more specific on query parameters, see https://jsonapi.org/format/1.1/#query-parameters-custom

This means for a valid implementation of format query parameter, `URL_FORMAT_OVERRIDE` would need to be set to `filter[format]`. This sounds a bit odd though but defining a [profile](https://jsonapi.org/format/1.1/#profiles) is properly a bit of a overkill.

n2ygk:
> Not sure how to understand https://jsonapi.org/format/#query-parameters whether it allows additional query parameters or not.
> 
> Another way would be to change [URL_FORMAT_OVERRIDE](https://www.django-rest-framework.org/api-guide/settings/#url_format_override).
> 
> @n2ygk What do you think?

It says you can have additional "non-standard" query parameters but with the "additional requirement that they MUST contain at least one non a-z character" in order to avoid naming collisions. It's their version of an `X-` header.

I personally don't like use of the `format=` query parameter despite what the browseable API offers (which is a carryover from DRF which has no rules about query parameters). The functionality can be achieve using `Accepts: application/vnd.api+json` header. Perhaps best to override `format` to be something containing a non a-z character? It's not really a JSON API compliance thing since it's just used by the Browseable API: an HTML response is not covered by JSON API and it's not a `filter`.

sliverc:
What we can do is to allow query parameter which set by `URL_FORMAT_OVERRIDE` in QueryParameterValidationFilter and add a recommendation to set `URL_FORMAT_OVERRIDE` to `contentFormat`.

sliverc:
@serenecloud Are you open to adjust your PR as [commented](https://github.com/django-json-api/django-rest-framework-json-api/pull/535#issuecomment-453937183)?

serenecloud:
@sliverc I'm not sure what the change involves at this point. Changing to `contentFormat` would require a change on the Django end to have this working, yes?

sliverc:
@serenecloud 
In the README.md there is a section how rest framework needs to be configured to work well with DJA. There the config option [URL_FORMAT_OVERRIDE](https://www.django-rest-framework.org/api-guide/settings/#url_format_override) would need to set to `contentFormat` as a recommendation. The QueryParameterValidationFilter would need to be adjusted not to hard code `format` query parameter but use the URL_FORMAT_OVERRIDE value.

sliverc:
@n2ygk I am a bit confused on your comment as you have already upvoted my [comment](https://github.com/django-json-api/django-rest-framework-json-api/pull/535#issuecomment-453937183) and I thought you agreed with it. Browsable API can be used in production and we even document it this way in the README.md so it needs to work.
FORMAT_SUFFIX_KWARG is only used for custom url patterns and what we are looking for is [URL_FORMAT_OVERRIDE](https://www.django-rest-framework.org/api-guide/settings/#url_format_override) and README.md simply needs to be adjusted to configure this according to JSON API spec (like contentFormat).

We can simply ignore URL_FORMAT_OVERRIDE in QueryParameterValidationFilter if it is not changed from format to something JSON API valid but to be honest this rather sounds like "perfect is the enemy of good"... :smile: if someone wants to break json api spec we cannot stop them, but as long as our recommend configuration doesn't we are safe.

n2ygk:
> @n2ygk I am a bit confused on your comment as you have already upvoted my [comment](https://github.com/django-json-api/django-rest-framework-json-api/pull/535#issuecomment-453937183) and I thought you agreed with it. Browsable API can be used in production and we even document it this way in the README.md so it needs to work.
> FORMAT_SUFFIX_KWARG is only used for custom url patterns and what we are looking for is [URL_FORMAT_OVERRIDE](https://www.django-rest-framework.org/api-guide/settings/#url_format_override) and README.md simply needs to be adjusted to configure this according to JSON API spec (like contentFormat).
> 
> We can simply ignore URL_FORMAT_OVERRIDE in QueryParameterValidationFilter if it is not changed from format to something JSON API valid but to be honest this rather sounds like "perfect is the enemy of good"... 😄 if someone wants to break json api spec we cannot stop them, but as long as our recommend configuration doesn't we are safe.

I upvoted your comment but the PR hardcodes "format" at https://github.com/django-json-api/django-rest-framework-json-api/pull/535/commits/0da5fc22a88f4723b0d7434b432f9537c699355f

I'm happy with them breaking the spec explicitly as they like, but this makes the default be to allow `format` in the Query Parameters. (And yes, I should have commented on the specific code change, sorry.)

n2ygk:
I'm not sure what I'm doing differently, but the browseable API seems to work for me with no special setup. Not sure why that is. Is the `format` QP processed and popped before the `QueryParameterValidationFilter` backend? See my [settings](https://github.com/columbia-it/django-jsonapi-training/blob/master/training/settings.py). Will investigate further and report back.

sliverc:
@n2ygk 
Ahh now I think I understand where the confusion comes from.. :smile: I think you have answered @serenecloud  question before I did but in the GitHub history my comment comes above yours (at least on my view) but the timestamp of your answer is actually before mine...
So I have thought you have answered to my [comment](https://github.com/django-json-api/django-rest-framework-json-api/pull/535#issuecomment-464747820) so I was confused... Hope this makes sense to you.

So to clarify are you ok with this PR being adjusted as I have outlined in [comment](https://github.com/django-json-api/django-rest-framework-json-api/pull/535#issuecomment-464747820)? It is just rephrasing what is basically written in the previous [comment](https://github.com/django-json-api/django-rest-framework-json-api/pull/535#issuecomment-453937183) but includes how to do it.

sliverc:
@n2ygk 
Our answers overlapped again :wink: Sometimes a chat application would be the better way of communication :smile: 

I think you need to configure another renderer like `rest_framework.renders.JSONRenderer` to activate the option as seen in the screenshot of initial PR comment.

n2ygk:
> @n2ygk
> Our answers overlapped again 😉 Sometimes a chat application would be the better way of communication 😄
> 
> I think you need to configure another renderer like `rest_framework.renders.JSONRenderer` to activate the option as seen in the screenshot of initial PR comment.

Let me quote your reply to sync up;-)

I see what happened in my https://github.com/django-json-api/django-rest-framework-json-api/pull/535#issuecomment-464811426. If I just GET /resource in the browseable API, no format QP is sent in yet it seems to work right, I guess because of a default somewhere. If I explicitly add `?format=api` via the GET drop-down in the UI then I can reproduce @serenecloud's original issue.

Per your https://github.com/django-json-api/django-rest-framework-json-api/pull/535#issuecomment-464747820 I still think that "silently" checking whether it's for hardcoded `format` or the `URL_FORMAT_OVERRIDE` value (which defaults to `format`) still means that the default action will be to allow a non-JSONAPI compliant query parameter to pass strict `QueryParameterValidationFilter`. Maybe be more explicit with a DJA setting or `QueryParameterValidatioFilter` attribute to say it's being allowed or vice-versa? Then the documented default can allow it explicitly and when I deploy a backend service that is strictly jsonapi, I can make sure I am not allowing that QP to sneak in.

sliverc:
@n2ygk 
What we could do is to log a warning when [URL_FORMAT_OVERRIDE](https://www.django-rest-framework.org/api-guide/settings/#url_format_override) is set to `format` and a uri contains a `format` query parameter, noting user to configure URL_FORMAT_OVERRIDE to contentFormat. How does this sound?

n2ygk:
The client app won’t see the error. Jsonapi makes the point of being
client-friendly by reporting useful errors. (Something needing improvements
in DJA). Let’s say I have an attribute called format and I meant
‘filter[format]=api’ but instead I gave ‘format=api’....

Better to make this explicitly configured.

sliverc:
@n2ygk 
I see your point but I think we should really avoid adding additional configuration options for what is available in DRF.
Instead let's raise an error after all when [URL_FORMAT_OVERRIDE](https://www.django-rest-framework.org/api-guide/settings/#url_format_override) is `format` and query parameter is `format`.

n2ygk:
> @n2ygk
> I see your point but I think we should really avoid adding additional configuration options for what is available in DRF.
> Instead let's raise an error after all when [URL_FORMAT_OVERRIDE](https://www.django-rest-framework.org/api-guide/settings/#url_format_override) is `format` and query parameter is `format`.

I don't quite understand what you mean here. The default value of URL_FORMAT_OVERRIDE is `format`. The Browseable API requires `format=api`. So how does one explicitly allow the DRF browseable API *or* explicitly enforce only valid jsonapi query parameters?

sliverc:
@n2ygk 
We document to change `URL_FORMAT_OVERRIDE` to `contentFormat` in our README.
In the `QueryParameterValidationFilter` it's checked whether a query parameter is passed on which is equal configured value in `URL_FORMAT_OVERRIDE`. If this is true and the value of `URL_FORMAT_OVERRIDE` is set to something else than `format` then we let the user pass. If `URL_FORMAT_OVERRIDE` is still set to `format` a 400 error is raised (as would be the case today already without this change).

n2ygk:
@sliverc Where'd we end up with this? I've lost track.

sliverc:
@n2ygk This PR should be adjusted as outlined in https://github.com/django-json-api/django-rest-framework-json-api/pull/535#issuecomment-466902615

PR #540, opened 2019-01-01
Remove requested resources from included field

### LINKED ISSUE #524: `included` parameter potentially includes the object itself
Let's say we have a class `Room` and a class `Person` with a foreign key to `Room`. Now I want to create a new person in the room and reload all the people already in the room because adding the person might have changed them for some reason. Therefor I send a `POST` request containing the new person to 
```
persons/?include=room.persons`
```
This lead to the newly created person to be included twice, once in the `data` property of the reply and once in the `included` property. Even though this isn't explicitly forbidden by jsonapi.org, Ember doesn't like this. IMHO an object shouldn't be included twice in a reply - meaning that this should be fixed on the DJA side of things.

puremath:
Fixes #524 

## Description of the Change

The resource(s) in the data field of the response are removed from the included_cache before building the included field for the response.

I added an integration test which checks whether this works on a single comment. If I request a comment and ask for all the comments of its associated blog entry with `?include=entry.comments` all comments except the one which I requested should turn up in the included field.

## Checklist

- [ ] unit-test added (I added a integration test since I couldn't figure out how to unit test this.)
- [ ] documentation updated (couldn't find any relevant information on the documentation.

### DIFF
--- a/rest_framework_json_api/renderers.py
+++ b/rest_framework_json_api/renderers.py
@@ -644,6 +644,21 @@ class JSONRenderer(renderers.JSONRenderer):
             render_data['data'] = json_api_data
 
         if included_cache:
+            if isinstance(json_api_data, list):
+                objects = json_api_data
+            else:
+                objects = [json_api_data]
+
+            for object in objects:
+                obj_type = object.get('type')
+                obj_id = object.get('id')
+                if obj_type in included_cache and \
+                   obj_id in included_cache[obj_type]:
+                    del included_cache[obj_type][obj_id]
+                if not included_cache[obj_type]:
+                    del included_cache[obj_type]
+
+        if included_cache:
             render_data['included'] = list()
             for included_type in sorted(included_cache.keys()):
                 for included_id in sorted(included_cache[included_type].keys()):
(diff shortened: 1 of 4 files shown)


rest_framework_json_api/renderers.py
```diff
@@ -643,6 +643,21 @@ def render(self, data, accepted_media_type=None, renderer_context=None):
...
+        if included_cache:
+            if isinstance(json_api_data, list):
+                objects = json_api_data
+            else:
+                objects = [json_api_data]
+
+            for object in objects:
+                obj_type = object.get('type')
+                obj_id = object.get('id')
+                if obj_type in included_cache.keys() and \
```
sliverc: Could be simplified to:
```suggestion
                if obj_type in included_cache and \
```

puremath:
We could add the actual resource to the included_cache and then remove it again later. However this introduces complexity and reduces performance for every request only to optimize performance for an edge case. After all it will rarely happen that the actual resource also turns up in the included field.

rest_framework_json_api/renderers.py
puremath: Thanks for pointing this out. I some missed this one.

sliverc:
Fair point. Let's leave it this way.

Could you rebase to master and solve linting issue (see https://travis-ci.org/django-json-api/django-rest-framework-json-api/jobs/474761705)

Once this is done it is ready for merging.

sliverc:
Merged. Thanks for your contribution.

puremath:
Thanks for your help to fix this.

PR #559, opened 2019-01-14
Release version 2.7.0

sliverc:
A few fixes have been implemented since the last version so I think it would be good to do another release.


sliverc:
Once this is merged I will create the tag and push it to pypi.

PR #576, opened 2019-02-18
Respect serializer's `id` field.

amw:
## Description of the Change
Until now the renderer ignored `id` field provided by the serializer. This prevented API author from using separate values for internal database key (`pk`) and for user facing resource ID. One reason for using separate values like this is if the pk has to be small for use by foreign keys and external id can be longer, for example a human readable string.

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/example/tests/unit/test_renderer_class_methods.py
+++ b/example/tests/unit/test_renderer_class_methods.py
@@ -41,6 +41,40 @@ def test_build_json_resource_obj():
         serializer.fields, resource, resource_instance, 'user') == output
 
 
+class PrefixedId(serializers.CharField):
+    def get_attribute(self, instance):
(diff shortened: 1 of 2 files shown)


sliverc:
This PR seems to be related to issue #155

For complete support to overwrite primary key of a resource it also needs to be possible to reference such in relationships. How would you go about this with your approach?
Other question is also how the view would need to look like for custom primary key lookup?

Datamance:
What is the recommended workaround for this, for the time being? 

Glad this is being worked on!

sliverc:
Closing this PR as there hasn't been any progress and I don't quite see how this approach will work with relationships.

Please feel free to comment also in #155 for any suggestions how to move forward.

PR #600, opened 2019-03-18
Allow defining of `select_related` per include

### LINKED ISSUE #591: Do `select_related` for ?include
I've found that we have nice `prefetch_for_includes` option on the ModelViewSet. 
I would also like `select_for_includes` mixin that will call add `select_related` and pass there values accordingly to the mapping. 
If you like the idea I'll create PR for that

Anton-Shutik:
Fixes #591

## Description of the Change

This PR adds  `SelectAndPrefetchForIncludesMixin` which handles `include` GET parameter and adds `select_related` and/or `prefetch_related` calls to queryset in order to save some sql queries.

This also deprecates `PrefetchForIncludesHelperMixin`.

`AutoPrefetchMixin` is modified so if developer declared `prefetch_for_includes` that means that it's handled "manually" and no need to auto prefetch related entities. 

Let's say we have a code like below:
```python
class Product(models.Model):
    name = CharField()
    tags = M2MField()
    variant = FKField()


class ProductSerializer():
    tags = ResourceRelatedField()
    

class VariantSerializer():
    product = ResourceRelatedField()


class VariantViewset():
    serializer_class = VariantSerializer
    prefetch_for_include = {
    'product': 'product__tags'
}
    
```
Line `'product': 'product__tags'`   will save one query per included product, because it will prefetch tags.

## Checklist

- [x] unit-test added
- [x] documentation updated

### DIFF
--- a/rest_framework_json_api/views.py
+++ b/rest_framework_json_api/views.py
@@ -119,6 +119,8 @@ class AutoPreloadMixin(object):
             included_model = None
             levels = included.split('.')
             level_model = qs.model
+            # Suppose we can do select_related by default
+            can_select_related = True
             for level in levels:
                 if not hasattr(level_model, level):
                     break
@@ -126,15 +128,19 @@ class AutoPreloadMixin(object):
                 field_class = field.__class__
 
                 is_forward_relation = (
-                    issubclass(field_class, ForwardManyToOneDescriptor) or
-                    issubclass(field_class, ManyToManyDescriptor)
+                    issubclass(field_class, (ForwardManyToOneDescriptor, ManyToManyDescriptor))
                 )
                 is_reverse_relation = (
-                    issubclass(field_class, ReverseManyToOneDescriptor) or
-                    issubclass(field_class, ReverseOneToOneDescriptor)
+                    issubclass(field_class, (ReverseManyToOneDescriptor, ReverseOneToOneDescriptor))
                 )
                 if not (is_forward_relation or is_reverse_relation):
                     break
+
+                # Figuring out if relation should be select related rather than prefetch_related
+                # If at least one relation in the chain is not "selectable" then use "prefetch"
+                can_select_related &= (
+                    issubclass(field_class, (ForwardManyToOneDescriptor, ReverseOneToOneDescriptor))
+                )
 
                 if level == levels[-1]:
                     included_model = field
@@ -151,7 +157,10 @@ class AutoPreloadMixin(object):
                         level_model = model_field.model
 
             if included_model is not None:
-                qs = qs.prefetch_related(included.replace('.', '__'))
+                if can_select_related:
+                    qs = qs.select_related(included.replace('.', '__'))
+                else:
+                    qs = qs.prefetch_related(included.replace('.', '__'))
 
         return qs
 
--- a/rest_framework_json_api/views.py
+++ b/rest_framework_json_api/views.py
@@ -70,7 +70,7 @@ class PrefetchForIncludesHelperMixin(object):
         return qs
 
 
-class SelectAndPrefetchForIncludesMixin(object):
+class PreloadIncludesMixin(object):
     """
     This mixin provides a helper attributes to select or prefetch related models
     based on the include specified in the URL.
@@ -92,7 +92,7 @@ class SelectAndPrefetchForIncludesMixin(object):
         }
     """
     def get_queryset(self):
-        qs = super(SelectAndPrefetchForIncludesMixin, self).get_queryset()
+        qs = super(PreloadIncludesMixin, self).get_queryset()
 
         includes = self.request.GET.get('include', '').split(',') + ['__all__']
 
@@ -237,14 +237,14 @@ class RelatedMixin(object):
 
 
 class ModelViewSet(AutoPrefetchMixin,
-                   SelectAndPrefetchForIncludesMixin,
+                   PreloadIncludesMixin,
                    RelatedMixin,
                    viewsets.ModelViewSet):
     pass
 
 
 class ReadOnlyModelViewSet(AutoPrefetchMixin,
-                           SelectAndPrefetchForIncludesMixin,
+                           PreloadIncludesMixin,
                            RelatedMixin,
                            viewsets.ReadOnlyModelViewSet):
     pass
--- a/rest_framework_json_api/views.py
+++ b/rest_framework_json_api/views.py
@@ -111,11 +111,6 @@ class AutoPrefetchMixin(object):
     def get_queryset(self, *args, **kwargs):
         """ This mixin adds automatic prefetching for OneToOne and ManyToMany fields. """
         qs = super(AutoPrefetchMixin, self).get_queryset(*args, **kwargs)
-
-        # Prefetch includes handled by another mixin, let's do not mix them
-        if hasattr(self, 'prefetch_for_includes'):
-            return qs
-
         included_resources = get_included_resources(self.request)
 
         for included in included_resources:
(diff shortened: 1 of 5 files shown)


rest_framework_json_api/views.py
```diff
@@ -62,10 +70,52 @@ class MyViewSet(viewsets.ModelViewSet):
...
+        return qs
+
+
 class AutoPrefetchMixin(object):
     def get_queryset(self, *args, **kwargs):
         """ This mixin adds automatic prefetching for OneToOne and ManyToMany fields. """
         qs = super(AutoPrefetchMixin, self).get_queryset(*args, **kwargs)
+
+        # Prefetch includes handled by another mixin, let's do not mix them
+        if hasattr(self, 'prefetch_for_includes'):
```
sliverc: This is not optimal as if I simply want to add one prefetch to one include but leave the rest I cannot do it. It is also not backwards compatible.

If I want to I do not need to derive from both mixins so I would simply leave this out.

rest_framework_json_api/views.py
```diff
@@ -62,10 +70,52 @@ class MyViewSet(viewsets.ModelViewSet):
...
+            qs = qs.select_related(*selects)
+
+        if hasattr(self, 'prefetch_for_includes'):
+            prefetches = [self.prefetch_for_includes.get(inc) for inc in includes]
+            qs = qs.prefetch_related(*prefetches)
+
+        return qs
+
+
 class AutoPrefetchMixin(object):
```
sliverc: I would also add a new `AutoPreloadMixin` for naming consistenency. It should also automatically auto load select_related (as outline in #591)

rest_framework_json_api/views.py
```diff
@@ -62,10 +70,52 @@ class MyViewSet(viewsets.ModelViewSet):
         return qs
 
 
+class SelectAndPrefetchForIncludesMixin(object):
```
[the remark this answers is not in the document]
Anton-Shutik: Agree  :) renamed

rest_framework_json_api/views.py
Anton-Shutik: Well, that might cover not all cases. What if I have related model 
```python
class SomeModel():
    product = ForeignKey('')
```

then in serializers.py I have:
```python 
class SomeSerializer():
    product_info = RelatedField(source='product')
```

That Mixin will try to select_related `product_info`  which does not exist, right ? But we can easily handle that with customer dictionary like we do with `prefetch_for_includes`

What do you think ?

rest_framework_json_api/views.py
Anton-Shutik: Agree, removed the change

rest_framework_json_api/views.py
sliverc: `select_for_includes` dictionary certainly makes sense. What I would like to see is for the common cases that auto loading happens e.g. when source of related field references a foreign key or a reverse relation.
If it doesn't (e.g. in case of using SerializerMethodResourceRelated field or similar) user can still use dictionary `select_for_includes`.

This way this PR it is consistent with prefetching where it also auto prefetches but it is also possible to manually prefetch if necessary.

Anton-Shutik: So we just need just a check like below?
```python
class AutoPrefetchMixin(object):
    def get_queryset(self, *args, **kwargs):
        """ This mixin adds automatic prefetching for OneToOne and ManyToMany fields. """
        qs = super(AutoPrefetchMixin, self).get_queryset(*args, **kwargs)

        if hasattr(self, 'select_for_includes'): 
            return qs

        included_resources = get_included_resources(self.request)
        ......
```

sliverc: Not quite what I wanted to say... :wink:

Currently the `AutoPrefetchMixin` only uses prefetch_related [see code](https://github.com/django-json-api/django-rest-framework-json-api/blob/master/rest_framework_json_api/views.py#L107).

As this PR completes manual loading by `select_related` the auto loading should also for completeness and better performance distinguish between `select_related` and `prefetch_related`.

Meaning if it is a to one relation use `select_related` or otherwise `prefetch_related` as today.

Hope this makes sense?

Anton-Shutik: Yes, it does. Sorry for late response.
I'm not sure we can handle it this way.
Example:

```python
class Order():
    customer = ForeignKey()

class Customer():
    preferences = ManyToManyField()


class Preference()
    name = CharField()
```
Let's say we have request `GET api/orders/?include=customer__preferences`, so it should produce queryset like `Order.objects.select_related(customer__preferences)` which will fail, because `preference` should be FK but not m2m relation.
And we can't do `select_related` and pass custom queryset like this:
Order.objects.select_related(Select('customer', queryset=Customer.objects.prefetch_related('preferences')))

So, `prefetch_related` covers _all_ cases. For everything else use `PreloadIncludesMixin`

Let me know if you know any other way we can prefetch and select automatically

sliverc: It needs to be a mix of course what performs better (select related if possible of prefetch_related if M2M)
Anyhow on second thought it is properly better to work on improving prefetch related in another PR at a later point (e.g. also in cases where include is not used but prefetching would be necessary for creating relationships ids)

So let's keep the changes to what there are at the moment. Could you follow up on the other points in the checklist (test, changelog, documenation etc.)? That would be great.

Anton-Shutik: @sliverc I've added a case when we definitely can use `select_related`. Let's say we have a relation chain: `OrderItem -> Order -> Customer`. They're all "[toOne](https://github.com/django-json-api/django-rest-framework-json-api/pull/600/files#diff-0ca988ea2e19e056dcd62de1c07683ddR142)" descriptors, so we can use select_related for them. Otherwise - use prefetch_related.

Also, I've merged these two mixins (AutoPreloadMixin and PrefetchForIncludesHelperMixin) into one, so it preloads automatically, but user/developer might override specific include(s) via `select_for_includes` and/or `prefetch_for_includes`. PrefetchForIncludesHelperMixin is no longer a base class of ModelViewSet, but still in the code as deprecated,  will remove it after this PR will be released.

Test added, docs updated.

Pls take another look.
Thank you

rest_framework_json_api/views.py
```diff
@@ -62,33 +70,78 @@ class MyViewSet(viewsets.ModelViewSet):
...
+        qs = super(AutoPreloadMixin, self).get_queryset(*args, **kwargs)
         included_resources = get_included_resources(self.request)
 
-        for included in included_resources:
+        for included in included_resources + ['__all__']:
+            # Custom defined "select_related" and "prefetch_related" is a priority
+            select_related = self.get_select_related(included)
+            if select_related is not None:
+                qs = qs.select_related(*select_related)
+                continue
```
sliverc: this might be confusing to user when defining a select and prefetch for some reason that only one of both happens.

also same issue as already [discussed](https://github.com/django-json-api/django-rest-framework-json-api/pull/600#discussion_r268538762)

Best to remove both continue statements. As it is now only one mixin it won't be easy to disable auto prefetching though. After refactoring the method length is quite long so a solution might be to split it up into several method or split it into two mixins again. Maybe there are other solutions but I would want to avoid a flag somehow. Open for suggestions.

Anton-Shutik: Spitted into 2 mixins. 
But not actually sure if there are cases when user needs both select_related and prefetch_related and AutoPreload. I guess having AutoPreload by default with ability to override  it (via select or prefetch dict) would work just fine

sliverc: As it is now it is the safest as it behaves as it was before.

Most likely this won't be a problem that auto prefetching and manual preloading is done together as normally manual prefetching is only needed in cases of properties or serializer methods where auto prefetching won't work anyway.

PR #611, opened 2019-04-13
Added django 2.2 LTS support

auvipy:
work on progress

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ matrix:
     - env: TOXENV=py36-df20-django21-drfmaster
     - env: TOXENV=py37-df20-django21-drfmaster
 
+
   include:
     - python: 3.6
       env: TOXENV=flake8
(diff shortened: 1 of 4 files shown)


auvipy:
should we drop python 2.7 support?

auvipy:
@sliverc

sliverc:
We support Python 2.7 as long as Django REST Framework does (of course only for the Django Version which support 2.7 as well).

In terms of this PR please keep it to adding test for Django 2.2 (there are other changes not relevant to Django 2.2 which also make CI fail).

Thanks for working on this.

auvipy:
OK. I will modify this PR. drf is going to drop python 2.7 in 3.10 due to be released in next month. And I am working on the python2.7 dropping works there.

sliverc:
I have worked on this in PR #640 

I guess once DRF 3.10 is released we will need to do another DJA release and then cleanup the testing matrix.

PR #624, opened 2019-05-11
Fix deprecation warnings regarding collections.abc

### LINKED ISSUE #613: Deprecation warning in renderers.py
```
.../python3.7/site-packages/rest_framework_json_api/renderers.py:5: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    from collections import Iterable, OrderedDict, defaultdict
```

Python 3.7 throws a deprecation warning for the collections import in renderers.py

tirkarthi:
Fixes #613 

## Description of the Change

* Fix `DeprecationWarning` over usage of `collections` instead of `collections.abc`

## Checklist

- [ ] unit-test added
- [ ] documentation updated

The PR doesn't seem to have user impacting changes. Let me know if these need to be updated.

Thanks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ any parts of the framework not mentioned in the documentation should generally b
 
 * Avoid exception when trying to include skipped relationship
 * Don't swallow `filter[]` params when there are several
+* Fix DeprecationWarning regarding collections.abc import in Python 3.7
 
 ## [2.7.0] - 2019-01-14
 
--- a/rest_framework_json_api/compat.py
+++ b/rest_framework_json_api/compat.py
@@ -1,0 +1,4 @@
+try:
+    import collections.abc as collections_abc
+except ImportError:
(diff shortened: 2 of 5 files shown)


sliverc:
Thanks for working on this.

I would prefer if we could move those imports into a compat.py as DRF [does](https://github.com/encode/django-rest-framework/blob/master/rest_framework/compat.py) This way we will find the deprecation fixes again in the future in case we can delete it again.

I think a CHANGELOG entry will also be needed below Fixed as a deprecation warning fix is user facing e.g. for users who turn deprecation warnings into errors.

tirkarthi:
Currently there is no `compat.py`. Do you want me to create one and move my changes there?

sliverc:
yes best create one. This way we follow how DRF handles such issues.

tirkarthi:
I have modified my PR to use compat.py and also added a note on Changelog.

PR #632, opened 2019-05-16
Use REST framework serializer functionality to extract includes

Anton-Shutik:
Fixes #

Let's say we have code like below:
```python
class ParentSerializer(HyperlinkedModelSerializer):
    child = json_api_relations.ResourceRelatedField(
        related_link_view_name='parent-child',
        related_link_url_kwarg='parent_pk',
        self_link_view_name='parent-relationships',
        read_only=True,
        source='intermidiate_object.child',
        model=Child
    )
```
Url `GET api/parent/1/` works fine until I want to "include" child in the response (`GET api/parent/1/?include=child`). It does not add the included section, because renderer fails to calculate `child` property in `extract_relation_instance` method. 
Child is not just Foreign key, it is a property of `intermidiate_object` that we have on the parent. I'm pretty sure we have to calculate it exactly same way we do when serializing parent serializer.

## Checklist

- [ ] unit-test added
- [ ] documentation updated


sliverc:
I like where this PR is going. This change might not be 100% backwards compatible especially the serializer_method part. However a change like suggested in #639 could solve this.

sliverc:
:smile: just looked at the `SerializerMethodResourceRelatedField` implementation which actually overwrites `get_attribute`, so your change should work. If you could add a test and changelog that would be great.

Hint: currently failing tests should be fixed with #634

Anton-Shutik:
@sliverc Done

example/tests/unit/test_renderer_class_methods.py
```diff
@@ -148,3 +149,20 @@ def test_extract_root_meta_invalid_meta():
     serializer = InvalidExtractRootMetaResourceSerializer()
     with pytest.raises(AssertionError):
         JSONRenderer.extract_root_meta(serializer, {})
+
+
+def test_extract_relation_instance():
```
sliverc: better to be renamed to `extract_relation_instance_skipped_field`.

example/tests/unit/test_renderer_class_methods.py
```diff
@@ -9,12 +9,13 @@
 
 class ResourceSerializer(serializers.ModelSerializer):
     version = serializers.SerializerMethodField()
+    skipped_field = serializers.ReadOnlyField(source='foobar')
```
sliverc: This test currently only covers an error case I assume you have added to cover `except SkipField` line.

In your PR description you have outlined a case which would only fail if `include` is used and only works if this PR gets merged. Could you add such a test as well?

Anton-Shutik: Added

Anton-Shutik:
@sliverc Are there any more updates required ?

example/tests/test_renderers.py
```diff
@@ -0,0 +1,49 @@
...
+from rest_framework_json_api import serializers
+from rest_framework_json_api.renderers import JSONRenderer
+
+
+class TestJSONRenderer(TestCase):
+
+    def setUp(self):
+        pass
+
+    def test_extract_relation_instance(self):
```
sliverc: move this test to example/tests/unit/test_renderers.py as a pytest

example/tests/test_renderers.py
```diff
@@ -0,0 +1,49 @@
...
+
+        got = JSONRenderer.extract_relation_instance(
+            field=serializer.fields['comments'], resource_instance=entry
+        )
+        assert str(got.query) == str(Comment.objects.filter(entry=entry).query)
+
+        got = JSONRenderer.extract_relation_instance(
+            field=serializer.fields['blog_name'], resource_instance=entry
+        )
+        assert got == blog.name
```
sliverc: this is a bit of an odd test because `extract_relation_instance` is only used for includes and not for simple attributes (like a blog name).
So the test is better to be adapted to return an actual relation which could be returned as an include to be a real test case.

Anton-Shutik: Agree, updated

example/tests/unit/test_renderers.py
```diff
@@ -137,3 +139,14 @@ class EmptyRelationshipViewSet(views.ReadOnlyModelViewSet):
     assert 'relationships' in result['data']
     assert 'bio' in result['data']['relationships']
     assert result['data']['relationships']['bio'] == {'data': None}
+
+
+@pytest.mark.django_db
+def test_extract_relation_instance(comment_factory):
```
sliverc: pytest-factory creates one instance per registered factory. So you can simply use `comment` fixture here and delete the next line.

Anton-Shutik: done

sliverc:
Merged. Thanks for your work.

Anton-Shutik:
Cool, thanks. Any expected release date ?

PR #640, opened 2019-05-24
Add Django 2.2 support

sliverc:
## Checklist

- [x] PR only contains one change (considered splitting up PR)
- [ ] unit-test added
- [x] documentation updated
- [x] `CHANGELOG.md` updated (only for user relevant changes)
- [x] author name in `AUTHORS`

PR #641, opened 2019-05-24
Add missing envlist for Django 2.2 support

sliverc:
Missing change in #640

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -5,6 +5,7 @@ envlist =
     py{34,35,36}-df20-django20-drf{37,38,39,master},
     py37-df20-django20-drf{39,master},
     py{35,36,37}-df20-django21-drf{39,master},
+    py{35,36,37}-df20-django22-drf{39,master},
 
 [testenv]
 deps =


n2ygk:
@sliverc weirdness on the home front again. Travis build went just fine but when I do `tox -e py37-df20-django22-drf39` all the test_filters fail:
```
E                   django.db.utils.IntegrityError: The row in table 'example_entry_authors' with primary key '1' has an invalid foreign key: example_entry_authors.author_id contains a value '1' that does not have a corresponding value in example_author.id.

.tox/py37-df20-django22-drf39/lib/python3.7/site-packages/django/db/backends/sqlite3/base.py:318: IntegrityError
```

I'm running a full `tox` now to see if this happens for other versions. So far on older versions there are no complaints. I wonder what other environment/cached thing is causing this...

sliverc:
The blogentry.json has invalid authors referenced. This shouldn't have worked before either... yeah well PR is coming up.

sliverc:
It is properly a difference in sqlite versions (where SQLite Version on Travis is much older). PR #642 fixing this.

PR #643, opened 2019-05-28
remove disallowed PUT method.

n2ygk:
Fixes #614 

## Description of the Change

This removes the PUT method which is not specified in the JSONAPI specification. Rather, PATCH
is used for updates. PUT and PATCH were both implemented as a [PATCH](https://tools.ietf.org/html/rfc5789) which is incorrect as a [PUT is expected to be a complete replacement](https://tools.ietf.org/html/rfc7231#section-4.3.4) while PATCH is a partial replacement.

**POTENTIAL BREAKING CHANGE**: There were tests and documentation that used `put` that should have used `patch` and it is possible that clients may be using PUT. 

The most visible outcome of removing PUT is that forthcoming DRF 3.10 `generateschema` would incorrectly add the `put` method to the openapi schema doc.

## Checklist

- [x] PR only contains one change (considered splitting up PR)
- [x] unit-test added
- [x] documentation updated
- [x] `CHANGELOG.md` updated (only for user relevant changes)
- [x] author name in `AUTHORS`

PR #651, opened 2019-06-03
Remove not always supported auto `select_related`

sliverc:
Addresses testing issue in https://github.com/django-json-api/django-rest-framework-json-api/pull/643 

## Description of the Change

Polymorphic models do not always support [select_related](https://django-polymorphic.readthedocs.io/en/stable/advanced.html?highlight=select_related#about-queryset-methods) and might actually be [slower](https://django-rest-framework-json-api.readthedocs.io/en/stable/usage.html?highlight=usage#performance-improvements).

Hence reverting change of `AutoPreloadMixin` as it has not been released yet.

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,6 @@ any parts of the framework not mentioned in the documentation should generally b
 ### Changed
 
 * Allow to define `select_related` per include using [select_for_includes](https://django-rest-framework-json-api.readthedocs.io/en/stable/usage.html#performance-improvements)
-* Reduce number of queries to calculate includes by using `select_related` when possible
 * Use REST framework serializer functionality to extract includes. This adds support like using
   dotted notations in source attribute in `ResourceRelatedField`.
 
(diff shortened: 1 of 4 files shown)


Anton-Shutik:
Probably we want Polymorpihc stuff make using custom select/prefetch related config, but not removing `auto_select_related` just because it is not supported by Polymorphic. But backwards compatibility suffers in that case.

sliverc:
First I thought it would be a good idea to automatically choose `select_related` but when thinking about it again in some cases it might only be a very small improvement but in some others it might be a major slow down all depending on the database structure as explained [here](https://django-rest-framework-json-api.readthedocs.io/en/stable/usage.html?highlight=usage#performance-improvements)

So all in all also for backwards compatibility it is better to remove it. And in some case if someone has a data structure which is faster with `select_related` they can now use `select_for_includes`.

PR #655, opened 2019-06-06
Allow partial update with polymorphic serializers

charlieallatson:
## Description of the Change
Passes the partial argument to the subclass of a polymorphic serializer when `to_internal_value` is called. This prevents a validation error being raised if you don't provide all fields when making a PATCH request to the resource.

## Checklist

- [x] unit-test added
- [ ] documentation updated


sliverc:
Thanks for your PR. Could you add a test to ensure that this fix works and a changelog entry? Thanks.

PR #656, opened 2019-06-07
django-filter>=2.1

n2ygk:
Follows #561

## Description of the Change

Updates tox/travis to use `django-filter>=2.1`

## Checklist

- [ ] unit-test added
- [ ] documentation updated


n2ygk:
@sliverc I'm not sure it's strictly necessary to do this but figured perhaps the test requirements should be consistent with what's in requirements-development.txt.

sliverc:
It doesn't really make a difference in terms of what deps are installed but doesn't really hurt either. Hence merging.

PR #681, opened 2019-08-06
Replace relative imports to make isort happy

amw:
README suggest running flake8 before contributing, but current `master` reports issues.
```
$ flake8
./example/tests/test_views.py:14:1: I001 isort found an import in the wrong position
```

This fixes the current issue. Long term you'd probably want to add `flake8` to the test suite and reject PRs that have flake8 problems. If you don't care about `flake8` issues then maybe we should remove it from README?

### DIFF
--- a/example/api/resources/identity.py
+++ b/example/api/resources/identity.py
@@ -6,8 +6,8 @@ from rest_framework.response import Response
 
 from rest_framework_json_api import utils
 
-from ..serializers.identity import IdentitySerializer
-from ..serializers.post import PostSerializer
+from example.api.serializers.identity import IdentitySerializer
+from example.api.serializers.post import PostSerializer
 
 
 class Identity(viewsets.ModelViewSet):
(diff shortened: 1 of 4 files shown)


sliverc:
This change raises an error for me locally. Could it be that you have an outdated isort?

Also see #686

What I find really odd though is why CI is green as flake8 is already [enforced](https://github.com/django-json-api/django-rest-framework-json-api/blob/master/.travis.yml#L16). When I run `tox -e flake8` locally with this change it also fails. :thinking:

amw:
You mean it raises an `isort` error or does the app/pytest fail?

I don't think my `isort` is outdated. It's version is locked by requirements file.
```
$ cat requirements-development.txt | grep isort
flake8-isort==2.7.0
isort==4.3.21
$ pip freeze | grep isort
flake8-isort==2.7.0
isort==4.3.21
$ pytest -q -o console_output_style=classic
.........................................................................................................................................................................................................
201 passed in 7.41 seconds
```

sliverc:
Odd I get the exact same error you get with the same isort version you use but with using your fix.

Most likely an oddness in isort itself. However I don't think using relative and absolute imports mixed is a good idea. So updated this PR not to use relative imports which should also fix this error.

PR #697, opened 2019-08-30
properly anchor ends of url patterns

n2ygk:
## Description of the Change

The example usage of `url()` in `example/urls.py` and `example/urls_test.py` was missing
the `$` anchor at the end of most patterns.  Best practice is to fully specify an anchored URL to avoid inadvertently matching a substring.  

For example: `url(r'foo/bar')` will match `/foo/bar` but also `/foo/barf`.
 
## Checklist

- [x] unit-test added (existing unit tests continue to work)
- [x] documentation updated (all doc examples already had `$` anchors)


n2ygk:
> ...Removed changelog entry though as I think this change is not relevant to current DJA users but only for new users which will start with the example app anyway.

@sliverc Not a big deal, but I expect existing users trying to use the new generateschema functionality will be burned if their code follows the examples and lacks `$` or `/` due to the django simplify_regex bug fix: https://github.com/django/django/pull/11728 which likely won't appear until Django 3.0.

sliverc:
@n2ygk this is properly better explained in the documentation of `generateschema` as someone not having based its urls.py on the example app might have the exact same issue as well.

PR #733, opened 2019-11-12
Make HyperLinkedOnly field includable

Anton-Shutik:
For now it is impossible to skip "data" rendering for any relation and be able to `include` it via GET parameter. This PR fixes that

## Description of the Change

```python
class ProductSerializer():
    variants = json_api_relations.HyperlinkedRelatedField(
        many=True
    )
```
It fails to GET api/product/1/?include=variants

But with this PR it will work

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/rest_framework_json_api/renderers.py
+++ b/rest_framework_json_api/renderers.py
@@ -301,15 +301,20 @@ class JSONRenderer(renderers.JSONRenderer):
         Determines what instance represents given relation and extracts it.
 
         Relation instance is determined exactly same way as it determined
-        in parent serializer
+        in parent serializer, but if it is "skipped" on the serializer
+        it still should be returned via include
         """
-        try:
-            res = field.get_attribute(resource_instance)
-            if isinstance(res, PKOnlyObject):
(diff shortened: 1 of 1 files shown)


Anton-Shutik:
@sliverc would you mind taking a look at this PR ?

sliverc:
Not so sure this is a good idea. A relationship doesn't need to have a relationship pointing back to its parent and therefore we will have an included resource object which is not linked in the compound document.

Also see [specification](https://jsonapi.org/format/#document-compound-documents) in terms of full linkage. What in the specification do you think would allow combining includes and links in a way this PR allows it?

sliverc:
Any feedback to https://github.com/django-json-api/django-rest-framework-json-api/pull/733#issuecomment-553892623 or shall we close this PR?

Anton-Shutik:
@sliverc Yep, the comment makes a sense and looks like we cannot include related objects if there are no links from resource object due to the spec. I think it can be closed

PR #757, opened 2019-12-16
Bundle dependency PRs on a bi-weekly basis

sliverc:
Currently pyup causes a lot of noise so when looking at the closed
PR the real PRs can be easily overlooked.
As we only have test dependencies simply updating on a bi-weekly basis
should be sufficient.

## Checklist

- [ ] unit-test added
- [ ] documentation updated


n2ygk:
@sliverc Sorry, I just noticed these review requests. I've been busy with work/life imbalance. I'll try to get to these in the next week or so.

PR #763, opened 2020-01-16
Enforce conflict response when requested PATCH resource object is unequal to endpoint

sapiosexual:
## Description of the Change

A server MUST return 409 Conflict when processing a PATCH request in which the resource object’s type and id do not match the server’s endpoint.

## Checklist

- [x] unit-test added
- [x] documentation updated


rest_framework_json_api/parsers.py
```diff
@@ -139,6 +139,16 @@ def parse(self, stream, media_type=None, parser_context=None):
         if not data.get('id') and request.method in ('PATCH', 'PUT'):
             raise ParseError("The resource identifier object must contain an 'id' member")
 
+        if request.method in ('PATCH', 'PUT'):
+            kwarg_name = view.lookup_url_kwarg if view.lookup_url_kwarg else view.lookup_field
```
sliverc:
```suggestion
            lookup_url_kwarg = view.lookup_url_kwarg or view.lookup_field
```

I would prefer to use this simpler form as it is also used in [DRF](https://github.com/encode/django-rest-framework/blob/0407a0df8a16fdac94bbd08d49143a74a88001cd/rest_framework/generics.py#L86) and also use same naming of variables (in this case `lookup_url_kwarg`)

example/tests/test_model_viewsets.py
```diff
@@ -185,6 +185,24 @@ def test_patch_requires_id(self):
...
         self.assertEqual(response.status_code, 400)
 
+    def test_patch_requires_correct_id(self):
+        """
+        Verify that 'id' is required to be passed in an update request.
+        """
+        data = {
+            'data': {
+                'type': 'users',
+                'id': 3,
```
sliverc:
```suggestion
                'id': self.miles.pk + 1,
```

it could be that self.miles.pk is actually three... so this to be safe.

example/tests/test_model_viewsets.py
```diff
@@ -185,6 +185,24 @@ def test_patch_requires_id(self):
 
         self.assertEqual(response.status_code, 400)
 
+    def test_patch_requires_correct_id(self):
+        """
+        Verify that 'id' is required to be passed in an update request.
```
sliverc:
```suggestion
        Verify that 'id' is the same then in url
```

rest_framework_json_api/parsers.py
```diff
@@ -139,6 +139,16 @@ def parse(self, stream, media_type=None, parser_context=None):
...
+            if str(data.get('id')) != str(view.kwargs[kwarg_name]):
+                raise exceptions.Conflict(
+                    "The resource object's id ({data_id}) does not match url ({url_id})".format(
```
sliverc: Not sure this is clear maybe better `does not match url's lookup id`?

PR #764, opened 2020-01-22
Pass on instance when using polymorphic serializers

### LINKED ISSUE #759: Pass on instance when using polymorphic serializers
I'm trying to understand the code in `PolymorphicModelSerializer` and I wonder if this is a bug or me just not understanding something, but in `PolymorphicModelSerializer.to_internal_value`, in the end, when in inits `serializer_class` like so: `return serializer_class(data, context=self.context, partial=self.partial).to_internal_value(data)`, why is it passing `data` as the first argument?

For what I see, the init in `BaseSerializer` is expecting the first argument to be `instance` and the second one to be `data`: `def __init__(self, instance=None, data=empty, **kwargs):`. Wouldn't it make more sense for `PolymorphicModelSerializer.to_internal_value` to init `serializer_class` this other way: `return serializer_class(self.instance, data, context=self.context, partial=self.partial).to_internal_value(data)`?

That way the child serializer will keep the instance if there is any and will set the `initial_data` too.

josebama:
Fixes #759

## Description of the Change
As per issue #759 pass `self.instance` as the first parameter when initialising the child serializer from a polymorphic serialiser. This does not affect basic functionality, but makes the child serializer instance consistent with its usage of `instance` and `data`, which can fix potential issues when extending serilaizers that are part of a polymorphic serializer.

## Checklist

- [x] ~~unit-test added~~ -> I couldn't find anywhere relevant to add such a test without making the code less maintainable and harder to refactor in the future
- [ ] ~~documentation updated~~ -> Being such a minimal change that doesn't affect normal usage of the library, I think it would complicate things more than anything else
- [x] ~~`CHANGELOG.md` updated (only for user relevant changes)~~ -> This is not a user relevant change
- [x] author name in `AUTHORS` -> It already was there

### DIFF
--- a/example/tests/test_serializers.py
+++ b/example/tests/test_serializers.py
@@ -232,6 +232,7 @@ class TestPolymorphicModelSerializer(TestCase):
                 instance, data, **kwargs
             )
 
+        original_init = ArtProjectSerializer.__init__
         ArtProjectSerializer.__init__ = init_with_asserts
 
         parent_serializer.is_valid(raise_exception=True)
@@ -240,3 +241,6 @@ class TestPolymorphicModelSerializer(TestCase):
         parent_serializer.save()
 
         # Asserts in the overridden `__init__` method declared above
+
+        # Restore original init to avoid affecting other tests
+        ArtProjectSerializer.__init__ = original_init
--- a/example/tests/test_serializers.py
+++ b/example/tests/test_serializers.py
@@ -7,6 +7,7 @@ from django.utils import timezone
 from rest_framework.request import Request
 from rest_framework.test import APIRequestFactory
 
+from example.factories import ArtProjectFactory
 from rest_framework_json_api.serializers import (
     DateField,
     ModelSerializer,
@@ -15,7 +16,11 @@ from rest_framework_json_api.serializers import (
 from rest_framework_json_api.utils import format_resource_type
 
 from example.models import Author, Blog, Entry
-from example.serializers import BlogSerializer
+from example.serializers import (
+    BlogSerializer,
+    ProjectSerializer,
+    ArtProjectSerializer,
+)
 
 request_factory = APIRequestFactory()
 pytestmark = pytest.mark.django_db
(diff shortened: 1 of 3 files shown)


josebama:
I added a test. It's very hacky, but it's the only way I could think how to do it without making it too tied to the implementation

example/tests/test_serializers.py
```diff
@@ -193,3 +198,40 @@ def test_model_serializer_with_implicit_fields(self, comment, client):
...
+        """
+        # Arrange
+        self.serializer_instance = None
+
+        def save(serializer_self):
+            """
+            Override `ArtProjectSerializer.save` to get the instance serializer,
+            which is later used assertion
+            """
+            self.serializer_instance = serializer_self.instance
```
sliverc: I would add the asserts right here. This way no additional variables are needed. Additionally can you also assert for serializer_self.data and serializer_self.partial as well for completeness.

example/tests/test_serializers.py
```diff
@@ -193,3 +198,40 @@ def test_model_serializer_with_implicit_fields(self, comment, client):
...
+        serializer = ProjectSerializer(
+            instance=self.project, data=data, partial=True
+        )
+        serializer.is_valid(raise_exception=True)
+
+        # Act
+        serializer.save()
+
+        # Assert
+        assert self.serializer_instance is not None
```
sliverc: I would move this into the save method and assert for self.project and not just for None.

josebama: I added the asserts here and used `self.serializer_instance` because I wanted to have all the asserts in the test itself and not in inner methods, which I find easier to read and follow what the test is doing. But I'll move them to the `save` method, it's one of those decisions that are based purely in personal preference.

example/tests/test_serializers.py
```diff
@@ -193,3 +199,48 @@ def test_model_serializer_with_implicit_fields(self, comment, client):
...
+
+        parent_serializer.is_valid(raise_exception=True)
+
+        # Run save to force `ProjectSerializer` to init `ArtProjectSerializer`
+        parent_serializer.save()
+
+        # Asserts in the overridden `__init__` method declared above
+
+        # Restore original init to avoid affecting other tests
+        ArtProjectSerializer.__init__ = original_init
```
sliverc: what when an assert fails? This will never be called and following test might therefore also fail.

Unfortunatly currently we have a mix between unit test and pytest styles in DJA. You could split it up that overwriting of init happens in setup and teardown but not sure this makes it more readable.

In a a case like this one using the newer pytest style would be much easier where you could use [monkeypatch](https://docs.pytest.org/en/latest/monkeypatch.html).

josebama: Yes you are right, I did't think of that. Yes, pytest would be ideal and make a fixture or use monkeypatch, but I think we should keep it consistent with the rest of the file, so I'd go for using the unittest `tearDown`.

If I do the override of the init in the `setUp`, then I think I should move the asserts out of the overridden init and back into the test, using the instance of the test class to pass the args and kwargs from the init to the test.

The other option would be to keep the override in the test, but in the setup getting the reference to the original `ArtProjectSerializer.__init__` and in the tearDown restore it. So it would be something like:

```python
class TestPolymorphicModelSerializer(TestCase):
    def setUp(self):
        self.project = ArtProjectFactory.create()
        self.art_project_serializer_init = ArtProjectSerializer.__init__

    def tearDown(self):
        ArtProjectSerializer.__init__ = self.art_project_serializer_init

    def test_polymorphic_model_serializer_passes_instance_to_child(self):
        pass  # The rest of the stuff that's already there
```

What do you think?

sliverc: It is fine with me if you use unittest and move the asserts into the test method again.

PR #767, opened 2020-02-06
Fix related resource on inherited polymorphic model

### LINKED ISSUE #621: get_related_resource_type failing for polymorphic serializers
We have run into a problem where DRF JSON API resolves the wrong related resource type when trying to include a resource using a polymorphic serializer. Specifically, DRF JSON API will resolve the base model instead of the inherited model.

Below is an abbreviated example for the issue we are running into. In the following example `utils.get_related_resource_type` will resolve the `parent_model` to `models.Product` instead of `models.ArtProduct` (line [here](https://github.com/django-json-api/django-rest-framework-json-api/blob/eb445137ba3731b091c99e269d82adcea36b60e0/rest_framework_json_api/utils.py#L257)). Subsequently, the method will try to access `art_details` on the `parent_model` which will fail since it only exists on `models.ArtProduct` and not `models.Product` (line [here](https://github.com/django-json-api/django-rest-framework-json-api/blob/eb445137ba3731b091c99e269d82adcea36b60e0/rest_framework_json_api/utils.py#L264)). 

A proposed fix would be to resolve the `parent_model` by checking if the `parent_serializer` is a polymorphic serializer and resolving the `parent_model` with something along the lines of: 
`parent_serializer.get_polymorphic_serializer_for_instance(parent_serializer.instance).Meta.model`

If desired, I can provide further clarification and prepare a PR. 

```python3
class ArtProductSerializer(serializers.ModelSerializer):
    price = MoneyField(max_digits=19, decimal_places=4)

    included_serializers = {
        "art_details": ArtDetailSerializer,
    }

    class JSONAPIMeta:
        included_resources = ("art_details",)
        resource_name = "ArtProduct"

    class Meta:
        fields = (
            "id",
            "price",
            "art_details",
        )
        model = models.ArtProduct


class PolymorphicProductSerializer(serializers.PolymorphicModelSerializer):
    polymorphic_serializers = [ArtProductSerializer]

    class Meta:
        model = models.Product


class CartItemSerializer(serializers.ModelSerializer):

    included_serializers = {
        "product": PolymorphicProductSerializer,
    }

    class Meta:
        model = models.CartItem
        fields = ("id", "product")


class CartSerializer(serializers.ModelSerializer):
    included_serializers = {
        "items": CartItemSerializer,
    }

    class Meta:
        model = models.Cart
        fields = ("id", "items")
        read_only_fields = ("id", "items")

    class JSONAPIMeta:
        included_resources = ("items", "items.product")
```

4nickel:
Fixes #621

## Description of the Change

When adding a `ResourceRelatedField` to a serializer that is included in the `polymorphic_serializers`
list of a polymorphic serializer, the `parent_model` is not correctly resolved in `get_related_resource_type`: instead of resolving to the inherited type, the base type is resolved. This can cause an `AttributeError` if the field in question is not present on the base model.

We ran into this, and then found the closed and seemingly abandoned issue referenced above. This PR implements the fix suggested by the original author, who deserves all the credit. The example code has been expanded to trigger the error when running the test-suite.

Thanks in advance for your time and please let us know if you have any suggestions for improvement.

## Checklist

- [x] ~~unit-test added~~ test-suite failure is introduced, but no unit-test added
- [x] ~~documentation updated~~ unnecessary


4nickel:
Thanks for the quick merge, your work is much appreciated.

PR #776, opened 2020-03-25
support nested structures

### LINKED ISSUE #769: DRF J-A renders nested serializers out of spec.
Hi there.
Json-api spec states:

> Attributes may contain any valid JSON value.

>Complex data structures involving JSON objects and arrays are allowed as attribute values. However, any object that constitutes or is contained in an attribute MUST NOT contain a relationships or links member, as those members are reserved by this specification for future use.

Source: https://jsonapi.org/format/#document-resource-object-attributes

However, considering [this](https://github.com/django-json-api/django-rest-framework-json-api/blob/master/rest_framework_json_api/renderers.py#L63) and [this](https://github.com/django-json-api/django-rest-framework-json-api/blob/master/rest_framework_json_api/renderers.py#L246) any nested serializer would be rendered as resource relations even though the nested structure is not a resource.

sapiosexual:
Fixes #769

## Checklist

- [x] unit-test added
- [ ] documentation updated


sapiosexual:
@sliverc Hi. I made PR we discussed here #769
Please link that issue with the pull request.

sapiosexual:
@auvipy yes it is complaint with 1.0 json-api spec.
Please, read #769

sapiosexual:
@auvipy Yeah, I have checked. I really messed up.

Btw, don't merge this pr, I found out that there is an issue with error object generation, which I will resolve on Thursday.

example/tests/test_rendering_strategies.py
```diff
@@ -0,0 +1,209 @@
+from __future__ import absolute_import
```
sliverc, line 1: this test actually tests the renderer and not the serializer itself. Also one strategy is actually deprecated and will be removed in the future so no need to add an additional test module. Best move this tests to `test_renderers` and see how the other tests are written there so there doesn't need to be a full request/response cycle in the test itself.

pytest.ini
```diff
@@ -1,5 +1,5 @@
 [pytest]
 DJANGO_SETTINGS_MODULE=example.settings.test
 filterwarnings =
-    error::DeprecationWarning
-    error::PendingDeprecationWarning
+    ignore::DeprecationWarning
```
sliverc, line 4: those deprecation warning errors are here for a reason so when new updates of dependencies are automatically added they won't be automatically committed if the introduce deprecation warnings.

It is actually possible to ignore test on a test level see https://docs.pytest.org/en/latest/warnings.html and we need to redesign the deprecation warning so it only occurs in cases where a user is affected. See my comments below.

rest_framework_json_api/settings.py
```diff
@@ -6,14 +6,19 @@
...
 
+RELATIONS_RENDERING_STRATEGY = 'RELATIONS'
+ATTRIBUTE_RENDERING_STRATEGY = 'ATTRIBUTE'
+
 DEFAULTS = {
     'FORMAT_FIELD_NAMES': False,
     'FORMAT_TYPES': False,
     'PLURALIZE_TYPES': False,
     'UNIFORM_EXCEPTIONS': False,
+    'NESTED_SERIALIZERS_RENDERING_STRATEGY': RELATIONS_RENDERING_STRATEGY
```
sliverc, line 21: up to this point all DJA settings are booleans. And I don't think we will have a third strategy and this option will be removed in the future anyhow. So I think it is better to have a boolean setting instead of arbitrary values. It will also be easier to configure this way for the user.

Such setting could be called `SERIALIZE_NESTED_SERIALIZERS_AS_JSON` or something.

rest_framework_json_api/settings.py
```diff
@@ -27,6 +32,22 @@ def __init__(self, user_settings=settings, defaults=DEFAULTS):
...
+            JSON_API_SETTINGS_PREFIX + 'NESTED_SERIALIZERS_RENDERING_STRATEGY',
+            self.defaults['NESTED_SERIALIZERS_RENDERING_STRATEGY'])
+
+        if value not in (RELATIONS_RENDERING_STRATEGY, ATTRIBUTE_RENDERING_STRATEGY):
+            raise AttributeError("Invalid value '%s' for JSON API setting "
+                                 "NESTED_SERIALIZERS_RENDERING_STRATEGY" % value)
+        if value == RELATIONS_RENDERING_STRATEGY and \
+                not hasattr(self.user_settings,
+                            JSON_API_SETTINGS_PREFIX + 'NESTED_SERIALIZERS_RENDERING_STRATEGY'):
+            warnings.warn(DeprecationWarning(
```
sliverc, line 46: The nested serializer issue currently doesn't affect many users but with this deprecation warning every user will be notified about a case which might not concern them at all.

I also think a deprecation warning should be more specific which states what serializer and what field is affected. Otherwise it is really for a DJA user hard to figure out what needs to be changed. So I suggest to move this deprecation warning into the renderer.

rest_framework_json_api/renderers.py
```diff
@@ -61,10 +63,16 @@ def extract_attributes(cls, fields, resource):
...
-                    field, (relations.RelatedField, relations.ManyRelatedField, BaseSerializer)
+                    field, (relations.RelatedField, relations.ManyRelatedField)
             ):
                 continue
 
+            if isinstance(field, BaseSerializer):
+                if rendering_strategy == RELATIONS_RENDERING_STRATEGY:
+                    continue
+                elif rendering_strategy == ATTRIBUTE_RENDERING_STRATEGY:
+                    pass
```
sliverc, line 74: I'd prefer we don't have an empty elif clause but comment above why in a certain condition rendering of a field is skipped.

example/tests/test_nested_errors.py
```diff
@@ -0,0 +1,212 @@
...
+                    ]
+                }
+            }
+        }
+
+        self.perform_error_test(data, '/data/attributes/comments/0/attachment/data')
+
+
+@pytest.mark.filterwarning('default::DeprecationWarning:rest_framework_json_api.serializers')
+def test_deprecation_warning(recwarn):
```
sliverc, line 202: in this module also a mix between unit test and pytest... why not just using pytest for the tests above as well?

rest_framework_json_api/serializers.py
```diff
@@ -115,8 +119,41 @@ def validate_path(serializer_class, field_path, path):
...
+                    "{} in serializer {} or set JSON_API_SERIALIZE_NESTED_SERIALIZERS_AS_ATTRIBUTE"
+                    " to False".format(nested_class, clazz)))
+        return fields
+
+
+# If user imports serializer from here we can catch class definition and check
+# nested serializers for depricated use. Probably it is not bad idea to add
+# sparse and included mixins to this definition, in case people want to use
+# DRF-JA without models underlying.
+class Serializer(Serializer, metaclass=SerializerMetaclass):
```
sliverc, line 150: I agree extend this class with those mixin and remove the comment above.

rest_framework_json_api/utils.py
```diff
@@ -312,29 +312,34 @@ def format_drf_errors(response, context, exc):
...
+            errors.extend(format_error_object(message, '/data', response))
     # handle all errors thrown from serializers
     else:
         for field, error in response.data.items():
             field = format_value(field)
             pointer = '/data/attributes/{}'.format(field)
             # see if they passed a dictionary to ValidationError manually
+            # The bit tricky problem is here. It is may be nested drf thing in format
+            # name: error_object, or it may be custom error thrown by user. I guess,
+            # if it is drf error, dict will always have single key
```
sliverc, line 324: The comment is not quite clear what is happening here. Can you elaborate on this issue?

sapiosexual: For the record, all of this needed to put the correct pointer into an error object.

Well, the issue here is that the `error` object may be any valid dict with any string as a key and any object as a value.

Let consider two examples. First one is [validator method in identity serializer](https://github.com/django-json-api/django-rest-framework-json-api/blob/ff0f93ac041d3729e898236917c732bf9eaaeab4/example/api/serializers/identity.py#L16-L25)
As you can see users can specify any dict whatever they want as a detailed message, and old behavior would be "pass it as is, do nothing". By the way, I am not entirely sure that old behavior is correct and we still should pass this object as is. At least we have to append pointer here if it is possible.

The second example is [second level attribute exception](https://github.com/django-json-api/django-rest-framework-json-api/blob/873b48e5d238727559dd9faca68bff6016f5b09c/example/tests/test_errors.py#L121-L134)
In this case, an error object will be 
```
{
    "body": ["This field is required"]
}
```
where the body is a field name. Moreover, if we consider a more deep nested structure, the error object will be more complicated with nested dicts and lists. This is just how DRF validation works and passes errors.
But I am quite sure if we get a single dict DRF error it will be always a dict with a single key consists of field name and either error message, dict or list depending on the serializer field type.

Since we don't know from where the error object come out neither we don't know serializer structure at this point it is a tricky question how we can separate custom error object defined by users from that errors generated by DRF itself. I assumed this heuristic with a length of the dict is the simplest solution, even though it doesn't feel entirely correct.

The other possible solutions are:

1. Overwrite serializer's validation process and extend the validation error class so we can put the correct pointer at the moment we get an error. I assume this is the most correct way to deal with these problems. However, it will take too much effort and I am not able to do that in the nearest future.

2. The spec states:
> An error object MAY have the following members:
> 
> `id`: a unique identifier for this particular occurrence of the problem.
> `links`: a links object containing the following members:
> about: a link that leads to further details about this particular occurrence of the problem.
> status: the HTTP status code applicable to this problem, expressed as a string value.
> code: an application-specific error code, expressed as a string value.
> `title`: a short, human-readable summary of the problem that SHOULD NOT change from occurrence to occurrence of the problem, except for purposes of localization.
> `detail`: a human-readable explanation specific to this occurrence of the problem. Like title, this field’s value can be localized.
> `source`: an object containing references to the source of the error, optionally including any of the following members:
> `pointer`: a JSON Pointer [RFC6901] to the associated entity in the request document [e.g. "/data" for a primary data object, or "/data/attributes/title" for a specific attribute].
> parameter: a string indicating which URI query parameter caused the error.
> `meta`: a meta object containing non-standard meta-information about the error.

We can come up with another heuristic to check that if dict contains any subset of these fields. However it doesn't protect us in two problems: first, these fields may overlap with serializer fields and second, users may pass as a dict key whatever they want which may be neither serializer field or spec stated field.

I am not sure what we should do with this comment, I can expand it with detailed explanation copy-pasted from here with todo inserted or leave it as is with the issue reported.
I would prefer to report this to the doc as a known issue and merge this pr as-is.

sapiosexual: After a night with an idea of overwriting the validation process, I don't think it entirely solves this problem. We would still get all of this problems in case a user uses plain drf serializers

sliverc: From a JSON API specification perspective the nested json is actually just a value and it suggests that an error is reported on an JSON API attribute level.
As we are having some issues concerning this would it not be better that we simply report an error on the attribute level and not on the nested value? I have a feeling this would also follow the spec closer although it doesn't seem to forbid to do more nested errors.

sapiosexual: @sliverc 
I am not sure.
In my opinion, JSON:API spec is quite strict about what exactly pointer attribute must contain
> `pointer`: a JSON Pointer [RFC6901] to the associated entity in the request document [e.g. "/data" for a primary data object, or "/data/attributes/title" for a specific attribute].

So if we have an error somewhere deep in a nested structure, but we report it as it would happen in top-level attribute, it just doesn't feel correct and seems to be out of spec.
Moreover, it is not clear what exactly we have to put into the `detail` field. We aren't seem permitted to put here JSON object, `detail` has to be string, as far as I understand. If we will try to extract this string from the depth of the error object, we would fail into the same problem.

sapiosexual: Also, I have found [this discussion](https://github.com/json-api/json-api/issues/899) that proves my point

sliverc: Thanks for the pointer to the discussion. Actually `detail` doesn't allow dicts anyway so it would not solve any issue.

```
detail: a human-readable explanation specific to this occurrence of the problem.
```

Let's try to think through what the implications are then. Issue is that in [error objects](https://jsonapi.org/format/#error-objects) there is no field required. So what could happen is that someone has a custom error object only with one member which then would be formatted as error. A workaround for such a use case would be simple though for a DJA user - they could simply add a second error object member. What the issue is though that such an error would not be noticed especially critical for existing users. This we should certainly avoid.

What I noticed is that an validation error detail always seems to be a [list](https://github.com/encode/django-rest-framework/blob/master/rest_framework/exceptions.py#L152) in the [error objects](https://jsonapi.org/format/#error-objects) none of the member may be a list though.

I haven't thoroughly checked but a check could be adjusted the following:
* in case an error is a dict, only has one key and the value of this key is a list, we execute format_error_object
* otherwise we simply append it as before

This would be more defensive and not affecting existing users. Not sure whether I have touched all corner cases but let me know what you think.

sliverc: To expand on my suggestion above it still won't be possible to have custom error object within nested serializers but as nested serializers support is a new feature we can document this known issue (maybe also create another issue to work on later) but for existing users they still can have error objects with one key as before without being affected by this change.

As you have dived into this issue a bit more deeply just let me know what you think.

sapiosexual: Hi.
You made a good point that every validation error has to be a list.
I have done some research and I'm quite sure that if every value object in a dictionary is a list it will be a dict with validation errors. In any other case, it is custom error raised by user.
So I decided to replace `if len(error) > 1` with `not all([isinstance(x, list) for x in message.values()])`.
Moreover, I moved this check into the recursive formatter, so we can catch custom made error object at any level of our nested object.

The last question we need to resolve is that should we inject the source field into a custom made error object if it doesn't have one?

example/tests/test_errors.py
```diff
@@ -0,0 +1,218 @@
...
+    serializer_class = EntrySerializer
+    resource_name = 'entries'
+
+    def post(self, request, *args, **kwargs):
+        serializer = self.serializer_class(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+
+urlpatterns = [
+    url(r'^entries-nested/$', DummyTestView.as_view(),
```
auvipy, line 50: url is deprecated in django 3.1, can you use path() instead?

rest_framework_json_api/utils.py
sapiosexual: I would like that it would be our case, but it is not. There are some specific situations when the dict has only one key and value is not a list (actually, another dict), but we still have to perform format_error_object. This situation is covered by the `test_third_level_dict_error`.
Moreover, I found the case when we got many keys in the dict and we have to perform format_error_object for every key.

However, I was wrong that we don't have access to the serializer, which means we can definitely say if the error object is custom or not. 

I will come up with solution when I will get a couple of free hours.

sapiosexual: By the way, I want to know should we inject the `source` object into the custom error object or not? I think It would be nice if we persist one if the error object doesn't have it already.

sliverc: great that we have access to the serializer. In case of injecting of the source object. Yes that would be good, but only if it doesn't already exist. It is kind of blowing up this PR though. Our main issue at this point is error handling and the rest would be mergeable in this PR. So what do you think if we move the part of error handling into a second PR and merge this PR? In the second PR injecting source could also be part.

We won't release before this second PR would have been landed.

auvipy: that sounds reasonable

sapiosexual: I agree with that.

sliverc: @sapiosexual Could you extract the error handling part from this PR into a new then? Thanks.

example/tests/test_errors.py
```diff
@@ -47,8 +47,8 @@ def post(self, request, *args, **kwargs):
 
 
 urlpatterns = [
-    url(r'^entries-nested/$', DummyTestView.as_view(),
-        name='entries-nested-list')
+    path(r'^entries-nested/$', DummyTestView.as_view(),
```
auvipy, line 50: it should be path(''entries-nested/', ..... )

PR #778, opened 2020-04-10
Avoid `AttributeError` for PUT and PATCH methods when using `APIView` 

sapiosexual:
Fixes silenced AttributeError in the case when `lookup_url_kwarg` doesn't exist in a view

## Checklist

- [x] unit-test added
- [x] documentation updated


sliverc:
As `lookup_field` and `lookup_url_kwarg` are set in `GenericAPIView` can you give an example in what case a lookup field is actually not set?

sapiosexual:
@sliverc both of these fields aren't set whenever you use default `APIView`. It is a valid DRF behavior.

PR #782, opened 2020-04-20
Allow repeated filter query parameters

### LINKED ISSUE #718: Filtering array query params
Some clients (e.g. Ember) send requests like `?filter[tags][]=1&filter[tags][]=2` for the membership in a list of values (JSON API).

To process this kind of filter I only changed (https://github.com/django-json-api/django-rest-framework-json-api/blob/master/rest_framework_json_api/django_filters/backends.py#L66) from
```
^filter(?P<ldelim>\[?)(?P<assoc>[\w\.\-]*)(?P<rdelim>\]?$)
```
to
```
^filter(?P<ldelim>\[?)(?P<assoc>[\w\.\-]*)(?P<rdelim>\]?)(\[\])?$
```
Everything seems to work fine. What do you think?

ro70:
Fixes #718

## Description of the Change
Allows repeated filter query parameters like `?filter[authors.id]=1&filter[authors.id]=2`.

## Checklist

- [ ] unit-test added
- [x] documentation updated


rest_framework_json_api/django_filters/backends.py
```diff
@@ -101,19 +101,19 @@ def get_filterset_kwargs(self, request, queryset, view):
...
         data = request.query_params.copy()
-        for qp, val in request.query_params.items():
+        for qp, val in request.query_params.lists():
             m = self.filter_regex.match(qp)
             if m and (not m.groupdict()['assoc'] or
                       m.groupdict()['ldelim'] != '[' or m.groupdict()['rdelim'] != ']'):
                 raise ValidationError("invalid query parameter: {}".format(qp))
             if m and qp != self.search_param:
-                if not val:
+                if not val[0]:
```
sliverc, line 110: The validation error is a bit vague but shouldn't this raise an error when a value of one filter query parameter is empty?
If so then all values of the list should be checked and not just the first one.

@n2ygk can you confirm what the intention of the validation error is? Maybe we need to update the validation error description.

ro70: @sliverc @n2ygk I thought it suffices to have one valid parameter value at least (the first one). But now I see that the code fails if there are non-valid values. So you are right, we have to validate all parameter values.
What about the error description?

n2ygk: @ro70 Sorry that I'm late to this discussion but DJA's *implementation* of filters assumes that they are ANDed together such that `?filter[tags]=1&filter[tags]=2` would never return a result as you are asking for the value of tags to be both 1 and 2 at the same time.  (This even reads as AND since the HTTP separator for query parameters is the `&` symbol.) I believe I said this pretty far back in comments to https://github.com/django-json-api/django-rest-framework-json-api/issues/718#issuecomment-541679826. This PR breaks that.

See the [documentation](https://github.com/django-json-api/django-rest-framework-json-api/blob/master/docs/usage.md#DjangoFilterBackend) of DjangoFilterBackend.

You can implement what you want with `?filter[tags.in]=1,2` by adding this to a view: 
```
class MyViewSet(ModelViewSet):
    filterset_fields =  = {
        'tags': ('in',)
    }
```

If you don't like having to expose the Django [field lookup](https://docs.djangoproject.com/en/3.0/ref/models/querysets/#field-lookups) nomenclature, 
you could probably also define a FilterSet named `tags` that implements `tags__in` something like this:

```
class TagFilterSet(filters.FilterSet):
    """
    Make a filter "alias" that implements `filter[tags]=1,2,3` without having to say `filter[tags.in]=1,2,3`
    """
    tags = filters.CharFilter(field_name="tags", lookup_expr="in")

class MyViewSet(ModelViewSet):
    filterset_class = TagFilterSet
```

So I don't think this PR is warranted at it is intentional not to allow repeated query parameters as they would confuse the client when they make the same error of assuming that repeats of the same query implies an OR rather than an AND. 

There is a mechanism I've outlined above (but not tested:-) that seems to get you the behavior you want.

@sliverc do you agree? We should close this PR and #718.

Perhaps a documentation PR is warranted to add how to configure a `filterset_class` [here](https://github.com/django-json-api/django-rest-framework-json-api/blob/master/docs/usage.md#configuring-filter-backends) -- something I also said in #718.

sliverc: @n2ygk 
I first thought the same but @ro70 outlined in https://github.com/django-json-api/django-rest-framework-json-api/issues/718#issuecomment-612417511 a intersection use case which is not covered by the `in` lookup. And the example he uses with tags very well illustrates this.

Furthermore as I see it the DRF DjangoFilterBackend also allows multiple values of the same query parameter. So if the JSON API spec doesn't specifically disallow it (which I don't think it does) we are actually limiting our users how they can use `django-filter`.

The initial question here was actually why this `ValidationError` is needed as the error message is not quite clear. But if I look at the code it seems that it raises an validation error when the value is empty.
If this was the initial intention I would also prefer this to be removed as again this limits the user how to use `django-filter` (as far as I remember I even had such a use case in the past where an empty value was interpreted by a custom filter class). Unless of course JSON API spec disallows empty query parameter values - which I don't think it does.

Please correct me if one of my above assumption are not valid.

n2ygk: @sliverc Oh, because tags is multi-valued. Sure. Makes sense.

sliverc: @ro70 Let's check for emptiness for all values (this keeps it the same as before).

An error message could be in the line of `missing value for query parameter {}`.

Once this is done and the PR is updated with current master I think it should be ready for merging.

ro70: @sliverc I did your proposed changes and updated my PR.

PR #806, opened 2020-08-11
Attempting to repro this bug

### LINKED ISSUE #805: Support `field_name_mapping` in `RelatedMixin`
I have a case where my client is attempting to access both the `relationships` and `related` views with a different `related_field` name.

I was able to work around this for the relationships view with the `field_name_mapping` but not for my "related" view (based on ModelViewSet).

I poked at things and made a small change to `RelatedMixin` that works around the issue for me.  Would it make sense to submit this or is it intended that the `RelatedMixin` behavior does not support `field_name_mapping`?

steder:
Fixes #805 

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/example/serializers.py
+++ b/example/serializers.py
@@ -19,7 +19,7 @@ from example.models import (
     Project,
     ProjectType,
     ResearchProject,
-    TaggedItem
+    TaggedItem,
 )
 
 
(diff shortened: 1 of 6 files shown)


steder:
It looks like maybe this isn't an issue with the latest code (these changes & tests are just working) so I'm going to revisit what's going on in my codebase.

sliverc:
I assume you have been able to figure out the issue you had, therefore closing. In case this is still an issue feel free to comment.

PR #819, opened 2020-08-26
Adjust batches

sliverc:
# Description of the Change

* Remove gitter chat badge as not being used
* Add PyPI badge

## Checklist

- [ ] unit-test added
- [ ] documentation updated


sliverc:
@n2ygk 
As additional note I am never connected to gitter. I don't know about you but I think we should not promote it when we are actually not connected.

Issue #803, opened 2020-08-07
django-polymorphic now does support Django>3.0


n2ygk:
I don't use polymorphic but noticed that the docs build was failing with this traceback:
```
Traceback (most recent call last):
  File "/Users/alan/src/kezy182/.tox/docs/lib/python3.6/site-packages/sphinx/config.py", line 319, in eval_config_file
    execfile_(filename, namespace)
  File "/Users/alan/src/kezy182/.tox/docs/lib/python3.6/site-packages/sphinx/util/pycompat.py", line 88, in execfile_
    exec(code, _globals)
  File "/Users/alan/src/kezy182/docs/conf.py", line 31, in <module>
    django.setup()
  File "/Users/alan/src/kezy182/.tox/docs/lib/python3.6/site-packages/django/__init__.py", line 24, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/Users/alan/src/kezy182/.tox/docs/lib/python3.6/site-packages/django/apps/registry.py", line 114, in populate
    app_config.import_models()
  File "/Users/alan/src/kezy182/.tox/docs/lib/python3.6/site-packages/django/apps/config.py", line 211, in import_models
    self.models_module = import_module(models_module_name)
  File "/Users/ac45/.pyenv/versions/3.6.10/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/Users/alan/src/kezy182/.tox/docs/lib/python3.6/site-packages/polymorphic/models.py", line 17, in <module>
    from .base import PolymorphicModelBase
  File "/Users/alan/src/kezy182/.tox/docs/lib/python3.6/site-packages/polymorphic/base.py", line 18, in <module>
    from .managers import PolymorphicManager
  File "/Users/alan/src/kezy182/.tox/docs/lib/python3.6/site-packages/polymorphic/managers.py", line 10, in <module>
    from polymorphic.query import PolymorphicQuerySet
[...]

PR #823, opened 2020-09-03
django >=2.2, <=3.1 (drops 1.11)

### LINKED ISSUE #822: add Django 3.1
[Django 3.1](https://docs.djangoproject.com/en/3.1/releases/3.1/) is recently released. Upstream DRF has passing tests for it but has not yet published support for it.

n2ygk:
Fixes #822, #803

## Description of the Change

Remove Django 1.11 from and add Django 3.1 to test matrix, setup.py

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,9 @@ matrix:
     - env: TOXENV=py36-django30-drfmaster
     - env: TOXENV=py37-django30-drfmaster
     - env: TOXENV=py38-django30-drfmaster
+    - env: TOXENV=py36-django31-drfmaster
+    - env: TOXENV=py37-django31-drfmaster
+    - env: TOXENV=py38-django31-drfmaster
 
   include:
     - python: 3.6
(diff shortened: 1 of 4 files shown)


n2ygk:
ugh, forgot that .travis.yml separately lists the environents.

sliverc:
Thanks for looking out for this. I think we should wait with this PR though as DRF 3.11 is not officially supporting Django 3.1. Also locally on my machine the tests are failing with DRF 3.11 and Django 3.1. Reason seems that the `setup.py` is not updated yet in terms of Django range and Travis actually doesn't install Django 3.1 but still 3.0.

See for example [this job](https://travis-ci.org/github/django-json-api/django-rest-framework-json-api/jobs/723871737)

n2ygk:
OK, I've updated setup.py to add py38, and django<3.2. Let's see if that at least fixes the travis job.

n2ygk:
@sliverc FYI - I've opened https://github.com/encode/django-rest-framework/issues/7518 as it does appear that Django 3.1 is supported as of DRF 3.11.1 but the documentation is lagging.

sliverc:
It was quicker to create a new PR #835 instead of updating this. Superseded therefore.

PR #833, opened 2020-09-29
support only drf 3.12

### LINKED ISSUE #832: Add DRF 3.12 support
See https://www.django-rest-framework.org/community/3.12-announcement/

n2ygk:
Fixes #832 

## Checklist

- [ ] unit-test added
- [x] documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,10 +14,15 @@ any parts of the framework not mentioned in the documentation should generally b
 
 * Removed support for Django 1.11.
 * Removed support for Django 2.1.
+* Removed support for Django REST framework 3.10, 3.11
+
+### Added
+* Added support for Django REST framework 3.12
+
 
 ## [3.2.0] - 2020-08-26
 
-This is the last release supporting Django 1.11 and Django 2.1.
+This is the last release supporting Django 1.11, Django 2.1, DRF 3.10 and DRF 3.11.
 
 ### Added
 
--- a/tox.ini
+++ b/tox.ini
@@ -1,16 +1,14 @@
 [tox]
 envlist =
-    py{35,36,37}-django22-drf{310,311,master},
-    py38-django22-drf{311,master},
-    py{36,37,38}-django30-drf{311,master},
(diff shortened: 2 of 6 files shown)


tox.ini
```diff
@@ -1,16 +1,15 @@
 [tox]
 envlist =
     py{35,36,37}-django22-drf{310,311,master},
```
sliverc, line 3: it should be possible to combine this line and the next two into one with something like the following (not tested):

```
py{35,36,37,38}-django{22,30}-drf{312,master}
```

CHANGELOG.md
```diff
@@ -14,6 +14,11 @@ any parts of the framework not mentioned in the documentation should generally b
 
 * Removed support for Django 1.11.
 * Removed support for Django 2.1.
+* Removed support for Django REST framework 3.10, 3.11
+
+### Added
+* Added support for Django REST framework 3.12
```
sliverc, line 20: in release 3.2.0 below please also note, that that will be the last release supporting DRF 3.10 and DRF 3.11

sliverc (changes requested):

tox.ini
auvipy: python 3.5 is EOL

n2ygk: > python 3.5 is EOL

Django 2.2 supports python 3.5.

n2ygk: > it should be possible to combine this line and the next two into one with something like the following (not tested):
> 
> ```
> py{35,36,37,38}-django{22,30}-drf{312,master}
> ```

I believe py35 is not supported by django 3, hence the separation there but I can combine a couple of other lines. Testing now.

sliverc: fair point Django 3 indeed only supports 3.6+

n2ygk:
Wait. this latest push is incorrect. Should only test drf 3.12... Standby, testing again. This is not meant to be merged until 4.0....

sliverc (approved):

sliverc:
We already have backwards incompatible changes in master so next version will be 4.0.0 in any case. So merging this.

Thanks.

PR #842, opened 2020-10-15
Stopped SparseFieldsetsMixin interpreting invalid fields parameter

### LINKED ISSUE #519: SparseFieldsetsMixin catches invalid key from URL
Hi,

Once I want to use sparse fields in URL eg.:

`/api/something/?fields[bar]=somefield`

The SparseFieldsetsMixin is checking for the key as substring:

`key for key in request.query_params if sparse_fieldset_query_param in key`

That causes the situation where not only 'fields[bar]' matches the if case but also prefixes:

`/api/something/?someotherchars_fields[bar]=somefield`

or 

`/api/something/?foooooooofields[bar]=somefield`

Such behavior can cause some unexpected situations once for instance other mixin will use 'fields' key in some way to perform other operations on the same endpoint.

sliverc:
Fixes #519

## Description of the Change

Stopped SparseFieldsetsMixin interpreting invalid fields query parameter

## Checklist

- [x] unit-test added
- [ ] documentation updated


n2ygk (approved):
Woops. Thanks for the fix!

PR #855, opened 2020-11-04
Create tests module

sliverc:
## Description of the Change

This is the basis to start migrating tests from example app.

## Checklist

- [x] PR only contains one change (considered splitting up PR)
- [x] unit-test added
- [ ] documentation updated
- [ ] `CHANGELOG.md` updated (only for user relevant changes)
- [x] author name in `AUTHORS`

PR #856, opened 2020-11-04
Convert utils test to pytest styled tests

sliverc:
# Description of the Change

Added generic models as by example of [DRF](https://github.com/encode/django-rest-framework/blob/master/tests/models.py). Those are not example models as before but generic in representing different constellations possible with Django models. 

## Checklist

- [x] unit-test added
- [ ] documentation updated


sliverc:
Currently marked as draft as #855 needs to be merged first and this rebased.

n2ygk (approved):
Thanks for breaking these tests down into bite-sized PRs.

Issue #857, opened 2020-11-05
Format code with black


sliverc:
Django project [proposed](https://github.com/django/deps/blob/master/accepted/0008-black.rst) to format code with [black](https://github.com/psf/black).

I have been using black in many other projects and had very good experiences with it. As I am currently working on converting tests I very much missed black as a couple of times I had to format code to finally suite flake8 instead of concentrating on coding.

The [proposal](https://github.com/django/deps/blob/master/accepted/0008-black.rst) outlines very well the pros and cons and why Django decided to prose using black.

sliverc:
@n2ygk What do you think about this? I have added black configuration to other projects as well so I could follow up with a PR. As this is larger change though I think it is good to get some consensus before i get started.

n2ygk:
I'm OK with complying with whatever the rules are, but I'm personally cringing at: "88 characters per line and double quotes".... Most projects I work on seem to feel comfortable with line length 120. And single quotes don't require holding down the shift key, at least on a US keyboard.

Also it looks like [PyCharm's built-in formatting does not yet support Black-style](https://youtrack.jetbrains.com/issue/PY-39750). It was planned for the current release (2020.2) but got removed from that release. You can run black on file save hook but that's not the same as having the IDE implement the style as I type...

So I'm not super-excited but will follow your lead.

sliverc:
Thanks for your thoughts.

We should not enforce something which makes working on DJA harder then it should. However before I started to use black I also had some doubts about it and now I miss black any time a project is not using it.

Some thoughts:
* With black I can just format my code as I type without worrying whether I use double quote, or single quotes whether the line length is too long or linting issues. Black will solve this later on for me. So even with black feel free to continue using single quotes when you write new DJA code.
* The same counts with pull request reviews - no more discussions about code formatting because black has made those decisions already
* the line length of 88 is actually not a hard limit very well outlined [here](https://github.com/psf/black/blob/master/docs/the_black_code_style.md#line-length).
* integration into an IDE is nice to have but actually not necessary. What I recommend is we also implement [pre-commit](https://pre-commit.com/) so the code actually gets formatted when user commits it to git. So no worry about how IDE formats the code.

Does this makes sense to you?

I am also open to wait a bit more using black in DJA till `PyCharm` has support for black. Although it doesn't seem like there is any deadline.

n2ygk:
Sure, let's give it a shot and keep the default 88 for now and see how it goes.

[...]

Issue #468, opened 2018-08-29
Convert all tests to pytest


sliverc:
Currently some tests are written in unit test and some in pytest.

As pytest is more flexible all tests should be converted to use pytest to be consistent.

It also needs to be considered whether `pytest_runner` in setup.py is really a good thing or whether just simply running `pytest` would not be more "pytesty".

n2ygk:
Is this consistent with Django's [testing guidelines](https://docs.djangoproject.com/en/2.1/topics/testing/)? It appears to prefer unittest. This is, after all, a Django-centric project.

sliverc:
what we use internally for testing doesn't affect the user so that's up to DJA to decide.

We simply have to make sure that a DJA user can use unittest to write tests. DJA already supports this by allowing [DRF testing configuration](http://www.django-rest-framework.org/api-guide/testing/#configuration).

Why pytest over unittest:
[unittest](https://docs.python.org/3/library/unittest.html) is a Java construct copied into the python world where as pytest has been written with Python in mind and makes writing test so much simpler (especially the fixture part)

Also see [pytest-django](https://pytest-django.readthedocs.io/en/latest/) for more details.

sliverc:
Whenever I review PRs or write PRs myself the most apparent issue of DJA is that the testing code is very messy and confusing for contributors how to write a test and where.

The issues I see at this point:
* mixed styles of tests (Pytest, Unittest)
* mixture between example app and test itself
* unclear naming
* test data shared across several modules (once test data needs to be adjusted totally unrelated tests fail)
* relatively low test coverage

I might have some time at hand to work on this. First I want though that we agree on how to do it before I start.

Following points I see as a way forward:
* create tests module where test will reside (like in DRF)
* use pytest styled tests
* each module defines its own test data
* use snapshot testing where applicable
* strip of the example app of all test related items
* aiming at 100% test coverage or otherwise mark code with no cover if testing is not applicable.

This can be done gradually. So test module by test module can be moved to tests and converted to pytest. Once all is done the example app can be cleaned up

@n2ygk This issue has been open for a while and the main issue here I guess was whether we can agree on converting tests to pytest . My point is that pytest has been around quite a while in the Python community and has been well established. In my experience it makes writing tests much simpler. And if we take the effort to clean up the tests we should also take the opportunity to convert them to pytest style.

What is your take on this? Also comment on what you think is important when cleaning up the tests which I might not have mentioned yet.

[...]

PR #860, opened 2020-11-10
Allow users to overwrite `get_serializer_class` while using related urls

### LINKED ISSUE #859: Allow overwriting of get_serializer_class with related urls
In https://github.com/django-json-api/django-rest-framework-json-api/blob/a67a521327c9ec62a18a875abf13504a72d8eaf0/rest_framework_json_api/views.py#L182
the serializer class defined in the view is used directly. This is a problem if one wants to override `get_serializer_class()`; e.g., to provide a different serializer depending on the view action. Is there a reason to prevent dynamic serializer selection here via `get_serializer_class()`?

uliSchuster:
Fixes #859 

## Description of the Change
Allows users to overwrite the `get_serializer()` and `get_serializer_class()` methods in subclassed views that have related or included serializers.

This change seems to interfere with OpenAPI Schema generation. As it stands, the tests `test_path_with_id_parameter` and `test_path_without_parameters` are failing, because the result no longer matches the expected snapshot. I suspect that the snapshots were incorrect in the first place, but I am not able to tell for sure.

## Checklist

- [o] unit-test added: I did not add a new unit test, but modified the test setup such that the problems became apparent in the existing test cases.
- [-] documentation updated

### DIFF
--- a/example/serializers.py
+++ b/example/serializers.py
@@ -260,6 +260,12 @@ class AuthorSerializer(serializers.ModelSerializer):
     def get_first_entry(self, obj):
         return obj.entries.first()
 
+class AuthorListSerializer(AuthorSerializer):
+    pass
+
+class AuthorDetailSerializer(AuthorSerializer):
+    pass
+
 
 class WriterSerializer(serializers.ModelSerializer):
     included_serializers = {
--- a/example/serializers.py
+++ b/example/serializers.py
@@ -260,8 +260,10 @@ class AuthorSerializer(serializers.ModelSerializer):
     def get_first_entry(self, obj):
         return obj.entries.first()
 
+
 class AuthorListSerializer(AuthorSerializer):
     pass
+
 
 class AuthorDetailSerializer(AuthorSerializer):
     pass
--- a/example/tests/test_views.py
+++ b/example/tests/test_views.py
@@ -367,16 +367,16 @@ class TestRelatedMixin(APITestCase):
         got = view.get_related_instance()
         self.assertEqual(got, self.author.id)
 
-    def test_get_serializer_class(self):
+    def test_get_related_serializer_class(self):
         kwargs = {'pk': self.author.id, 'related_field': 'bio'}
         view = self._get_view(kwargs)
-        got = view.get_serializer_class()
+        got = view.get_related_serializer_class()
         self.assertEqual(got, AuthorBioSerializer)
 
-    def test_get_serializer_class_many(self):
+    def test_get_related_serializer_class_many(self):
         kwargs = {'pk': self.author.id, 'related_field': 'entries'}
         view = self._get_view(kwargs)
-        got = view.get_serializer_class()
+        got = view.get_related_serializer_class()
         self.assertEqual(got, EntrySerializer)
 
     def test_get_serializer_comes_from_included_serializers(self):
@@ -384,15 +384,15 @@ class TestRelatedMixin(APITestCase):
         view = self._get_view(kwargs)
         related_serializers = view.serializer_class.related_serializers
         delattr(view.serializer_class, 'related_serializers')
-        got = view.get_serializer_class()
+        got = view.get_related_serializer_class()
         self.assertEqual(got, AuthorTypeSerializer)
 
         view.serializer_class.related_serializers = related_serializers
 
-    def test_get_serializer_class_raises_error(self):
+    def test_get_related_serializer_class_raises_error(self):
         kwargs = {'pk': self.author.id, 'related_field': 'unknown'}
         view = self._get_view(kwargs)
-        self.assertRaises(NotFound, view.get_serializer_class)
+        self.assertRaises(NotFound, view.get_related_serializer_class)
 
     def test_retrieve_related_single_reverse_lookup(self):
         url = reverse('author-related', kwargs={'pk': self.author.pk, 'related_field': 'bio'})
--- a/example/views.py
+++ b/example/views.py
@@ -15,9 +15,9 @@ from rest_framework_json_api.views import ModelViewSet, RelationshipView
 
 from example.models import Author, Blog, Comment, Company, Entry, Project, ProjectType
 from example.serializers import (
-    AuthorSerializer,
-    AuthorListSerializer,
     AuthorDetailSerializer,
+    AuthorListSerializer,
+    AuthorSerializer,
     BlogDRFSerializer,
     BlogSerializer,
     CommentSerializer,
@@ -190,7 +190,7 @@ class AuthorViewSet(ModelViewSet):
     serializer_classes = {
         "list": AuthorListSerializer,
         "retrieve": AuthorDetailSerializer}
-    serializer_class = AuthorSerializer # fallback
+    serializer_class = AuthorSerializer  # fallback
 
     def get_serializer_class(self):
         try:
--- a/example/views.py
+++ b/example/views.py
@@ -16,6 +16,8 @@ from rest_framework_json_api.views import ModelViewSet, RelationshipView
 from example.models import Author, Blog, Comment, Company, Entry, Project, ProjectType
 from example.serializers import (
     AuthorSerializer,
+    AuthorListSerializer,
+    AuthorDetailSerializer,
     BlogDRFSerializer,
     BlogSerializer,
     CommentSerializer,
@@ -185,7 +187,16 @@ class NoFiltersetEntryViewSet(EntryViewSet):
 
 class AuthorViewSet(ModelViewSet):
     queryset = Author.objects.all()
-    serializer_class = AuthorSerializer
+    serializer_classes = {
+        "list": AuthorListSerializer,
+        "retrieve": AuthorDetailSerializer}
+    serializer_class = AuthorSerializer # fallback
+
+    def get_serializer_class(self):
+        try:
+            return self.serializer_classes.get(self.action, self.serializer_class)
+        except AttributeError:
+            return self.serializer_class
 
 
 class CommentViewSet(ModelViewSet):
--- a/rest_framework_json_api/utils.py
+++ b/rest_framework_json_api/utils.py
@@ -52,7 +52,10 @@ def get_resource_name(context, expand_polymorphic_types=False):
         resource_name = getattr(view, 'resource_name')
     except AttributeError:
         try:
-            serializer = view.get_serializer_class()
+            if 'kwargs' in context and 'related_field' in context['kwargs']:
+                serializer = view.get_related_serializer_class()
+            else:
+                serializer = view.get_serializer_class()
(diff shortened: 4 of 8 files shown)


example/serializers.py
```diff
@@ -260,6 +260,12 @@ class Meta:
     def get_first_entry(self, obj):
         return obj.entries.first()
 
+class AuthorListSerializer(AuthorSerializer):
```
n2ygk, line 263: The reason the openapi components/schemas name changed is because your subclassed view is named `AuthorListSerializer`, and the class it extends is `AuthorSerializer`. DRF generates a default schema name by stripping `Serializer` off the class name.

example/views.py
```diff
@@ -185,7 +187,16 @@ class NoFiltersetEntryViewSet(EntryViewSet):
 
 class AuthorViewSet(ModelViewSet):
     queryset = Author.objects.all()
-    serializer_class = AuthorSerializer
+    serializer_classes = {
```
n2ygk, line 190: Is this new `serializer_classes` functionality something that perhaps belongs in upstream DRF?

n2ygk (changes requested):
Please document how someone might use this new functionality.

example/serializers.py
uliSchuster: Ah, thanks for the explanation. Does that imply I simply should change the snapshots accordingly?

uliSchuster: Fixed...

uliSchuster (commented):
Thanks for pointing out the issue with the OpenAPI Tests. I fixed the relevant snapshots. And I added a more extensive comment on the reason for the changes.

sliverc (commented):
@uliSchuster Thanks for working on this. There are some linting issue which need to be fixed (run flake8 or see CI). I have adjusted the changelog to clarify when this issue is encountered.
I am not too happy about changing an existing test for this but I know tests are currently messy. So I think this is fine for now and can be cleaned up with #468

@n2ygk As you have already started reviewing it I leave you with this to give the final call in terms of approval.

uliSchuster:
I wasn't quite sure which linter you are using and didn't want to break other things (I'm running black and prospector on my project. But I'm happy to fix all remaining linting issues.

About the tests: As I understood from the source, the tests rely on a single example application. The data model, views and serializers of this application effectively are the entities under test. The test cases present already do cover the methods where the issue arises if the entities under test would have included the special case at hand. Writing additional tests alone would not have shown the issue without additional serializers, so I added these serializers.

The alternative would be to add a model with its own views and serializers, and then write tests that use these new classes. I think this would habe been lots of additional code without that much additional benefit.

n2ygk:
@uliSchuster:
> I wasn't quite sure which linter you are using and didn't want to break other things (I'm running black and prospector on my project. But I'm happy to fix all remaining linting issues.

Prior to pushing your commits, you can use [tox -e lint](https://github.com/django-json-api/django-rest-framework-json-api/blob/master/tox.ini#L23-L29) locally to perform a lint test. See also #857 regarding converting the codebase to use black, adding pre-commit hooks, etc. 

Also use `tox -e docs` to locally generate the documentation as html in `docs/_build/html/index.html`.

> About the tests: As I understood from the source, the tests rely on a single example application. The data model, views and serializers of this application effectively are the entities under test. The test cases present already do cover the methods where the issue arises if the entities under test would have included the special case at hand. Writing additional tests alone would not have shown the issue without additional serializers, so I added these serializers.

Per #855 and #856 @sliverc has recently begun converting tests away from using the example app and to consistently use pytest. So, for now, doing them in the example is the right way -- which is why it is so ugly.

`tox` will run all the tests against the matrix of python, Django, DRF and DJA versions.

(Hmm, this should all get added to [CONTRIBUTING.md](https://github.com/django-json-api/django-rest-framework-json-api/blob/master/docs/CONTRIBUTING.md) I suppose: #861)

example/serializers.py
n2ygk: Yes, change the snapshots, or declare your class as AuthorSerializer...

uliSchuster: done

rest_framework_json_api/utils.py
```diff
@@ -52,7 +52,10 @@ def get_resource_name(context, expand_polymorphic_types=False):
         resource_name = getattr(view, 'resource_name')
     except AttributeError:
         try:
-            serializer = view.get_serializer_class()
+            if 'kwargs' in context and 'related_field' in context['kwargs']:
```
n2ygk, line 55: @sliverc please take a look here. I'm not quite sure if there's any potential negative impact of this change.

rest_framework_json_api/views.py
```diff
@@ -144,10 +144,15 @@ def retrieve_related(self, request, *args, **kwargs):
         if isinstance(instance, Iterable):
             serializer_kwargs['many'] = True
 
-        serializer = self.get_serializer(instance, **serializer_kwargs)
+        serializer = self.get_related_serializer(instance, **serializer_kwargs)
         return Response(serializer.data)
 
-    def get_serializer_class(self):
```
n2ygk, line 150: removing this exposes the upstream DRF `get_serializer_class()`.  It seems this needs to remain for compatibility. @sliverc Do you agree?

example/tests/test_views.py
```diff
@@ -367,32 +367,32 @@ def test_get_related_instance_model_field(self):
         got = view.get_related_instance()
         self.assertEqual(got, self.author.id)
 
-    def test_get_serializer_class(self):
```
n2ygk, line 370: See review comment in views.py regarding removing get_serializer_class().

n2ygk (changes requested):
@sliverc please review my review comments as I like the depth of understanding that you have.

rest_framework_json_api/views.py
uliSchuster: My whole point in making the change here is to do exactly this: expose the upstream method, so that it can be overwritten. `get_serializer_class()` is a public method on the base class. As such, it is part of DRF's public API - the [DRF documentation](https://www.django-rest-framework.org/api-guide/generic-views/#attributes) explicitly mentions that the method can be overwritten. In my opinion, an extension library should preserve the upstream API.

The current version of DJA does break the path to the upstream method. There might be cases where DJA users rely on this bug in some way. The change here would break their code. I'm just a first-time user of DJA without any idea about how it is being used in other projects, what the effects might be, and who might be affected.

n2ygk: Forgive me for being dense but how is get_serializer_class() in DJA not able to be overridden by you? You've completely removed the function from DJA; not just added `get_related_serializer()`.

This extension library routinely overrides the upstream DRF because it is extending it.

uliSchuster: I do Full understand that you don't want to run the risk of breaking someone's application. And given that I am fairly new to DJA, I might not fully understand its inner workings, either.

I came across the issue the present PR is trying to solve, when I had to overwrite `get_serializer_class()` on a custom viewset, in exactly the same way I now added as part of the `AuthorViewSet` in the test example. I investigated the issue and discovered that when processing a GET request for a related instance, `RelatedMixin.get_related_instance()` directly accessed the `self.serializer_class` field instead of getting the class via its accessor method. Consequently, the upstream `get_serializer_class()` would not be called, either, which would be my overwritten version. Hence my remark that DJA effectively prevents `get_serializer_class()` to be overwritten on a custom viewset, which I consider to be a bug. 

My first attempt to fix the issue was to have `RelatedMixin.get_related_instance` call the upstream method instead of accessing the field directly. But this breaks things badly. During two exchanges with @sliverc (see the comments on the issue here) showed that the functionality of `RelatedMixin.get_serializer_class` is actually not about getting the serializer class, but getting the class of _related_ serializers. I followed the hint of @sliverc to _rename_ the method from `get_serializer_class()` to `get_related_serializer_class()`, which better captures its functionality. This way, `RelatedMixin.get_related_instance` can call `get_serializer_class()` of the parent class (potentially overwritten), while all functionality that has to do with getting serializers for related views is now handled by aptly named `get_related_*` methods.

To make this work, I need to do some other adjustments along the call chain. In particular, `retrieve_related()`, which is the entry point for a GET request on a related resource, now calls the renamed `get_related_serializer_class()`. One more change, also pointed out by @sliverc, was in `utils.get_resource_name()`: Here, I had to make the distinction if the  serializer of the present viewset is needed, or if the serializer of a related viewset should be returned. The distinction between the two cases depends on the `related_field` kwarg, similar to the condition in `get_serializer_class()`(not `get_related_serializer_class()`.

I hope this helps to provide some understanding of what I am trying to accomplish. I was very much focussed on the functionality of getting related resources; because of my limited understanding of DJA, I am not able to tell if other parts of its functionality are affected by the changes introduced in `RelatedMixin`. Here, I relied on the tests, which are all passing. If there is something else I can do to clear things up or to provide better explanation, please let me know. I am also available for a quick online call, if that can help to better assess the changes I introduced.

sliverc: Let me try to shed some light on this... `RelatedMixin` mistakenly shadowed `get_serializer_class` to get related serializer. Shadowing happened as `RelatedMixin` only works when used with a DRF generic view which provides a [get_serializer_class method](https://www.django-rest-framework.org/api-guide/generic-views/#get_serializer_classself) 
For `RelatedMixin` to be able to get parent serializer it then simply used `self.serializer_class` which is not DRF conform. However this only happened when [related urls](https://django-rest-framework-json-api.readthedocs.io/en/stable/usage.html#related-urls) have been configured as otherwise this [if](https://github.com/django-json-api/django-rest-framework-json-api/blob/master/rest_framework_json_api/views.py#L153) is always false and there is no difference than when the method hasn't been overwritten in `RelatedMixin`.

If someone overwrote `get_serializer_class` in its view without using related urls it would just work as `RelatedMixin` did not do anything anyway. If someone tried to overwrite `get_serializer_class` with related urls this user would run into exactly the same problem as @uliSchuster has.

Therefore I don't see that this change affects existing users.

rest_framework_json_api/utils.py
sliverc: As outlined below I don't see that this has a negative effect as this only takes place when a related url is called.

sliverc (approved):
This looks good now and is in my opinion safe to be merged.

It actually follows our [second goal](https://github.com/django-json-api/django-rest-framework-json-api#goals) to be compatible with DRF as much as possible.

@n2ygk I hope my inline explanations make sense and I suggest we merge this. Otherwise let me know.

n2ygk (approved):

n2ygk:
Thanks for your contribution @uliSchuster and apologies for not understanding it. Thanks as usual @sliverc for explaining.

uliSchuster:
Thanks very much! Are you planning on a release that includes this change anytime soon?

PR #862, opened 2020-11-13
Format code with black

### LINKED ISSUE #857: Format code with black

sliverc:
Fixes #857

## Description of the Change

Format code with black.

There are no other code changes then just formatting of the code. I split up configuration and formatting into two different commits so we can still have a functioning [blame](https://github.com/psf/black#migrating-your-code-style-without-ruining-git-blame).

Therefore **PLEASE** no squash commit on this PR when merging.

Issue which are not directly related to black I will follow up with other PRs:

* Introduce pre-commit and update contributing
* remove ignore of F405 error

## Checklist

- [ ] unit-test added
- [ ] documentation updated


example/migrations/0001_initial.py
```diff
@@ -2,93 +2,154 @@
 # Generated by Django 1.9.5 on 2016-05-02 08:26
```
n2ygk, line 2: As above, exclude migrations.

example/tests/snapshots/snap_test_errors.py
```diff
@@ -4,116 +4,97 @@
```
n2ygk, line 4: Also exclude auto-generated snapshots

n2ygk (changes requested):
Sorry it took me so long to get to this.

If you can make sure auto-generated files are not reformatted, that would be great. Otherwise we'll have a bunch unnecessary diffs.

sliverc:
Thanks for having a look at this.

I deliberately formatted the generated files as well because:
 
* the Django Enhancement Proposal [DEP-8](https://github.com/django/deps/blob/master/accepted/0008-black.rst#specification) specifies it this way

> All code Django generates will also be Black-formatted (startproject, migrations, inspectdb, etc.), at least if the user has Black installed.

* this allows us to enable linting on generated files as most of the generated files actually are edited by humans - this way human error can be minimized. (e.g. migrations can also have user input if a manual migration needs to be written).

Does this makes sense?

n2ygk (approved):

PR #866, opened 2020-11-22
Introduce pre-commit to simplify contributions

### LINKED ISSUE #861: improve CONTRIBUTING
Improve the [CONTRIBUTING.md](https://github.com/django-json-api/django-rest-framework-json-api/blob/master/docs/CONTRIBUTING.md) to document local development and testing steps prior to pushing/submitting a PR.

sliverc:
Fixes #861

## Description of the Change

Add pre-commit for editor independent checking of linting and formatting of code before committing changes to git.

I have added a [pre-commit](https://pre-commit.com/#usage) configuration which uses the locally resp. in the virtual environment installed version of flake8, isort and black. pre-commit also has a way of handling the installation of such tools. However PyUp cannot update those version references and we might end up with different version in pre-commit and requirements.txt which we should avoid.  

The only down side to run tools locally is that a virtual env needs to be activated before committing. This needs to be the case anyway to run tests.

## Checklist

- [ ] unit-test added
- [x] documentation updated

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -1,30 +1,53 @@
-Contributing
-============
+# Contributing
 
-DJA should be easy to contribute to.
+Django REST Framework JSON API (aka DJA)  should be easy to contribute to.
 If anything is unclear about how to contribute,
 please submit an issue on GitHub so that we can fix it!
 
-How
----
+Before writing any code, have a conversation on a GitHub issue to see
+if the proposed change makes sense for the project.
 
-Before writing any code,
-have a conversation on a GitHub issue
-to see if the proposed change makes sense
-for the project.
+## Setup development environment
 
-Fork DJA on [GitHub](https://github.com/django-json-api/django-rest-framework-json-api) and
-[submit a Pull Request](https://help.github.com/articles/creating-a-pull-request/)
-when you're ready.
+### Clone
 
-For maintainers
----------------
+To start developing on Django REST Framework JSON API you need to first clone the repository:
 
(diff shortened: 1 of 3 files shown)


n2ygk (changes requested):
Consider adding something like the following to tox.ini as a mean of making it easy for someone to set up the venv:
```
[testenv:devenv]
envdir = env
basepython = python3.6
usedevelop = True
deps =
    -rrequirements.txt
commands =
    /usr/bin/printf '\n\033[0;31m  dont forget to source env/bin/activate\033[0m\n'
```

sliverc:
@n2ygk Actually to run tox there is no need to create a virtual env as tox manages this for the user. DRF also suggests to install [tox](https://github.com/encode/django-rest-framework/blob/master/CONTRIBUTING.md#running-against-multiple-environments) globally. It is a good reminder though that I missed out on documenting how to use tox. I have integrated it now the same way as DRF documented it.

docs/CONTRIBUTING.md
```diff
@@ -1,30 +1,60 @@
...
-(venv)$ python setup.py sdist bdist_wheel
-(venv)$ twine upload dist/*
-(venv)$ git tag -a v1.2.3 -m 'Release 1.2.3'
-(venv)$ git push --tags
-```
+### Testing
+
+To run tests clone the repository, and then:
+
+     # Setup the virtual environment
```
n2ygk, line 22: @sliverc my point of `tox -e devenv` is you can turn these several steps into one.

n2ygk (approved):
The tox -e devenv stuff can wait for another PR if you see the value of it.

PR #868, opened 2020-12-04
Support many related fields on plain Serializers

### LINKED ISSUE #704: Including lists of resources from non-model annotated objects
I have a `Organization` model with a serializer whose instances get some extra data bound to then when using a specific view. this extra data is two lists of `Product` objects, which is not a django model (just a plain old python object) but its representation should look like a regular relation: the attrs for the organization and then the product lists in `included`. I tried the following:

```python

class OrganizationSerializer(serializers.ModelSerializer):
    included_serializers = {
        "current_cycle": ProductSerializer,
        "next_cycle": ProductSerializer,
    }

    current_cycle = ResourceRelatedField(
        many=True, read_only=True, model=Product
    )
    next_cycle = ResourceRelatedField(
        many=True, read_only=True, model=Product
    )

    class Meta:
        model = Organization
        fields = ("pk", "current_cycle", "next_cycle")

    class JSONAPIMeta:
        included_resources = ("current_cycle", "next_cycle")
```
But this results in `AttributeError: type object 'Organization' has no attribute 'current_cycle'`, which happens in `utils.get_related_resource_type`.

Stepping through the code there, it seems like although it's checking the `model` attribute directly on the relation, it's not reaching into the `child_relation` attr when the `many` is set to True. It does a `child` attribute which in my case wasn't there although `child_relation` is.

Doing a quick and dirty fix by including 
```python
    elif (
        hasattr(relation, "child_relation")
        and hasattr(relation.child_relation, "model"
    )):
        relation_model = relation.child_relation.model
```
in the big if/elif/else in `get_related_resource_type` seems to work. I'm however bery new to this library and maybe I'm just missing something. Otherwise I can make a PR with this fix, of course. Any help would be greatly appreciated :) Thanks!

nattyg93:
Fixes #704 

## Description of the Change

Fix issue preventing resolving the resource type of `ResourceRelatedField(many=True)` fields on plain serializers

## Checklist

- [x] unit-test added
- [n/a] documentation updated

### DIFF
--- a/rest_framework_json_api/utils.py
+++ b/rest_framework_json_api/utils.py
@@ -16,7 +16,6 @@ from django.utils.module_loading import import_string as import_class_from_dotte
 from django.utils.translation import gettext_lazy as _
 from rest_framework import exceptions
 from rest_framework.exceptions import APIException
-from rest_framework.serializers import ManyRelatedField
 
 from .settings import json_api_settings
 
--- a/rest_framework_json_api/utils.py
+++ b/rest_framework_json_api/utils.py
@@ -16,6 +16,7 @@ from django.utils.module_loading import import_string as import_class_from_dotte
 from django.utils.translation import gettext_lazy as _
 from rest_framework import exceptions
 from rest_framework.exceptions import APIException
+from rest_framework.serializers import ManyRelatedField
 
 from .settings import json_api_settings
 
@@ -215,6 +216,11 @@ def get_related_resource_type(relation):
                 return get_related_resource_type(parent_model_relation)
 
     if relation_model is None:
+        # For ManyRelatedFields on plain Serializers the resource_type
+        # cannot be determined from a model, so we must get it from the
+        # child_relation
+        if hasattr(relation, "child_relation"):
+            return get_related_resource_type(relation.child_relation)
         raise APIException(
             _("Could not resolve resource type for relation %s" % relation)
         )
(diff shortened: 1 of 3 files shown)


rest_framework_json_api/utils.py
```diff
@@ -215,6 +215,8 @@ def get_related_resource_type(relation):
                 return get_related_resource_type(parent_model_relation)
 
     if relation_model is None:
+        if hasattr(relation, "child_relation"):
```
sliverc, line 218: not so sure this is a good spot for this fix as it is unexpected and many unnecessary checks have already been done on the relation to end up here.

What about adding a check on line 158 whether relation is instance of `ManyRelatedField` and if yes assign `relation = relation.child_relation`?

sliverc (changes requested):
Thanks for working on this! See my inline comment.

rest_framework_json_api/utils.py
nattyg93: I'm happy to consider other options.

Adding the check to 158 was my original solution but causes issues with resolving `ManyRelatedField`s on Polymorphic serializers. The test `example/tests/integration/test_polymorphism.py::test_polymorphism_on_included_relations` fails due to `relation.parent` resolving to the `ManyRelatedField` rather than the serializer that it is bound to, which results in the wrong `parent_model` being resolved.

sliverc: Hmm... the method `get_related_resource_type` is really messy and I think needs to have a better look how to better accomplish this. But I cannot see another easy way right now. So could you add at least a block comment above the if explaining why this is needed?

This will make it more comprehensible. Thanks.

sliverc (approved):

sliverc:
LGTM. Thanks for your work.

PR #872, opened 2020-12-12
Add warning to docs on permission check with related urls

### LINKED ISSUE #864: Permission checks on related resources not executed
I do have a `ModelViewSet`, say `MyModelViewSet` with custom permissions, and several related resources that are available via hyperlinks. The related resources have permissions that are different from the parent ViewSet. When I access the related resource via the provided hyperlinks, the permissions on the related resource are not enforced.

My hypothesis on what happens is as follows:
As recommended in the documentation, the related resource hyperlinks are made available through the parent view; i.e. through `MyModelViewSet`. The URL configuration is
```python
  path(
        "mymode/<pk>/<related_field>/",
        MyModelViewSet.as_view({"get": "retrieve_related"}),
        name="mymodel-related",
    )
```
When I access the related resource, the method `retrieve_related` on the parent view is executed. Therefore, the permission checks on the parent view are run, but not the permission checks on the related view.

Is this behavior intended? It dod not find it referenced anywhere in the documentation. To me, it came as a surprise and would have opened up serious security problems.

sliverc:
Fixes #864

## Description of the Change

Add notice to make sure users understand permission model when using related urls.

## Checklist

- [ ] unit-test added
- [ ] documentation updated


n2ygk (approved):
Sorry for the delay.

n2ygk:
force merging per #877

PR #876, opened 2020-12-24
Support formatting URL segments via new FORMAT_LINKS setting

### LINKED ISSUE #790: inflection ignored in lookup by RelationshipView
I did a quick search and didn't find a similar issue, but I apologize if I missed one on this topic. And, maybe this is an enhancement request rather than a bug report. It is an inconsistency in type names between the url and rendered document.

For an api view of a model I have eg http://localhost/api/customer-accounts/ where the relationship type (lets just call it relationship-name) is inflected correctly according to the setting JSON_API_FORMAT_TYPES (dasherized, in this case)

```json
{
    "links": {
        "first": "http://localhost/api/customer-accounts/?page%5Bnumber%5D=1",
        "last": "http://localhost/api/customer-accounts/?page%5Bnumber%5D=1",
        "next": null,
        "prev": null
    },
    "data": [
        {
            "type": "customer-accounts",
            "id": "1",
            "attributes": {
                "account-id": 1,
                "account": "1234",
                "insert-ts": "2020-05-20T20:42:52.895506Z",
                "update-ts": "2020-05-20T20:42:52.895643Z"
            },
            "relationships": {
                "relationship-name": {
                    "data": {
                        "type": "relationship-name",
                        "id": "1"
                    }
                }
            }
        }
    ],
    "meta": {
        "pagination": {
            "page": 1,
            "pages": 1,
            "count": 1
        }
    }
}
```

Then I go to eg http://localhost/api/customer-accounts/1/relationships/relationship-name/ and get a 404. Instead, I find the relationship at http://localhost/api/customer-accounts/1/relationships/relationship_name/ which looks like:

```json
{
    "data": {
        "type": "relationship-name",
        "id": "1"
    }
}
```

The rendered document has the correctly inflected type, which doesn't match the url.

I am working around this for now using field name mapping in RelationshipView
```python
field_name_mapping = {
        'relationship-name': 'relationship_name'
    }
```
Could RelationshipView pass relationship-name through the inflector first to get it back to the internal python (underscore) convention?

platinumazure:
Fixes #790

## Description of the Change

Added new setting `JSON_API_FORMAT_LINKS`. This setting is used to inflect serializer properties (i.e., field names) on URLs specifically.

This allows users to have different inflections for properties in JSON responses (using `JSON_API_FORMAT_FIELD_NAMES`, existing behavior) vs properties in relationship and related resource links (using `JSON_API_FORMAT_LINKS`, new behavior, previously no inflection was supported).

I used `JSON_API_FORMAT_LINKS` as the setting name, but I feel it's very important we select a setting name that is as intuitive for users as possible. So, if we think this name could be improved, I'd love to discuss that here and then make a change once we have reached consensus.

This is my first contribution to a Python project, so please rip this apart :smile: Note that I was able to run `black`, `flake8`, and `pytest` individually, but I was not able to get `tox` to work, so I apologize in advance if I've missed something as a result.

## Checklist

- [x] unit-test added
- [x] documentation updated


tests/test_views.py
```diff
@@ -0,0 +1,51 @@
...
+        "dasherize",
+        "camelize",
+        "capitalize",
+        "underscore",
+    ],
+)
+def test_get_related_field_name_handles_formatted_link_segments(format_links):
+    url_segment = format_value(related_model_field_name, format_links)
+
+    request = RequestFactory().get("/basic_models/1/{}".format(url_segment))
```
sliverc, line 27: use [RequestFactory fixture](https://pytest-django.readthedocs.io/en/latest/helpers.html#rf-requestfactory) of pytest-django so it is extendible if we need to in the future. Also a good stop to use [f-strings](https://realpython.com/python-f-strings/#simple-syntax)

tests/test_views.py
```diff
@@ -0,0 +1,51 @@
...
+from rest_framework_json_api import serializers, views
+from rest_framework_json_api.relations import ResourceRelatedField
+from rest_framework_json_api.utils import format_value
+
+from .models import BasicModel
+
+related_model_field_name = "related_field_model"
+
+
+@pytest.mark.urls("tests.urls")
```
sliverc, line 13: I think this line can be removed as you are using a request factory below and are not relying on urls in the test code.

tests/urls.py
```diff
@@ -0,0 +1,22 @@
+from django.conf.urls import re_path
+from rest_framework.routers import SimpleRouter
+
+from .views import BasicModelRelationshipView, BasicModelViewSet
+
+router = SimpleRouter()
+router.register(r"basic_models", BasicModelViewSet, basename="basic-model")
+
+urlpatterns = [
```
sliverc, line 9: I am a bit on the fence with this as we are currently just migrating to the new test module and are not sure how things are turning out. But one issue in the past we had is that we had too much shared test code between different tests making maintanance tricky. As this is currently only needed in test_relations.py I prefer we move it into that module for now. We can always extract it again if need to later on.

tests/views.py
```diff
@@ -0,0 +1,12 @@
+from rest_framework_json_api.views import ModelViewSet, RelationshipView
+
+from .models import BasicModel
+
+
+class BasicModelViewSet(ModelViewSet):
```
sliverc, line 6: Same as above better to be moved to test_relations.py for now.

tests/test_relations.py
```diff
@@ -0,0 +1,46 @@
...
+
+    field = HyperlinkedRelatedField(
+        self_link_view_name="basic-model-relationships",
+        related_link_view_name="basic-model-related",
+        read_only=True,
+    )
+    field.field_name = "relatedField_name"
+
+    expected = {
+        "self": "/basic_models/{}/relationships/{}/".format(
```
sliverc, line 34: using [f-strings](https://realpython.com/python-f-strings/#simple-syntax) I think makes this a bit more readable.

sliverc (changes requested):
Thanks for your PR. Great work! There are some small comments I've added otherwise the PR is shaping up nicely.

Concerning naming of `FORMAT_LINKS`: I see that this could be misleading. But I can only come up with names which are really long and as the documentation states it well how it is used I think it is OK for me to leave it with this name (unless we come up with something better)

tests/urls.py
platinumazure: Your reasoning definitely makes sense. However, I don't know how to override the urlconf machinery besides using the decorator `@pytest.mark.urls()` and pointing to a separate module. Do you know how to tell Django to use a urlconf that is "inline" within the test module? If you can point me in the right direction, I'm happy to make this change.

tests/views.py
platinumazure: This effectively depends on figuring out how to "inline" the urlconf, as I asked about in the other unresolved conversation in this review. Once we can figure out how to move the urlconf inline, this change should be pretty trivial.

platinumazure:
Thanks @sliverc for the review!

I've addressed most of your feedback. The only outstanding item at this point is how to "inline" the urlconf in `tests/urls.py`; the URLs must be set up in order for the RelatedMixin links to reverse-resolve properly. So I need to figure out how to move them but still be able to use `@pytest.mark.urls`, or else I need guidance on mocking the reverse-resolution functionality to avoid depending on a urlconf.

If you have any suggestions off the top of your head, I would be grateful to hear them. Otherwise, I'll try to look into it some more in the next week or so.

tests/urls.py
sliverc: The urlpatterns needs to be moved into `test_relations.py` top level and then you can use the following url marker:

```python
@pytest.mark.urls(__name__)
```

In [test_parsers.py](https://github.com/django-json-api/django-rest-framework-json-api/blob/d8f26fc9f87a8c684a80eafb10fe05524f3b3856/example/tests/test_parsers.py) we already do this but instead of using the urls marker it overwrites [ROOT_URLCONF](https://github.com/django-json-api/django-rest-framework-json-api/blob/d8f26fc9f87a8c684a80eafb10fe05524f3b3856/example/tests/test_parsers.py#L136) setting. I prefer the marker though.

platinumazure: Perfect, I'll give that a shot. Thanks!

platinumazure:
I've made the requested changes, and I believe this is ready for another review. Thanks!

sliverc (approved):
Thanks! LGTM

platinumazure:
Oh darn it, I forgot to change the name to what I had suggested in my other comment.

@sliverc Thoughts on `JSON_API_FORMAT_RELATED_FIELD_LINKS` or `JSON_API_FORMAT_RELATED_LINKS`? Slightly more specific but not too wordy IMO

sliverc:
I think in the future if there are other parts of the urls being automatically determined by DJA the setting should be called `FORMAT_LINKS`. Currently this PR only affects related links so renaming it to `FORMAT_RELATED_LINKS` is a good idea. If in the future if more parts of the various links are generated automatically we can always deprecate `FORMAT_RELATED_LINKS` and move towards using `FORMAT_LINKS`.

It would be great if you could create a PR. Thanks.

platinumazure:
Will do, I'll have another PR up in the next day or two.

PR #883, opened 2021-01-13
Render meta fields of included resource

mast22:
## Description of the Change

When adding serializer as a field the meta fields are not rendered, This change provides the rendering capability, test and some fixes of related tests. 
## Checklist

- [X] unit-test added
- [ ] documentation updated
- [X] author name in `AUTHORS`

### DIFF
--- a/example/models.py
+++ b/example/models.py
@@ -1,6 +1,7 @@
 # -*- encoding: utf-8 -*-
 from __future__ import unicode_literals
 
+from datetime import datetime
 from django.contrib.contenttypes.fields import GenericForeignKey, GenericRelation
 from django.contrib.contenttypes.models import ContentType
 from django.db import models
@@ -129,6 +130,9 @@ class Comment(BaseModel):
 
     class Meta:
         ordering = ("id",)
+
+    def modified_days_ago(self):
+        return (datetime.now() - self.modified_at).days
 
 
 class ProjectType(BaseModel):
--- a/example/models.py
+++ b/example/models.py
@@ -1,7 +1,6 @@
 # -*- encoding: utf-8 -*-
 from __future__ import unicode_literals
 
-from datetime import datetime
 from django.contrib.contenttypes.fields import GenericForeignKey, GenericRelation
 from django.contrib.contenttypes.models import ContentType
 from django.db import models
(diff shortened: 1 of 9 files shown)


rest_framework_json_api/renderers.py
```diff
@@ -402,6 +407,11 @@ def extract_included(
                     included_cache[new_item["type"]][
                         new_item["id"]
                     ] = utils.format_field_names(new_item)
+                    meta = cls.extract_meta(field, serializer_data)
```
sliverc, line 410: Shouldn't this logic be moved into `build_json_resource_obj`? See comment above.

rest_framework_json_api/renderers.py
```diff
@@ -378,6 +378,11 @@ def extract_included(
                         included_cache[new_item["type"]][
                             new_item["id"]
                         ] = utils.format_field_names(new_item)
+                        meta = cls.extract_meta(field, serializer_resource)
```
sliverc, line 381: Shouldn't this logic be moved into build_json_resource_obj? Even the formatting of the field names could be moved there. This is just at a quick glance maybe in practice there are more hurdles.

sliverc (changes requested):
Thanks a lot for this PR. This brings certainly more consistency to DJA.

I have added some inline comments. Please also add a changelog entry. Thanks.

example/models.py
```diff
@@ -1,6 +1,7 @@
 # -*- encoding: utf-8 -*-
 from __future__ import unicode_literals
 
+from datetime import datetime
```
sliverc, line 4: you need to run `isort` to make CI happy.

rest_framework_json_api/renderers.py
sliverc: I found some time to look into this and I was able to move the extract meta logic into the `build_json_resource_obj` and updated your PR.

sliverc (approved):

sliverc:
Thanks for your contributions. As mentioned above I adjusted your PR a bit but all the rest looked good and ready for merging.

mast22:
Thanks, and sorry I could not complete it myself. Glad i could help

PR #893, opened 2021-03-08
Create security policy

sliverc:
## Description of the Change

We did not have any security issues in the past but there might be so I think it is important to have a security policy so users know how to report such with fully disclosing it in a GitHub issue. After all DJA exposes APIs which could be publicly exposed.

I've copied the policy from Django REST Framework and adjusted it. I recommend to read following [guide](https://github.com/google/oss-vulnerability-guide/blob/main/guide.md) which describes how security vulnerabilities are best addressed.

One question remains though is what means do we wanna use to privately communicate with us? Github has [Security Advisories](https://docs.github.com/en/github/managing-security-vulnerabilities/about-github-security-advisories) which I recommend we use. But only a admin can create security advisories. Currently as it seems GitHub does not provide a way for the initial communication.

DRF uses googlegroups for this. Not my favorite but do not see a alternatives. Or are there any other suggestions?

## Checklist

- [ ] unit-test added
- [x] documentation updated

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,9 @@
+# Security Policy
+
+## Reporting a Vulnerability
+
+If you believe you've found something in Django REST Framework JSON API which has security implications, please **do not raise the issue in a public forum**.
+
+Send a description of the issue via email to [whatemailaddressshouldweuse@example.net][security-mail].The project maintainers will then work with you to resolve any issues where required, prior to any public disclosure.
+
+[security-mail]: mailto:rest-framework-security@googlegroups.com
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -4,6 +4,6 @@
 
 If you believe you've found something in Django REST Framework JSON API which has security implications, please **do not raise the issue in a public forum**.
 
-Send a description of the issue via email to [rest-framework-security@googlegroups.com][security-mail].The project maintainers will then work with you to resolve any issues where required, prior to any public disclosure.
+Send a description of the issue via email to [rest-framework-security@googlegroups.com][security-mail]. The project maintainers will then work with you to resolve any issues where required, prior to any public disclosure.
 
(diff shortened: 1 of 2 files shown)


SECURITY.md
```diff
@@ -0,0 +1,9 @@
+# Security Policy
+
+## Reporting a Vulnerability
+
+If you believe you've found something in Django REST Framework JSON API which has security implications, please **do not raise the issue in a public forum**.
+
+Send a description of the issue via email to [whatemailaddressshouldweuse@example.net][security-mail].The project maintainers will then work with you to resolve any issues where required, prior to any public disclosure.
```
n2ygk, line 7: A good question. Maybe `rest-framework-jsonapi-security@googlegroups.com` where the members of that google group are the maintainer or core team? Someone will have to remember to update the group membership whenever maintainers change -- which is not often.

n2ygk (changes requested):

SECURITY.md
sliverc: Created. I will update the PR later on. But could you share the email address which I should add to the security mailing list?

n2ygk: @sliverc See maintainers discussion

sliverc:
@n2ygk I have added you to the security mailing list and also updated the PR with more details. Ready for review again.

sliverc:
@n2ygk Could you have a look again at this PR? Thanks.

SECURITY.md
n2ygk, line 7: @sliverc It still shows `whatemailaddressshouldweuse@example.net` here rather than the proper name.

image

n2ygk (changes requested):

SECURITY.md
sliverc: Oops missed that one. Updated now and also fixed some other typos.

n2ygk (approved):

PR #897, opened 2021-03-20
Properly support formatting of link segments in related urls

platinumazure:
## Description of the Change

Add logic to RelatedMixin (related views) and RelationshipView (relationship views) to make sure related field kwarg is properly converted to underscore format.

I don't have a test yet. Would appreciate some guidance on how to proceed. Basically I need to test the view action and make sure the related serializer is found based on the underscore-converted kwarg.

## Checklist

- [ ] unit-test added
- [x] documentation updated (no need)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,13 @@ any parts of the framework not mentioned in the documentation should generally b
 
 * Allow `get_serializer_class` to be overwritten when using related urls without defining `serializer_class` fallback
 * Preserve field names when no formatting is configured.
+* Properly support `JSON_API_FORMAT_RELATED_LINKS` setting in related urls. In case you want to use `dasherize` for formatting links make sure that your url pattern matches dashes as well like following example:
+  ```
+  url(r'^orders/(?P<pk>[^/.]+)/(?P<related_field>[-\w]+)/$',
+      OrderViewSet.as_view({'get': 'retrieve_related'}),
+      name='order-related'),
+  ```
+
 
 ### Deprecated
 
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -515,6 +515,11 @@ For example, with a serializer property `created_by` and with `'dasherize'` form
 
 The relationship name is formatted by the `JSON_API_FORMAT_FIELD_NAMES` setting, but the URL segments are formatted by the `JSON_API_FORMAT_RELATED_LINKS` setting.
 
+<div class="warning">
+    <strong>Note:</strong>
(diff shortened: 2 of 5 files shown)


sliverc:
I am a bit confused by the title but is this change not just adding support of formatting links to RelationshipView? As `RelatedMixin` is not touched in this PR.

When looking at your changes in this PR it occurred to me that we introduced a backwards incompatible change in #876 which we introduce again in this PR as well.

Imagine someone uses camel cased naming for fields in their serializer. After this change in this PR suddenly it would not find the field anymore as it is looking for a field with underscore naming even though they did not configure `FORMAT_RELATED_LINKS`. A rare case as it is not the python convention but it could certainly happen. In the [parser](https://github.com/django-json-api/django-rest-framework-json-api/blob/master/rest_framework_json_api/parsers.py#L45) we actually addressed this issue that we only underscore field names when the setting `FORMAT_FIELD_NAMES` is actually configured. We should do the same with formatting links.  

In terms of tests a integration test would work best like [one of those](https://github.com/django-json-api/django-rest-framework-json-api/blob/master/example/tests/test_views.py#L84)

platinumazure:
Only converting to underscore when the setting is enabled makes sense. Thanks for the guidance on the test.

As for RelatedMixin vs RelationshipView, probably just a mistype on my part, but I do want to emphasize that both `path/to/resource/relationships/relatedField` and `path/to/resource/relatedField` were both not working correctly, and I'm fixing both in this PR. Only the former uses RelationshipView as far as I understand. But maybe I used the wrong name for the latter URL pattern.

Thanks for the initial review.

sliverc:
In #909 the issue has been addressed that field names are preserved when no formatting has been configured. @platinumazure Do you think you will find time to continue working on this PR?

platinumazure:
@sliverc Hard to say, I've been super busy lately. Sorry.

sliverc:
I had a quick look into this and updated the PR.

To note is that for the formatted url to work dashes also need to be matched by the url pattern like the following:
```
  url(r'^orders/(?P<pk>[^/.]+)/(?P<related_field>[-\w]+)/$',
      OrderViewSet.as_view({'get': 'retrieve_related'}),
      name='order-related'),
  ```

The documentation before only stated `\w` which only matches underscores but not dashes so `dasherize` actually did not work. I've also added a hint in the changelog and the documentation for this. Hope this way it will work now. Happy for feedback.

sliverc (approved):

platinumazure:
Thank you so much @sliverc for finishing this up!

PR #904, opened 2021-04-03
Allow get_serializer_class to be overwritten for related urls without defining serializer_class fallback

SafaAlfulaij:
## Description of the Change
This fixes the fact that getting related view's serializer gets it from the super class, rather than the current class.
Bug can be reproduced by using:
```python
    url(
        r"^entries/(?P<pk>[^/.]+)/(?P<related_field>\w+)/$",
        EntryViewSet.as_view({"get": "retrieve_related"}),
        name="entry-related",
    ),
```
instead of:
```python
    url(
        r"entries/(?P<entry_pk>[^/.]+)/comments$",
        CommentViewSet.as_view({"get": "list"}),
        name="entry-comments",
    ),
```

Navigating to `entries/1/comments/` now will result in an error stating that no `get_serializer_class` has been defined, even though it exists.

## Checklist

- [x] unit-test ~added~ updated
- [ ] documentation updated

### DIFF
--- a/rest_framework_json_api/relations.py
+++ b/rest_framework_json_api/relations.py
@@ -27,7 +27,6 @@ LINKS_PARAMS = [
     "related_link_view_name",
     "related_link_lookup_field",
     "related_link_url_kwarg",
-    "related_link_related_name",
 ]
 
 
@@ -68,8 +67,6 @@ class HyperlinkedMixin(object):
         self.related_link_url_kwarg = kwargs.pop(
             "related_link_url_kwarg", self.related_link_lookup_field
         )
-
-        self.related_link_related_name = kwargs.pop("related_link_related_name", None)
 
         # We include this simply for dependency injection in tests.
         # We can't add it as a class attributes or it would expect an
@@ -121,11 +118,6 @@ class HyperlinkedMixin(object):
         self_kwargs = kwargs.copy()
         self_kwargs.update({"related_field": format_link_segment(field_name)})
         self_link = self.get_url("self", self.self_link_view_name, self_kwargs, request)
-
-        if self.related_link_related_name:
-            self_kwargs.update(
-                {"related_field": format_link_segment(self.related_link_related_name)}
-            )
 
         # Assuming RelatedField will be declared in two ways:
         # 1. url(r'^authors/(?P<pk>[^/.]+)/(?P<related_field>\w+)/$',
--- a/rest_framework_json_api/relations.py
+++ b/rest_framework_json_api/relations.py
@@ -27,6 +27,7 @@ LINKS_PARAMS = [
     "related_link_view_name",
     "related_link_lookup_field",
     "related_link_url_kwarg",
+    "related_link_related_name",
 ]
 
 
@@ -66,6 +67,10 @@ class HyperlinkedMixin(object):
         )
         self.related_link_url_kwarg = kwargs.pop(
             "related_link_url_kwarg", self.related_link_lookup_field
+        )
+
+        self.related_link_related_name = kwargs.pop(
+            "related_link_related_name", None
         )
 
         # We include this simply for dependency injection in tests.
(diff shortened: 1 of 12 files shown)


rest_framework_json_api/relations.py
```diff
@@ -27,6 +27,7 @@
     "related_link_view_name",
     "related_link_lookup_field",
     "related_link_url_kwarg",
+    "related_link_related_name",
```
sliverc, line 30: > This fixes the fact that getting related view's serializer gets it from the super class, rather than the current class.

The PR description states that this PR is addressing issue that related view does not properly get parent serializer class (it seems to be another clean up after #860). I guess [this change](https://github.com/django-json-api/django-rest-framework-json-api/pull/904/files#diff-f6ec81f584f0c3b87cc8ba41a977e3631365ab6242c69f1ec9db51916d65a588) addresses this issue.

So at a quick glance for me it is not quite clear why introducing of a new links parameters is needed to fix this issue. Can you elaborate?

rest_framework_json_api/views.py
```diff
@@ -154,7 +154,7 @@ def get_related_serializer(self, instance, **kwargs):
         return serializer_class(instance, **kwargs)
 
     def get_related_serializer_class(self):
-        parent_serializer_class = super(RelatedMixin, self).get_serializer_class()
```
SafaAlfulaij, line 157: Explanation 1:
This is the actual issue this PR is trying to solve.
Getting the `super()` serializer class will call vanilla DRF, which raises the exception:
> 'EntryViewSet' should either include a `serializer_class` attribute, or override the `get_serializer_class()` method.

rest_framework_json_api/relations.py
SafaAlfulaij: Explanation 2:
We need to add this parameter to specify the the `related` link for hyperlink fields (which can, most of the time, have a different related name).
For example in the example serializers, the `field_name` is `comments_hyperlinked`:
```python
comments_hyperlinked = relations.HyperlinkedRelatedField(
        related_link_view_name="entry-comments",
```
And the related view is:
```python
    url(
        r"entries/(?P<entry_pk>[^/.]+)/comments$",
        CommentViewSet.as_view({"get": "list"}),
        name="entry-comments",
    ),
```
So `comments_hyperlinked` relates to `comments`.
But [here](https://github.com/django-json-api/django-rest-framework-json-api/blob/master/rest_framework_json_api/relations.py#L116), we get the field_names, which will create such a url: `/entries/1/comments_hyperlinked`, which isn't really what we want.

rest_framework_json_api/schemas/openapi.py
```diff
@@ -392,6 +393,9 @@ def _find_related_view(self, view_endpoints, related_serializer, parent_view):
 
     def _field_is_one_or_many(self, field, view):
         serializer = view.get_serializer()
+        if field not in serializer.fields:
```
SafaAlfulaij, line 396: Explanation 3:
This was found out by testing, and I think it's because of dynamic fields.
For example, [here](https://github.com/django-json-api/django-rest-framework-json-api/blob/master/example/serializers.py#L91) we remove the "featured" field when the serializer is initialized, but it [exists](https://github.com/django-json-api/django-rest-framework-json-api/blob/master/rest_framework_json_api/schemas/openapi.py#L349) in the serializer class itself, and therefore being checked.

sliverc:
Thanks for your explanation. When reading through it seems that you are trying to address 3 different issues in one PR (each explanation 1).

1. the actual issue of the super class
2. support of hyperlinked related field
3. another bug

To ensure that each single issue is properly tested and better way of reviewing could you split up this PR? Out of urgency point number 1 would be high priority as it is a regression (I will see whether I will get to that as well at some point later on if you won't get to it before me).

SafaAlfulaij:
I can but not all the tests here will pass.

sliverc:
I added a test (resp. updated as before the test was invalid) and added changelog entry for the parent super class call issue.

Merging this. For the other two issue feel free to either create another PR or issue. Thanks for your testing effort - really appreciated that you found those hidden away bugs in the latest version.

sliverc (approved):

SafaAlfulaij:
I checked the issue with openapi schema generation, and it seems to be with related views as well.
I don't know now how to reproduce it, or if it's even valid.
And as for the hyperlinked related field, I'm not using those a lot, and as they are also related to related views, which I don't prefer, I'm not confident enough to work on it.

I'll open to issues about those problems, and let experts handle them :)

PR #908, opened 2021-04-07
Django 3.2 support

### LINKED ISSUE #907: Django 3.2 compatibility
Django 3.2 is now out, but we are currently blocked on upgrading to it due to djangorestframework-jsonapi specifically disallows that and newer versions as dependencies:
https://github.com/django-json-api/django-rest-framework-json-api/blob/0892e3a8a4dbad9630d70e2b78e18b242a8b057d/setup.py#L101

I couldn't find any other issues related to this so thought I would create one.

Are there any specific concerns with loosening the version constraint so users of this library also can upgrade to Django 3.2?

Tenzer:
Fixes #907

## Description of the Change

Allows the library to be used with Django 3.2.

I had to ignore a warning from Django related to django-debug-toolbar in order for tests to pass. I felt like that probably would be an acceptable way to work around it the deprecation warning without blocking this library to be upgraded to support Django 3.2, especially since the library doesn't depend directly on django-debug-toolbar.

## Checklist

- [ ] unit-test added
- [x] documentation updated

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.6", "3.7", "3.8", "3.9"]
-        django: ["2.2", "3.0", "3.1"]
+        django: ["2.2", "3.0", "3.1", "3.2"]
         django-rest-framework: ["3.12", "master"]
     env:
       PYTHON: ${{ matrix.python-version }}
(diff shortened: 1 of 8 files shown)


auvipy (approved):

sliverc (approved):
Thanks for your work. I have added missing changelog entry.

Concerning debug toolbar: Work is in progress (#468) to separate tests and example app so debug toolbar is not used within tests - so for now I am fine that we ignore this deprecation warning.

Tenzer:
Thank you. I left out the changelog update as I didn't know what format you used for yet-to-be released changes.

simkimsia:
is there a release for this?

Tenzer:
No, not yet. See https://github.com/django-json-api/django-rest-framework-json-api/issues/911 about this as well.

simkimsia:
@Tenzer copy that. I subscribed to #911

PR #909, opened 2021-04-08
Preserve field names when no formatting is configured

### LINKED ISSUE #899: Ensure field names are only reversed when formatting is configured
Through the discussion in #897 it occurred to me that we are not handling reversion of formatting field names very consistently.

In the parser ([here](https://github.com/django-json-api/django-rest-framework-json-api/blob/master/rest_framework_json_api/parsers.py#L45) and [here](https://github.com/django-json-api/django-rest-framework-json-api/blob/master/rest_framework_json_api/parsers.py#L58)) it correctly only reverses formatting of the field name when formatting actually has been configured. If not it just keeps the name as their are.

This should actually be always the case whenever we reverse formatting of field name as otherwise someone cannot name their serializer field names as they want but always need to use underscored naming or the format feature.

After a quick analysis I have found [this](https://github.com/django-json-api/django-rest-framework-json-api/blob/master/rest_framework_json_api/django_filters/backends.py#L123), [this](https://github.com/django-json-api/django-rest-framework-json-api/blob/master/rest_framework_json_api/filters.py#L41), [this](https://github.com/django-json-api/django-rest-framework-json-api/blob/master/rest_framework_json_api/renderers.py#L290), [this](https://github.com/django-json-api/django-rest-framework-json-api/blob/master/rest_framework_json_api/serializers.py#L126) and [this](https://github.com/django-json-api/django-rest-framework-json-api/blob/master/rest_framework_json_api/views.py#L190) code lines which needs to be looked at.

I guess best is to create another utility method which reverses the formatting only if the format setting has been configured.

sliverc:
Fixes #899

## Description of the Change

This way it is possible to have field names which do not follow the python underscore convention. As this introduces new utilities took the opportunity to make format methods more consistent.

## Checklist

- [x] unit-test added
- [ ] documentation updated


sliverc:
@n2ygk Do you think you will find time to have a look at this? Thanks.

n2ygk (approved):
Nice cleanup!

PR #914, opened 2021-04-16
Avoided error when using include query parameter on related urls

sliverc:
## Description of the Change

Using `include` on related urls always through an error even when correctly configured. 
This is a regression of version 4.1.0.

## Checklist

- [x] unit-test added
- [ ] documentation updated


n2ygk (approved):
LGTM

PR #926, opened 2021-04-18
Load included and related serializers in meta class

### LINKED ISSUE #858: included_serializers should be imported in meta class
There is a method `utils.get_included_serializers` which handles importing of included serializers as a serializer can also be defined as string.

This method is called a couple of times during one request (e.g. in the renderer, resource validation etc.). It is much cleaner though to import this in the serializer meta class at import time so it only needs to be done once. In the code someone can then easily access serializer.included_serializers without using a utility.

This should also bring a performance benefit (even though this would be very small I guess).

SafaAlfulaij:
Fixes #858

## Description of the Change
As mentioned in the issue, this caches `included_serializers` and `related_serializers` in the dictionary to ease the process of retrieving in various places.
This approach is also used by DRF for binding fields: https://github.com/encode/django-rest-framework/blob/71e6c30034a1dd35a39ca74f86c371713e762c79/rest_framework/utils/serializer_helpers.py#L137

## Checklist

- [x] unit-test ~added~ updated
- [ ] documentation updated

### DIFF
--- a/rest_framework_json_api/relations.py
+++ b/rest_framework_json_api/relations.py
@@ -16,7 +16,6 @@ from rest_framework_json_api.exceptions import Conflict
 from rest_framework_json_api.utils import (
     Hyperlink,
     format_link_segment,
-    get_included_serializers,
     get_resource_type_from_instance,
     get_resource_type_from_queryset,
     get_resource_type_from_serializer,
@@ -274,7 +273,7 @@ class ResourceRelatedField(HyperlinkedMixin, PrimaryKeyRelatedField):
                 inflection.singularize(field_name),
                 inflection.pluralize(field_name),
             ]
-            includes = get_included_serializers(parent)
+            includes = getattr(parent, "included_serializers", {})
             for field in field_names:
                 if field in includes.keys():
                     return get_resource_type_from_serializer(includes[field])
--- a/rest_framework_json_api/relations.py
+++ b/rest_framework_json_api/relations.py
@@ -273,10 +273,7 @@ class ResourceRelatedField(HyperlinkedMixin, PrimaryKeyRelatedField):
                 inflection.singularize(field_name),
                 inflection.pluralize(field_name),
             ]
-            includes = getattr(parent, "included_serializers", None)
-            if includes is None:
-                return None
-
+            includes = getattr(parent, "included_serializers", dict())
             for field in field_names:
                 if field in includes.keys():
                     return get_resource_type_from_serializer(includes[field])
--- a/rest_framework_json_api/relations.py
+++ b/rest_framework_json_api/relations.py
@@ -273,7 +273,10 @@ class ResourceRelatedField(HyperlinkedMixin, PrimaryKeyRelatedField):
                 inflection.singularize(field_name),
                 inflection.pluralize(field_name),
             ]
-            includes = getattr(parent, "included_serializers", {})
+            includes = getattr(parent, "included_serializers", None)
+            if includes is None:
+                return None
+
             for field in field_names:
                 if field in includes.keys():
                     return get_resource_type_from_serializer(includes[field])
--- a/rest_framework_json_api/serializers.py
+++ b/rest_framework_json_api/serializers.py
@@ -154,15 +154,15 @@ class IncludedResourcesValidationMixin(object):
 
 
 class LazySerializersDict(Mapping):
-    def __init__(self, klass, serializers):
-        self.klass = klass
+    def __init__(self, parent, serializers):
+        self.parent = parent
         self.serializers = serializers
 
     def __getitem__(self, key):
         value = self.serializers[key]
         if not isinstance(value, type):
             if value == "self":
-                value = self.klass
+                value = self.parent
             else:
                 value = import_class_from_dotted_path(value)
             self.serializers[key] = value
@@ -181,23 +181,23 @@ class LazySerializersDict(Mapping):
 
 class SerializerMetaclass(SerializerMetaclass):
     def __new__(cls, name, bases, attrs):
-        klass = super().__new__(cls, name, bases, attrs)
+        serializer = super().__new__(cls, name, bases, attrs)
 
         if attrs.get("included_serializers", None):
             setattr(
-                klass,
+                serializer,
                 "included_serializers",
-                LazySerializersDict(klass, attrs["included_serializers"]),
+                LazySerializersDict(serializer, attrs["included_serializers"]),
             )
 
         if attrs.get("related_serializers", None):
             setattr(
-                klass,
+                serializer,
                 "related_serializers",
-                LazySerializersDict(klass, attrs["related_serializers"]),
+                LazySerializersDict(serializer, attrs["related_serializers"]),
             )
 
-        return klass
+        return serializer
 
 
 # If user imports serializer from here we can catch class definition and check
--- a/rest_framework_json_api/serializers.py
+++ b/rest_framework_json_api/serializers.py
@@ -165,7 +165,6 @@ class LazySerializersDict(MutableMapping):
             if value == 'self':
                 value = self.klass
 
-            print(value)
             value = import_class_from_dotted_path(value)
             self.serializers[key] = value
 
@@ -189,17 +188,15 @@ class LazySerializersDict(MutableMapping):
 
 class SerializerMetaclass(SerializerMetaclass):
     def __new__(cls, name, bases, attrs):
-        #print(name)
-        #print(cls.__module__)
-        #print(attrs)
+        serializer_class_path = attrs['__module__']+'.'+name
 
         included_serializers = attrs.get('included_serializers', None)
         if included_serializers:
-            attrs['included_serializers'] = LazySerializersDict(attrs['__module__']+'.'+name, included_serializers)
+            attrs['included_serializers'] = LazySerializersDict(serializer_class_path, included_serializers)
 
         related_serializers = attrs.get('related_serializers', None)
         if related_serializers:
-            attrs['related_serializers'] = LazySerializersDict(attrs['__module__']+'.'+name, related_serializers)
+            attrs['related_serializers'] = LazySerializersDict(serializer_class_path, related_serializers)
 
         return super().__new__(cls, name, bases, attrs)
 
--- a/rest_framework_json_api/serializers.py
+++ b/rest_framework_json_api/serializers.py
@@ -154,15 +154,15 @@ class IncludedResourcesValidationMixin(object):
 
 
 class LazySerializersDict(MutableMapping):
-    def __init__(self, klass, serializers):
-        self.klass = klass
+    def __init__(self, serializers):  # klass, serializers):
+        # self.klass = klass
         self.serializers = serializers
 
     def __getitem__(self, key):
         value = self.serializers[key]
         if not isinstance(value, type):
-            if value == "self":
-                value = self.klass
+            # if value == "self":
+            # value = self.klass
 
             value = import_class_from_dotted_path(value)
             self.serializers[key] = value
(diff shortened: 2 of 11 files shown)


tests/test_utils.py
```diff
@@ -271,26 +271,28 @@ class PlainRelatedResourceTypeSerializer(serializers.Serializer):
     assert get_related_resource_type(field) == output
 
 
-def test_get_included_serializers():
-    class IncludedSerializersModel(DJAModel):
-        self = models.ForeignKey("self", on_delete=models.CASCADE)
-        target = models.ForeignKey(ManyToManyTarget, on_delete=models.CASCADE)
-        other_target = models.ForeignKey(ManyToManyTarget, on_delete=models.CASCADE)
+class IncludedSerializersModel(DJAModel):
+    self = models.ForeignKey("self", on_delete=models.CASCADE)
```
auvipy, line 275: isn't  self a reserve keyword?

SafaAlfulaij: It is not, self is just a convention.

rest_framework_json_api/serializers.py
```diff
@@ -149,8 +150,55 @@ def validate_path(serializer_class, field_path, path):
...
+        return len(self.serializers)
+
+    def __repr__(self):
+        return dict.__repr__(self.serializers)
+
+
 class SerializerMetaclass(SerializerMetaclass):
-    pass
+    def __new__(cls, name, bases, attrs):
+        serializer_class_path = attrs["__module__"] + "." + name
```
SafaAlfulaij, line 187: I'm not happy about this as it feels really hacky and can go wrong in many ways.
I wrote it just to not make a regression.
Django allows this, but one need proper understanding of Django to understand how this works.

rest_framework_json_api/utils.py
```diff
@@ -286,23 +284,6 @@ def get_default_included_resources_from_serializer(serializer):
     return list(getattr(meta, "included_resources", []))
 
 
-def get_included_serializers(serializer):
```
sliverc, line 289: This is public API. I am in favor of deleting this but this has to go through deprecation first. So best is when this method returns `serializer.included_serializers` and writes a deprecation warning to use `serializer.included_serializers` instead.

rest_framework_json_api/serializers.py
sliverc: What is the reason to use a lazy dictionary? This will only be done once during start up and a fairly simple task. And this would solve this issue that you do not need this workaround right?

tests/test_utils.py
```diff
@@ -271,27 +270,31 @@ class PlainRelatedResourceTypeSerializer(serializers.Serializer):
     assert get_related_resource_type(field) == output
 
 
-def test_get_included_serializers():
-    class IncludedSerializersModel(DJAModel):
-        self = models.ForeignKey("self", on_delete=models.CASCADE)
-        target = models.ForeignKey(ManyToManyTarget, on_delete=models.CASCADE)
-        other_target = models.ForeignKey(ManyToManyTarget, on_delete=models.CASCADE)
+class IncludedSerializersModel(DJAModel):
```
sliverc, line 273: Why are you moving the serializer out of the function. Is the suggested change not working anymore with inline serializers?

tests/test_utils.py
```diff
@@ -271,27 +270,31 @@ class PlainRelatedResourceTypeSerializer(serializers.Serializer):
...
+        "other_target": "tests.serializers.ManyToManyTargetSerializer",
+    }
+
+    class Meta:
+        model = IncludedSerializersModel
+        fields = ("self", "other_target", "target")
 
-    included_serializers = get_included_serializers(IncludedSerializersSerializer)
+
+def test_get_included_serializers():
```
sliverc, line 294: as this is not a utils anymore this test should actually be added to `test_serializers`. Unfortunantly those tests have not be converted yet but feel free to add it to `example/tests/test_serializers.py`

tests/test_utils.py
```diff
@@ -271,27 +270,31 @@ class PlainRelatedResourceTypeSerializer(serializers.Serializer):
     assert get_related_resource_type(field) == output
 
 
-def test_get_included_serializers():
```
sliverc, line 274: as `get_included_serializers` should be deprecated instead of being removed this test should remain, check that a deprecation warning has been raised and test as before. This way we can be sure that the code is backwards compatible.

tests/test_utils.py
```diff
@@ -271,27 +270,31 @@ class PlainRelatedResourceTypeSerializer(serializers.Serializer):
...
+    }
+
+    class Meta:
+        model = IncludedSerializersModel
+        fields = ("self", "other_target", "target")
 
-    included_serializers = get_included_serializers(IncludedSerializersSerializer)
+
+def test_get_included_serializers():
+    included_serializers = getattr(
```
sliverc, line 295: From an API perspective this is a bit ugly that `getattr` always needs to be called instead of simply `serializer.included_serializers`. I would prefer that `included_serializers` is set to `None` when not configured so it easier to read `included_serializers` for a DJA user.

sliverc (changes requested):
Thanks for working on this. See me inline comments and do not forget to write a changelog entry.

tests/test_utils.py
SafaAlfulaij: It works, but not self-referencing inline serializers.
The issue was that `self` refers to the inline serializer, which needs to be looked up by it's path in the module, and it can't be found since it's in the function scope...
So we do have a regression here:
- self-referencing inline serializers won't work.

rest_framework_json_api/serializers.py
SafaAlfulaij: This is only for `"self"` in included_serializers, as the serializer is yet to be initialized, and we have to somehow have a reference to it, so I use this hacky way to get the module path of the serializer, to be called later lazily...

SafaAlfulaij:
Changes done.
But I'm not sure what to do with "self" and inline self-related serializers.

rest_framework_json_api/serializers.py
sliverc: I just think this lazy loading is way to overkill for a simple task like this. So am I understanding you correct you only do this lazy loading because of `self` reference?

I have a feeling there needs to be a way not to do this lazily.

SafaAlfulaij: As you see, we are currently in the meta class of the class, meaning the class itself doesn't yet exist.
I don't see any other option but to lazy load it later somehow.
It seems that Django saves `to="self"` till it actually adds the field to the model.

sliverc: Sure here it would not work, but I was rather thinking whether there is a better spot for this code snippet (e.g. in the view). It seems code in another spot might get even more complex though.

Anyway this lazy dict in my opinion does not need to be a `MutableMapping` though, a `Mapping` is enough which makes the code less verbose as well (a few methods can be deleted).

To solve the `self` reference issue another idea I had is that super needs to be called first and the klass can be passed on to `LazySerializersDict`.

```python
    def __new__(cls, name, bases, attrs):
        klass = super().__new__(cls, name, bases, attrs)
        # assign lazy dict here where klass can be passed on for self
        return klass
``` 

Just on the top of my head not sure I have overlooked anything but I guess worth a try.

SafaAlfulaij:
Ping?

tests/test_utils.py
sliverc: This comment has not been addressed yet. Easiest I guess is to set `included_serializers` in the serializer base class to `None` so most of those `getattr` can be removed and simply property opened.

tests/test_utils.py
```diff
@@ -346,28 +346,32 @@ class PlainRelatedResourceTypeSerializer(serializers.Serializer):
 
 
 def test_get_included_serializers():
-    class IncludedSerializersModel(DJAModel):
+    class DeprecatedIncludedSerializersModel(DJAModel):
```
sliverc, line 349: I do not see that renaming of the classes are needed here because they are inline. I prefer to keep the test as is so to make sure there are no regressions. `deprecated_call` call needs to be added of course.

rest_framework_json_api/relations.py
```diff
@@ -274,7 +273,7 @@ def get_resource_type_from_included_serializer(self):
                 inflection.singularize(field_name),
                 inflection.pluralize(field_name),
             ]
-            includes = get_included_serializers(parent)
+            includes = getattr(parent, "included_serializers", dict())
```
sliverc, line 276: I copied this from another inline comment which got outdated by a commit but actually was not resolved:

From an API perspective this is a bit ugly that `getattr` always needs to be called instead of simply `serializer.included_serializers`. I would prefer that `included_serializers` is set to `None` when not configured so it easier to read `included_serializers` for a DJA user.

Easiest I guess is to set `included_serializers` in the serializer base class to `None`.

rest_framework_json_api/serializers.py
```diff
@@ -152,8 +153,51 @@ def validate_path(serializer_class, field_path, path):
         super(IncludedResourcesValidationMixin, self).__init__(*args, **kwargs)
 
 
+class LazySerializersDict(Mapping):
+    def __init__(self, klass, serializers):
```
sliverc, line 157: when thinking about it `klass` is actually not really descriptive - I guess `parent` would be a better name?

rest_framework_json_api/serializers.py
```diff
@@ -152,8 +153,51 @@ def validate_path(serializer_class, field_path, path):
...
+    def __repr__(self):
+        return dict.__repr__(self.serializers)
+
+
 class SerializerMetaclass(SerializerMetaclass):
-    pass
+    def __new__(cls, name, bases, attrs):
+        klass = super().__new__(cls, name, bases, attrs)
+
+        if attrs.get("included_serializers", None):
```
sliverc, line 186: a bit odd to access `attrs` again after it already has been used to create above klass.

After the API change that `included_serializers` is always set (see other inline comment), this could be accessed as `klass.included_serializers`. To make it clearer it might also be better to rename `klass` to `serializer`.

sliverc (changes requested):
I haven't forgotten about this PR but haven't really had time to look at it closely.

Seems some comments were not addressed but somehow through outdating were resolved by GitHub directly. So I moved those comment to the right code line.

Sorry if this has caused lots of notifications. The conversations which are still open are the issues which still need to be addressed. Thanks.

tests/test_utils.py
SafaAlfulaij: I remember Django was telling me that it's not possible to do this, perhaps because I moved it outside of the function.
Will try again.

SafaAlfulaij: Yup, it didn't work.

SafaAlfulaij:
I'm sorry but I can't work on this anymore. I tried an approach where I set `included_serializers` and `related_serializers` to `None` if not provided, but it resulted in a bit more complicated code, you'll find it before applying the commit "Simplify caching included serializers".

tests/test_utils.py
sliverc: I see there is a conflict between model names... Anyway this code only lives till the next major version so let's leave it like that.

rest_framework_json_api/relations.py
sliverc: I looked into this some more as well. Internally we will need to leave `getattr` anyway because DJA allows mixed serializers and DRF serializers won't have `included_serializers` set.

I'd also played around with the API to allow defaults. Quite a few changes would be needed, more than I have anticipated for this to work. I think this crosses the boundaries of this PR so let's just leave it as you had it for now. We certainly need to think about the DJA serializer API more broadly but in a different thread.

sliverc:
@SafaAlfulaij Thanks for having worked on this. I made the last minor changes and this PR should be ready now and be merged once CI is green.

sliverc (approved):

jarekwg:
@sliverc this is a pretty neat performance boost, what's the plan for the next release?

sliverc:
@jarekwg I would not overestimate this change in terms of performance because it just few imports. This is mainly a API cleanup so included serializers can be simply retrieved by `serializer.included_serializers` instead of a utility method.
In terms of next release I assume we will do one once Django 4.0 is released (according to [release notes](https://docs.djangoproject.com/en/4.0/releases/4.0/) expected in December).

auvipy:
besides django 4.0 compat new release is needed in drf as well.

PR #941, opened 2021-05-04
Added initial issue templates

sliverc:
My ideal would be that `Issues` is actually a list of work items with all the details for any contributor who wishes to work on it. All other discussion should happen in discussion. This follows the example of [DRF](https://github.com/encode/django-rest-framework/discussions/7740]

Adding issue templates will be one way forward into this direction to guide users how to report.


n2ygk (approved):
again, sorry for the delay getting to this.

PR #946, opened 2021-05-25
fix(perf): use preload includes mixin for read only model view set

anehx:
This allows the usage of `select_for_includes` and `prefetch_for_includes` on `ReadOnlyModelViewSet`


sliverc (approved):
Thanks a lot. LGTM.

PR #956, opened 2021-07-06
Remove invalid validation of default included_resources

### LINKED ISSUE #949: Validation of included_resources fails when nested

#900 introduced validation of `included_resources` but causes a regression that when one of `included_serializers` of the current serializer of called view uses `included_resources` it will fail.

I assume validation of `included_resources` should actually be done in the serializer meta class and only query parameters validated in `IncludedResourcesValidationMixin`. 

See discussion #318 where there is also some code to reproduce it.

## Checklist

sliverc:
Fixes #949 

## Description of the Change

#900 inadvertently introduced validation of default included_resources which does not work properly with nested included resources (such as when a default included resources also defines its own included resources).

This PR reverts this validation but leaving the default prefetching of default included resources as intended in PR #900. 

This is a quick fix to get this regression solved. In the long run we need to think whether it makes sense at all to do validation of include in init of the serializer anyway. But can be discussed later.

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 Note that in line with [Django REST Framework policy](http://www.django-rest-framework.org/topics/release-notes/),
 any parts of the framework not mentioned in the documentation should generally be considered private API, and may be subject to change.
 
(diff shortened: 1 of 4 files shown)


n2ygk (approved):
👍

PR #977, opened 2021-09-16
Use consistent naming for Django REST framework

sliverc:
## Description of the Change

This follows the naming DRF uses in its own [documentation](https://www.django-rest-framework.org/) with an uppercase `REST` and all lowercase `framework`.

## Checklist

- [ ] unit-test added
- [x] documentation updated


n2ygk (approved):
hmmm, I'd argue that DRF gets it wrong and the F should be Title case....

PR #981, opened 2021-09-23
Allow parser context to be None

sliverc:
## Description of the Change

Default parameter of `parser_context` was already set to `None` but would had always raised an error when it was actually `None`.

DRF parser allows `parser_context` to be `None` so this PR makes the DJA parser fully compatible with DRF API.

## Checklist

- [x] unit-test added
- [ ] documentation updated


sliverc:
Just as a follow up I discovered this issue while implementing PR #898. So it is basically an issue which can occur during testing.

n2ygk (approved):

PR #986, opened 2021-10-03
WIP: Add relationships support for pointers in errors

### LINKED ISSUE #413: Error pointer always refers to attributes
If I do a POST request with data containing a relationship:
`{
    "data": {
        "type": "comment",
        "attributes": {
            "text": "comment text"
        },
        "relationships": {
            "author": {
                "data": {
                    "id": "1 (invalid id)",
                    "type": "author"
                }
            }
        }
    }
}`
I got the following response:
`{
    "errors": [{
        "status": "400",
        "source": {"pointer": "/data/attributes/author"}, 
        "detail": "Error message"
    }]
}`
But would rather expect the following response:
`{
    "errors": [{
        "status": "400",
        "source": {"pointer": "/data/relationships/author"}, 
        "detail": "Error message"
    }]
}`
So the two responses are different when it comes to source/pointer.
Which version is the proper one? According to the specs the `pointer` should point to the associated entity in the request document.

mehdy:
Fixes #413 

## Checklist

- [ ] unit-test added
- [ ] documentation updated


mehdy:
@sliverc This is my draft attempt at solving the problem. I was wondering if you could give me some hints on whether I'm headed in the right direction or not?

sliverc:
Yes you are on the right track. `included_serializers` is optional though so not the right way to determine whether a field is a relationship. See [here](https://github.com/django-json-api/django-rest-framework-json-api/blob/master/rest_framework_json_api/renderers.py#L68) how an attribute is distinguished from relationship in the renderer.

mehdy:
@sliverc Great! Thanks!

mehdy:
@sliverc 
I think this is ready for review.
Just let me know if I should add it to `CHANGELOG.md` and/or the docs.

mehdy:
@sliverc I couldn't understand why `codecov` is failing. Should I do something?

rest_framework_json_api/utils.py
```diff
@@ -394,9 +411,14 @@ def format_drf_errors(response, context, exc):
             errors.extend(format_error_object(message, "/data", response))
     # handle all errors thrown from serializers
     else:
+        serializer = get_serializer_from_context(context)
```
sliverc, line 414: I assume you added this complex method `get_serializer_from_context` because of `test_errors.py`?

I have a feeling the test setup in `test_errors.py` is not correct as an `APIView` does not have a serializer and it just fakes to have one. Basically only views which derive from `rest_framework.generics.GenericAPIView` have serializer support. And then a simple call to `view.get_serializer()` will do.

The question remains though what do we do with views which directly derive from `APIView`?. We cannot correctly determine whether it is a relationship or an attribute. In the [JSON:API spec](https://jsonapi.org/format/#errors) pointer is optional.

So I guess the best strategy will be if the view does not derive from `GeneriAPIView` then simply set `pointer` to `None`.

rest_framework_json_api/utils.py
```diff
@@ -368,6 +368,23 @@ def get_relation_instance(resource_instance, source, serializer):
...
+    if hasattr(context["view"], "get_serializer"):
+        return context["view"].get_serializer()
+    elif hasattr(context["view"], "get_serializer_class"):
+        return context["view"].get_serializer_class()()
+    elif hasattr(context["view"], "serializer_class"):
+        return context["view"].serializer_class()
+
+
+def get_relationship_fields(fields):
+    return {
```
sliverc, line 381: As this is a public API I do not think how helpful this is as a utility on other cases. Actually in renderers.py the snippet ` isinstance(field, (relations.RelatedField, relations.ManyRelatedField)` is used twice. So if we had a method `is_relationship_field` those could be replaced and below in `format_drf_errors` you could do a simple dict comprehension.

In your code below you can then simply do a

example/tests/test_views.py
```diff
@@ -124,6 +124,29 @@ def test_relationship_view_errors_format(self):
         assert "data" not in result
         assert "errors" in result
 
+    def test_relationship_view_errors_has_correct_pointers(self):
```
sliverc, line 127: `RelationshipView` is a class in `views.py` but that is not what you are testing here. This should rather be added to `test_errors.py`.

AUTHORS
```diff
@@ -42,3 +42,4 @@ Tim Selman <timcbaoth@gmail.com>
 Tom Glowka <glowka.tom@gmail.com>
 Ulrich Schuster <ulrich.schuster@mailworks.org>
 Yaniv Peer <yanivpeer@gmail.com>
+Mehdy Khoshnoody <mehdy.khoshnoody@gmail.com>
```
sliverc, line 45: we keep this file alphabetically

sliverc (changes requested):
Thanks for working on this. Please see my inline comments.

rest_framework_json_api/utils.py
mehdy: I agree! will do some refactors on this as well.

example/tests/test_views.py
mehdy: Sure!

mehdy:
@sliverc Resolved all the issues you mentioned.

sliverc (changes requested):
This PR is shaping up nicely. See some small comments inline.

mehdy:
@sliverc Done as well. Thanks for the tips.

example/tests/test_errors.py
```diff
@@ -1,13 +1,33 @@
...
+    @property
+    def queryset(self):
+        return self
+
+    def create(self, *args, **kwargs):
+        pass
+
+    def get(self, pk):
+        if isinstance(pk, int):
+            return self
```
sliverc, line 27: Odd... I thought I'd commented here before but somewhat cannot find this comment anymore.

Did you consider using `example.models.Author` instead of this mock? Issue here is that in JSON:API the id is actually always a string so line 27 will never be covered (reason why CodeCov is failing).

PR #993, opened 2021-10-10
Deprecated usage of `type` field name

sliverc:
Deprecating of #376 

## Description of the Change

Partial support for field name `type` was added in #376 but only for non polymorphic fields.

However as per [specification](https://jsonapi.org/format/#document-resource-object-fields) `type` must not be a field name and therefore must be forbidden in DJA as well.

Some dependents might depend on being allowed to have a field name `type` so deprecating it now and remove it in next major version.

## Checklist

- [x] unit-test added
- [ ] documentation updated


auvipy:
won;t this break that PR? can any alternative name could be used?

auvipy:
ok. fair enough. can field_type be used instead? don't get annoyed by my questions btw

PR #1003, opened 2021-10-27
Resolved inflecting component attributes in OpenAPI SchemaGenerator

sha016:
Fixes #967

## Description of the Change
applied format_field_name to 3 instances of field.field_name and update snapshot

## Checklist

- [x] PR only contains one change (considered splitting up PR)
- [x] unit-test added
- [ ] documentation updated
- [x] `CHANGELOG.md` updated (only for user relevant changes)
- [x] author name in `AUTHORS`

PR #1005, opened 2021-10-28
Migrate to syrupy snapshot testing

sliverc:
## Description of the Change

[snapshottest](https://github.com/syrusakbary/snapshottest) is not really maintained anymore and causes issue in showing the diff as recent versions changed ordering (see #1003). Therefore suggesting that we move to [syrupy](https://github.com/tophat/syrupy) which is well maintained, has better pytest integration, has a proper serialization format (instead of simply writing python code as snapshottest does) and is extensible.

For now this PR is a simple migration without any changes in testing. In the future also when working on #468 we might create our own syrupy extensions to simplify our testing (e.g. in test_openapi there a quite a few json.dumps).

## Checklist

- [ ] unit-test added
- [ ] documentation updated


n2ygk (approved):
Looks good. Sorry for the long delay.

PR #1010, opened 2021-11-18
Added support for Django 4.0

### LINKED ISSUE #1000: Add support for Django 4.0

Django 4.0a1 has been [released](https://www.djangoproject.com/weblog/2021/sep/21/django-40-alpha-1-released/). As it is marked as feature freeze we can also add it to our test suite.

DRF has also already [added](https://github.com/encode/django-rest-framework/pull/8178) it. As a reference to implement this have a look at #908

## Checklist

sliverc:
Fixes #1000 

## Description of the Change

Add support for Django 4.0 based on the first beta version (feature freeze already happened so this should be safe.)

## Checklist

- [ ] unit-test added
- [x] documentation updated

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -89,7 +89,7 @@ Requirements
 ------------
 
 1. Python (3.6, 3.7, 3.8, 3.9)
-2. Django (2.2, 3.0, 3.1, 3.2)
+2. Django (2.2, 3.0, 3.1, 3.2, 4.0)
 3. Django REST framework (3.12)
 
 We **highly** recommend and only officially support the latest patch release of each Python, Django and REST framework series.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ any parts of the framework not mentioned in the documentation should generally b
 
 ## [Unreleased]
 
+### Added
+
+* Added support for Django 4.0.
+
 ### Fixed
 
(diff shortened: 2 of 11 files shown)


README.rst
```diff
@@ -89,7 +89,7 @@ Requirements
 ------------
 
 1. Python (3.6, 3.7, 3.8, 3.9)
-2. Django (2.2, 3.0, 3.1, 3.2)
+2. Django (2.2, 3.0, 3.1, 3.2, 4.0)
```
auvipy, line 92: django 3.0 is EOL already so can be dropped from CI/Docs

sliverc: We will remove this in the next DJA major version 5.0.0 as such a change breaks backwards compatible. As in current master we have new deprecation warnings the next DJA version will be 4.3.0 though (in major releases we do not want to introduce new deprecation).

On a side note, I appreciate that you keep an eye on dependencies being EOL. Instead of commenting in an unrelated PR (this PR is about adding Django 4.0 support) I would prefer that you create an issue like "Remove support for EOL Django 3.0". This can then be used as an action item where as a comment in an unrelated PR might just be overlooked and forgotten about. Thanks for understanding.

auvipy: ok

n2ygk (approved):

PR #1011, opened 2021-11-18
Use one GitHub runner per Python version

sliverc:
## Description of the Change

* Moving from tox-gh-actions to tox-py (the same as DRF is already using)
* Move configuration what tests are allowed to fail into tox instead of GitHub settings
* Simplification of configuration as tox has the lead and no additional configuration is needed for actions.

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,17 +5,12 @@ jobs:
   test:
     name: Run test
     runs-on: ubuntu-latest
-    continue-on-error: ${{ matrix.django-rest-framework == 'master' }}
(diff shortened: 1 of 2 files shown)


n2ygk (approved):

sliverc:
Support for Django 4.0 (see #1010) would add more complexity in the GitHub action matrix as it does not support all Python versions anymore. With this change supporting new versions of Django or DRF will be easier as only tox needs to be configured and no additional configuration anymore in GitHub project settings.

PR #1027, opened 2021-12-14
Updated urls to use path and re_path as is default in Django 2.2.

sliverc:
## Description of the Change

`django.urls.conf.url` is only there for Python 3.5 support which we already deleted anyway.

## Checklist

- [ ] unit-test added
- [ ] documentation updated


auvipy (approved):
lgtm

n2ygk (approved):

PR #1028, opened 2021-12-14
Adjusted to only use f-strings for slight performance improvement

sliverc:
## Description of the Change

Since we dropped Python 3.5 support let's make use of the new f-string feature in Python 3.6+ for code improvement and readability.

## Checklist

- [ ] unit-test added
- [ ] documentation updated


n2ygk (changes requested):
@sliverc Please fix the CHANGELOG.md conflicts resulting from the prior PR being merged.

n2ygk (approved):

PR #1039, opened 2021-12-28
Removed support for Python 3.6

sliverc:
## Description of the Change

Removing Python 3.6 support as it is [end of life](https://endoflife.date/python).

## Checklist

- [ ] unit-test added
- [x] documentation updated

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@ def read(*paths):
     """
     Build a file path from paths and return the contents.
     """
-    with open(os.path.join(*paths), "r") as f:
+    with open(os.path.join(*paths)) as f:
         return f.read()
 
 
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -86,7 +86,6 @@ def get_package_data(package):
         "Operating System :: OS Independent",
         "Programming Language :: Python",
         "Programming Language :: Python :: 3",
-        "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9",
```
n2ygk, line 91: Add 3.10?

n2ygk (changes requested):

setup.py
sliverc: that is already added see a bit further below.

n2ygk: Got caught by the context diff...

n2ygk (approved):

PR #1048, opened 2022-02-06
Adhered to field naming format setting when generating schema

jokiefer:
## Description of the Change

Added additional field name formatting which got missed out in PR #1003

## Checklist

- [x] unit-test added
- [ ] documentation updated


sliverc (changes requested):
Thanks for working on this. Could you please also add a test which adds those cases?

I guess easiest is to extend the `AuthorViewSet` so it covers the cases which are fixed in this PR. The tests in `test_openapi.py` use snapshots so once the view has all the different cases added the snapshot only needs to be updated (pytest --snapshot-update).

sliverc:
@jokiefer Wanted to follow up whether you are still interested in working on this PR?

sliverc:
I had some time at hand and wanted to make sure that this PR lands in the next version (which is hopefully released soon, waiting for DRF to release a new version supporting Django 4.1). I added a test and made minor adjustments to the PR.

When adding the test, I noticed that there was actually an error in the code. It called `format_field_names` instead of `format_field_name` so this did not work. Once again, a great example of how important tests are. 

Anyway. Thanks for bringing this up and working on it.

Issue #332, opened 2017-03-09
Sparse fields nulling meta fields instead of removing them


approxit:
Hi,

JSON API specification doesn't say anything about meta fields while using sparse fields, but with `drf-json-api` 2.1.1 implementation we actually have control over meta files presence in response. The thing is - we can't get rid of meta fields totally. Values of meta fields are `null`ed, and this is a quite impact for the actual information.

Let's take example:
```
GET /resource1/1/
{
    "data": {
        "type": "resource1",
        "id": "1",
        "attributes": {
            "some_field": 1,
            "some_other_field": 2
        },
        "meta": {
            "some_meta_field": 3,
            "some_other_meta_field": 4
        }
    }
}
```
```
GET /resource1/1/?fields[resource1]=some_field
{
    "data": {
        "type": "resource1",
        "id": "1",
        "attributes": {
            "some_field": 1
        },
        "meta": {
            "some_meta_field": null,
            "some_other_meta_field": null
        }
    }
}
```
```
GET /resource1/1/?fields[resource1]=some_meta_field
{
    "data": {
        "type": "resource1",
        "id": "1",
        "attributes": {},
        "meta": {
            "some_meta_field": 3,
            "some_other_meta_field": null
        }
    }
}
```

Take note that `null` value is a totally natural and legal value in application logic. Here we receives wrong information. It's a bug for me. `drf-json-api` must decide to drop limiting meta fields by sparse fields or to remove meta fields totally by sparse fields.

PR #1049, opened 2022-02-07
differs correctly between `meta` properties and `attribute` propertie…

jokiefer:
…s in component schemas.

Fixes #

## Checklist

~- [ ] unit-test added~
~- [ ] documentation updated~

### DIFF
--- a/rest_framework_json_api/schemas/openapi.py
+++ b/rest_framework_json_api/schemas/openapi.py
@@ -648,6 +648,7 @@ class AutoSchema(drf_openapi.AutoSchema):
         required = []
         attributes = {}
         relationships = {}
+        meta = {}
 
         for field in serializer.fields.values():
             if isinstance(field, serializers.HyperlinkedIdentityField):
@@ -683,7 +684,12 @@ class AutoSchema(drf_openapi.AutoSchema):
                 schema["description"] = str(field.help_text)
             self.map_field_validators(field, schema)
 
-            attributes[format_field_name(field.field_name)] = schema
+            if field.field_name in list(
+                getattr(serializer.Meta, "meta_fields", list())
+            ):
+                meta[format_field_name(field.field_name)] = schema
+            else:
+                attributes[format_field_name(field.field_name)] = schema
 
         result = {
             "type": "object",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 Note that in line with [Django REST framework policy](https://www.django-rest-framework.org/topics/release-notes/),
 any parts of the framework not mentioned in the documentation should generally be considered private API, and may be subject to change.
 
+## [Unreleased] - 2022-x-xx
+
+### Fixed
+
+* differs correctly between `meta` properties and `attribute` properties in component schemas.
+
 ## [5.0.0] - 2022-01-03
 
 This release is not backwards compatible. For easy migration best upgrade first to version


rest_framework_json_api/schemas/openapi.py
```diff
@@ -683,7 +684,12 @@ def map_serializer(self, serializer):
                 schema["description"] = str(field.help_text)
             self.map_field_validators(field, schema)
 
-            attributes[format_field_name(field.field_name)] = schema
+            if field.field_name in list(
+                getattr(serializer.Meta, "meta_fields", list())
+            ):
+                meta[format_field_name(field.field_name)] = schema
```
sliverc, line 690: Thanks. Nice catch this can certainly be a problem.

meta dict is never used so I guess meta fields could simply be ignored? further up with a `if meta_field then continue` for instance. Otherwise would need to think about changing how we represent meta in the schema in general.

Also a test which test this use case will be needed for merging.

sliverc (changes requested):

rest_framework_json_api/schemas/openapi.py
jokiefer: i wouldn't remove it from the schema. 

Example usecase:
```json
  "components": {
    "schemas": {
      "WebMapService": {
        "type": "object",
        "properties": {
          "type": {
            "type": "string",
            "description": "The [type](https://jsonapi.org/format/#document-resource-object-identification) member is used to describe resource objects that share common attributes and relationships.",
            "enum": [
              "WebMapService"
            ]
          },
          "attributes": {
            "type": "object",
            "properties": {
            }
          },
          "relationships": {
          },
          "meta": {
            "type": "object",
            "properties": {
              "stringRepresentation": {
                "type": "string",
                "readOnly": true,
                "title": "The string representation of the object it self"
                "description": "This representation could be used to display a short title of the object"
              }
            }
          }
      }
    }
```

The `stringRepresentation`  could be the string from the `__str__` function of the model object. This would be usefull for several use cases, such as [autocomplete widgets](https://ant.design/components/auto-complete/).

Sure, the `stringRepresentation` could be part of the regular `attributes` of the object, but i dont know if it is the best place for it.

sliverc: @n2ygk You have more in-depth knowledge about openapi. What is your take on the meta issue?

jokiefer: if sparse fields are used, the meta object is still part of the response with possible null data. For that reason i would say it is better to place it in the attributes. Then it is possible to get only the stringRepresentation attribute in autocomplete cases.

sliverc: This is actually a bug see #332 and needs to be fixed.

sliverc: not really... what was the reason to represent meta on a resource level as stringRepresentation? Couldn't there be a more detailed representation as with attributes as all meta fields are actually serializer fields? Or did I overlook something here?

n2ygk: Spending all of 30 seconds looking at this, it looks like confusing Django Models' Meta attribute with the optional meta jsonapi field: https://jsonapi.org/format/#document-meta

They are not related.

sliverc: It is a bit confusing that within DJA we sometimes mix up the API with DRF but this is a discussion for another time... :smile: but actually in this case `meta_fields` even though defined in `Meta` class is a DJA feature ([see docs](https://django-rest-framework-json-api.readthedocs.io/en/stable/usage.html#meta))

https://github.com/django-json-api/django-rest-framework-json-api/blob/c46d495e38bf46c09277ce34230b966a4cd0ec1d/rest_framework_json_api/renderers.py#L421)

It works the way that serializer fields are used to build the JSON:API spec resource meta (not the top level which is build manually by overwriting `get_root_meta`). So as i see the resource meta class could be defined in more detail where as the root meta cannot.

sliverc: As a follow up as of my explanation above in my opinion can the meta fields be assigned to the result schema further below with something like this:

```python
        if relationships:
            result["properties"]["meta"] = {
                "type": "object",
                "properties": meta,
            }
```

PR #1065, opened 2022-05-30
Support many related field

### LINKED ISSUE #912: ResourceRelatedField.source doesn't allow nested paths
Compared to how DRF allow nested paths in source:
https://github.com/encode/django-rest-framework/blob/master/rest_framework/fields.py#L401

DJA doesn't allow it:
https://github.com/django-json-api/django-rest-framework-json-api/blob/51daed1a738aadb1536a6a3cbe787cbd88daaf05/rest_framework_json_api/utils.py#L204

Examples here: https://www.django-rest-framework.org/api-guide/fields/#source
`EmailField(source='user.email')`

bashlyss:
Fixes #912 

## Description of the Change

I'm not sure if this specifically fixes the entirety of the first issue linked, I haven't had any problems with nested relationships that aren't M2M. However, this solves nesting for M2M on non declared fields. (Requires workaround for DRF in `ModelSerializer.build_field` to handle dotted field sources. Separate bug in that repo)

Slightly contrived example documenting equivalent structure to my local issue
```
class CourseSectionSerializer(ModelSerializer):
    class Meta:
        model = Student
        fields = ("students",)  # M2M on "course" related field
        extra_kwargs = {
            "students": {"source": "course.students"}
        }
```

Currently, this crashes on the line highlighted in the issue linked. It appears that somewhere along the line DRF updated how many relationships work and the previous `elif` above the code I specified doesn't lookup the current attributes properly. I'm not sure if there is a version/edge case that still can flow through it or not so I left it as is.

It's my first time adding something to this repo, so please direct me for where an appropriate unit test should live or any docs to update. Right now this is a quick in browser edit to fix a bug I'm having in development.

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -2,6 +2,7 @@ Adam Wróbel <https://adamwrobel.com>
 Adam Ziolkowski <adam@adsized.com>
 Alan Crosswell <alan@columbia.edu>
 Anton Shutik <shutikanton@gmail.com>
+Ashley Loewen <github@ashleycodes.tech>
 Asif Saif Uddin <auvipy@gmail.com>
 Beni Keller <beni@matraxi.ch>
 Boris Pleshakov <koordinator.kun@gmail.com>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ any parts of the framework not mentioned in the documentation should generally b
 ### Fixed
 
 * Fixed invalid relationship pointer in error objects when field naming formatting is used.
+* Properly resolved related resource type when nested source field is defined.
 
 ## [5.0.0] - 2022-01-03
 
--- a/rest_framework_json_api/utils.py
+++ b/rest_framework_json_api/utils.py
@@ -219,6 +219,12 @@ def get_related_resource_type(relation):
         # For ManyToMany relationships, get the model from the child
         # serializer of the list serializer
         relation_model = relation.child.Meta.model
+    elif (
+        hasattr(relation, "child_relation")
+        and hasattr(relation.child_relation, "model")
+    ):
+        # For ManyRelatedField relationships, get the model from the child relationship
+        relation_model = relation.child_relation.model
     else:
         parent_serializer = relation.parent
         parent_model = None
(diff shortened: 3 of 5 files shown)


sliverc:
Thanks for working on this. I am not sure if this is the right fix though. We have to play around with it a little. So best is to add a test case first and then see what works. Best add a new test case for nested M2M source in the file `tests/test_utils.py`. There are already some tests at the bottom of the file starting with `test_get_related_resource_type`.

And then what might work and simplify the code is replacing the following code bit:
```python
        if parent_model is not None:
            if relation.source:
                if relation.source != "*":
                    parent_model_relation = getattr(parent_model, relation.source)
                else:
                    parent_model_relation = getattr(parent_model, relation.field_name)
            else:
                parent_model_relation = getattr(
                    parent_model, parent_serializer.field_name
                )

```

with something like this 
```python
 if parent_model is not None:
    parent_model_relation = relation.get_attribute(parent_model)
```

Needs to be tested whether this works.

bashlyss:
I've tried that and it doesn't work as far as I can tell, at least for the `ManyRelatedField` relation that I'm currently working with. When I have time to play around with it I will get the test in and see. Like I said, it may actually be a bit of a different (but similar) bug. 

In my environment this code resolves the ManyRelatedField issue (unsure if there is a larger global one in the current version). That is using against 4.2.1 rather than 5.0.0 and DRF 3.12.4.

bashlyss:
@sliverc either I didn't fully understand what you intended there (get_attribute isn't a method as it happens) or it doesn't work. Adjusted my code slightly and got the tests in.

Since this is an internal, I assume no changelog/docs changes are necessary?

sliverc (approved):
I had a deep dive into this issue again. The `get_related_resource_type` is very complex and therefore error-prone. Simpler would be to get the actual value of a field and resolve the resource type from the value. It would mean though to rewrite the renderer quite a bit, which is risky and best done in a complete new renderer class.

This PR is a simple bugfix though, so let us merge it as you proposed it and rethink the renderer issue at a later point. As I was looking over the code, I did some small modifications and added a changelog entry (as this actually fixes a bug).

Thanks for your work on it. Very much appreciated.

PR #1074, opened 2022-07-20
Avoid overwriting of pointer when formatting error object

### LINKED ISSUE #399: Avoid overwriting of pointer when using custom errors
The code that produces JSON:API error objects assumes that all fields being validated are under `/data/attributes`:

https://github.com/django-json-api/django-rest-framework-json-api/blob/b0257c065ac2439036ac0543139e7f5913755177/rest_framework_json_api/utils.py#L339

and there doesn't appear to be a way to override the `pointer` value when specifying the field on a serializer.

Use case: I would like to be able to specify a serializer for the top-level `"meta"` field and have any validation errors on that data generate responses properly populated with `"source": {"pointer": "/meta/foo"}`.

koriaf:
Fixes #399

Do not overwrite source.pointer field if it's already there; even if our pointer is calculated differently. Sorry for not updating the tests; tested it manually and it works for me and existing tests pass

- [x] unit-test added
- [x] documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ any parts of the framework not mentioned in the documentation should generally b
 
 * Fixed invalid relationship pointer in error objects when field naming formatting is used.
 * Properly resolved related resource type when nested source field is defined.
+* Prevented overwriting of pointer in custom error object
 
 ### Added
 
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -237,13 +237,28 @@ class MyViewset(ModelViewSet):
 
 ```
 
-### Exception handling
+### Error objects / Exception handling
 
 For the `exception_handler` class, if the optional `JSON_API_UNIFORM_EXCEPTIONS` is set to True,
 all exceptions will respond with the JSON:API [error format](https://jsonapi.org/format/#error-objects).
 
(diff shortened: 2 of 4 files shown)


sliverc (changes requested):
Thanks for working on this. The change looks good, we will need to have a test though, as otherwise we cannot be sure that this issue gets introduced again in the future.

A test could be added to "tests/test_utils.py" in a form like the following (method is just off the top of my head):
```python
@pytest.mark.parametrize(
    "message,pointer,response,result",
    [
        (
            "Test Message",
            "testing pointer",
            Response(status.HTTP_200_OK),
            ["how result should look like"],
        ),
    ],
)
def test_format_error_object(message, pointer, response, result):
    assert result == format_error_object(message, pointer, response)
```

You will only need to add one test case (resp. updating the item in parametrize) which covers what you are fixing.

Additionally, please also add yourself to the AUTHORS file and update the CHANGELOG.md. Let me know if you need additional support.

sliverc (approved):
So I had some time at hand to add a test and polish this PR with docs update etc. so it is ready to merge now. This gives you an example for a next time how a test could look like. Thanks for your contribution anyway. It is good to get this fixed, as this is quite a neat feature to have custom error object pointers.

koriaf:
Thanks for writing the test and doc, sorry for not being able to handle it myself :-)

PR #1090, opened 2022-09-22
Added support for DRF 3.14 and dropping support for DRF 3.12.

sliverc:
## Description of the Change

As per our policy, adding support for DRF 3.14 and dropping support for DRF 3.12 as we support the two last releases.

## Checklist

- [x] unit-test added
- [x] documentation updated


sliverc:
Once this is merged, I will follow up with a new release 6.0.0 unless there are any objections.

n2ygk (approved):

PR #1094, opened 2022-10-16
Order

auvipy:
Fixes #

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/rest_framework_json_api/metadata.py
+++ b/rest_framework_json_api/metadata.py
@@ -1,5 +1,3 @@
-from collections import OrderedDict
-
 from django.db.models.fields import related
 from django.utils.encoding import force_str
 from rest_framework import serializers
@@ -65,7 +63,7 @@ class JSONAPIMetadata(SimpleMetadata):
         )
 
     def determine_metadata(self, request, view):
-        metadata = OrderedDict()
+        metadata = {}
         metadata["name"] = view.get_view_name()
         metadata["description"] = view.get_view_description()
         metadata["renders"] = [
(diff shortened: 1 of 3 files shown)


auvipy:
probably not a feasible idea yet, or some minor adjustment would fix the failed tests

sliverc:
Thanks for the hinter. this is something to consider but I will need to do more research what the drawbacks could be of this change. however your PR does not change it properly as the ordered dict constructur expected a list of tuples. a dict uses the key:value syntax. So you would need to change that as well.

auvipy:
thanks, I also need some more research / study here.

sliverc:
I had a look into this a bit more and found [this article](https://realpython.com/python-ordereddict/) which summarizes the difference between Py 3.7+ dict and `OrderedDict` quite well.

The main points to use `OrderedDict` are:
1. Intent signaling: If you use OrderedDict over dict, then your code makes it clear that the order of items in the dictionary is important. You’re clearly communicating that your code needs or relies on the order of items in the underlying dictionary.
2. Control over the order of items: If you need to rearrange or reorder the items in a dictionary, then you can use .move_to_end() and also the enhanced variation of .popitem().
3. Equality test behavior: If your code compares dictionaries for equality, and the order of items is important in that comparison, then OrderedDict is the right choice.

In the JSON:API spec there does not seem to be any specific ordering. So if an API returns attributes as first key and then relationships or the other way around, the two resources are still considered the same. Comparing two objects where the order of the keys are different should actually still be equal, which is not the case when using `OrderedDict`.

In that case, it could even be considered a bug that we use `OrderedDict`. Although of course it is great when the order of the keys are always the same for better readability, which is now the case with Python 3.7+ dict anyway.

I try to clean up your PR here, so the tests are green. If you get to it first, feel free to update it yourself.

PR #1098, opened 2022-10-26
Added support to overwrite serializer methods in schema class

sliverc:
## Description of the Change

This change makes the DJA schema class compatible with the DRF change introduced with https://github.com/encode/django-rest-framework/pull/7424

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ any parts of the framework not mentioned in the documentation should generally b
 
 * Added support for Python 3.11.
 
+### Changed
+
+* Added support to overwrite serializer methods in customized schema class
+
 ## [6.0.0] - 2022-09-24
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


sliverc:
Thanks for merging. This changed actually got integrated 2 years ago and I have just recently stumbled upon it. There is a recent change https://github.com/encode/django-rest-framework/pull/7493 which will affect DJA as well, so it seems minor changes are still merged, but nothing major.

auvipy:
> Thanks for merging. This changed actually got integrated 2 years ago and I have just recently stumbled upon it. There is a recent change [encode/django-rest-framework#7493](https://github.com/encode/django-rest-framework/pull/7493) which will affect DJA as well, so it seems minor changes are still merged, but nothing major.

you are right, but crucial changes for thing important in DRF will also be merged. I know this from conversation with Tom Christie. DRF will deprecate core schema, but improve OpenAPI support.

PR #1101, opened 2022-11-08
Introduce flake8 bugbear plugin

sliverc:
## Description of the Change

Bugbear is a plugin for flake8 finding likely bugs and design problems. This PR adds it to DJA and already found some potential issues.

Also, it introduces an optional “line length too long” error which is more pragmatic than the E507 which we had disabled.
Currently, we disable line length altogether and leave it up to black to make it shorter. However, black does not format strings and comments etc. so some lines end up being very long. Bugbear will only flag lines which are longer than ~10% of the maximum length to solve this problem, but not to make the error becoming annoying.

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/rest_framework_json_api/views.py
+++ b/rest_framework_json_api/views.py
@@ -162,6 +162,10 @@ class RelatedMixin:
         if "related_field" in self.kwargs:
             field_name = self.get_related_field_name()
 
+            assert hasattr(parent_serializer_class, "included_serializers") or hasattr(
+                parent_serializer_class, "related_serializers"
+            ), 'Either "included_serializers" or "related_serializers" should be configured'
+
             # Try get the class from related_serializers
             if hasattr(parent_serializer_class, "related_serializers"):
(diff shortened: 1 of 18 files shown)


rest_framework_json_api/views.py
```diff
@@ -129,7 +129,10 @@ def get_queryset(self, *args, **kwargs):
 
 class RelatedMixin:
     """
-    This mixin handles all related entities, whose Serializers are declared in "related_serializers"
+    Mixing handling related links.
```
n2ygk, line 132: I don't understand what this added comment is about

n2ygk (approved):
Minor comment question but not worth holding back merging for.

rest_framework_json_api/views.py
sliverc: I only reformatted that comment into a proper docstring to satisfy bugbear. The comment is really not that clear and helpful, though. Added it to my list TODO to change it (most likely with a general docstring update).

PR #1107, opened 2022-11-16
Adjusted security policy

sliverc:
## Description of the Change

Finally, GitHub introduced a way to [report security vulnerability privately](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing/privately-reporting-a-security-vulnerability). This makes the Google Group obsolete, which I found less than ideal, as it is just another system we need to use.

Adjusted security policy accordingly.

## Checklist

- [ ] unit-test added
- [x] documentation updated


n2ygk (approved):
Nice to know they've added this.

PR #1113, opened 2022-12-08
Unified tox configuration

sliverc:
## Description of the Change

* Followed example of DRF using -f cmd line of tox v4 instead of tox-py
* For linting jobs updated to Python 3.8 as will be required by future updates of Flake8
* Added missing black job

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,9 +20,9 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install tox tox-py
+          pip install tox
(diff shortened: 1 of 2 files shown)


sliverc:
Something seems wrong with the Python 3.7 pipeline as it is now also testing Django 4.0 which it should not. I have to look into this again before this PR is ready.

sliverc:
I have found the error. With tox v4, the too broad definition of ignore test outcomes caused the issue that some testenv were run which should not have. I updated it accordingly and now it works as expected even with v4.

n2ygk (approved):

PR #1124, opened 2023-01-30
Refactored sort query parameter handling in schema generation

### LINKED ISSUE #1120: Incorrect handling of `sort` field in schema generation

I believe that the handling of the `sort` field in the schema generation is currently incorrect. (Please correct me if I'm wrong.)

The custom `AutoSchema` class calls [`self._get_sort_parameters(path, method)`]((https://github.com/django-json-api/django-rest-framework-json-api/blob/085554e28c663d9a336bf7174c58d81925d9d650/rest_framework_json_api/schemas/openapi.py#L425)) for all `GET` & `HEAD` requests:

```python
if method in ["GET", "HEAD"]:
    ...
    parameters += self._get_sort_parameters(path, method)
    ...
```

(_Issue 1_) The JSON:API spec says, "A server **MAY** choose to support requests to sort resource _collections_ according to one or more criteria (“sort fields”)." The way that this is currently handled will naively add the sort query parameter irrespective of whether or not `rest_framework_json_api.filters.OrderingFilter` is enabled. Also, this will add a `sort` query parameter for requests for single resources as well (though I think DRF also does this).

(_Issue 2_) The Django REST Framework's `OrderingFilter` class implements a method for `get_schema_operation_parameters` that will also [add a `sort` query parameter to the schema for the operation](https://github.com/encode/django-rest-framework/blob/0618fa88e1a8c2cf8a2aab29ef6de66b49e5f7ed/rest_framework/filters.py#L322). This leads to duplicate `sort` query parameters listed for the operation.

```python
def get_schema_operation_parameters(self, view):
        return [
            {
                'name': self.ordering_param,
                'required': False,
                'in': 'query',
                'description': force_str(self.ordering_description),
                'schema': {
                    'type': 'string',
                },
            },
        ]
``` 

## Current Workaround

I've currently worked around this by overriding the `OrderingFilter` class:
```python
from rest_framework_json_api.filters import OrderingFilter as DJAOrderingFilter

class OrderingFilter(DJAOrderingFilter):
    def get_schema_operation_parameters(self, view):
        return [{"$ref": "#/components/parameters/sort"}]
```

...and the `AutoSchema` class:
```python
from rest_framework_json_api.schemas.openapi import AutoSchema as DJAAutoSchema

class AutoSchema(DJAAutoSchema):
        def _get_sort_parameters(self, path, method):
            return []
```

...and configuring these in my `REST_FRAMEWORK` config so that the `sort` parameter is conditionally added based on whether or not the `OrderingFilter` is enabled, but still uses the schema provided by this lib. Just thought I'd include this as a potential fix to the problem. However, this patch will still include a `sort` query parameter for requests for single resources (mentioned above).

atkallie:
Fixes #1120 

## Description of the Change

Refactored `sort` query parameter handling in order to avoid duplicate declarations in the generated schema for each endpoint.

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -21,6 +21,7 @@ Jonas Kiefer <https://github.com/jokiefer>
 Jonas Metzener <jonas.metzener@adfinis.com>
 Jonathan Senecal <contact@jonathansenecal.com>
 Joseba Mendivil <git@jma.email>
+Kal <kal+oss@tedspot.com>
 Kevin Partington <platinumazure@gmail.com>
 Kieran Evans <keyz182@gmail.com>
 Léo S. <leo@naeka.fr>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ any parts of the framework not mentioned in the documentation should generally b
 
 * Added support to overwrite serializer methods in customized schema class
 
+### Fixed
+
+* Refactored handling of the `sort` query parameter to fix duplicate declaration in the generated schema definition
+
 ## [6.0.0] - 2022-09-24
 
 ### Fixed
--- a/example/tests/__snapshots__/test_openapi.ambr
+++ b/example/tests/__snapshots__/test_openapi.ambr
@@ -273,10 +273,7 @@
         "$ref": "#/components/parameters/fields"
       },
       {
-        "$ref": "#/components/parameters/sort"
-      },
-      {
-        "description": "Which field to use when ordering the results.",
+        "description": "[list of fields to sort by](https://jsonapi.org/format/#fetching-sorting)",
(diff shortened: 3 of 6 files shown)


sliverc:
Looks good to me. Could you add yourself to AUTHORS and add a changelog entry? Then we will be ready to merge it.

sliverc:
Great, thanks. I will merge this. In terms of `sort` on single resources, I have a feeling there won't be an easy fix and we should have a broader discussion. Could you open a discussion explaining the issue, potential workarounds, etc.? We can then discuss there what a way forward is.

sliverc (approved):

Issue #1126, opened 2023-02-09
Unable to serialize without a model: `dict` object has no attribute `pk`


axieum:
It appears that the JSON renderer expects every serializer to be a `ModelSerializer`, preventing regular `Serializer` use without a model.

**Reproduction**:

```py
from typing import TYPE_CHECKING

from rest_framework.permissions import AllowAny
from rest_framework.fields import CharField, Field
from rest_framework.views import Response
from rest_framework.viewsets import GenericViewSet
from rest_framework_json_api.serializers import Serializer

from .serializers import HealthStatusSerializer

if TYPE_CHECKING:
    from rest_framework. request import Request


class HealthStatusSerializer(Serializer):
    """The health status of all services"""

    def get_fields(self) -> dict[str, Field]:
        return {"Database: default": CharField()}  # for demo purposes - usually populate from installed health checks

    class Meta:
        resource_name = "HealthStatus"


class HealthStatusAPIView(GenericViewSet):
    """A health status view"""

    serializer = HealthStatusSerializer
    permission_clases = [AllowAny]
    pagination_class = None  # remove pagination
    filter_backends = []  # remove default filters

    def retrieve(request: Request, **kwargs) -> Response:
        """Retrieves the health status of all services"""

        errors: bool = False  # for demo purposes - usually check status of installed health checks
        return Response(
            self.get_serializer({"Database: default": "working"}).data,
            status_code=200 if not errors else 500,
        )
```

**Stacktrace**:

```
Traceback (most recent call last):
  File ".../django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
  File ".../django/core/handlers/base.py", line 220, in _get_response
    response = response.render()
  File ".../django/template/response.py", line 114, in render
    self. content = self.rendered_content
  File ".../rest_framework/response.py", line 70, in rendered_content
    ret = renderer. data, accepted_media_type, context)
  File ".../rest_framework_json_api/renderers.py", line 602, in render
    json_api_data = self.build_json_resource_obj(
  File ".../rest_framework_json_api/renderers.py", line 464, in build_json_resource_obj
    encoding. force_str(resource_instance.pk) if resource_instance else None,
AttributeError: 'dict' object has no attribute 'pk'
```

**Suspected code**:

https://github.com/django-json-api/django-rest-framework-json-api/blob/735ce291d7385fea39b5a50c107449c2d8fdd3d9/rest_framework_json_api/renderers.py#L460-L467

[...]

Issue #722, opened 2019-10-09
JSONRenderer does not work when no pk is provided


mdettelson:
Hi there!

I have a particular use case here: I have a denormalized table that contains one data type (let's call it 'pieces') that gets aggregated up to another model (let's call them 'pies').

Since I can't do anything with pies without having all of the pieces (technical requirement), I've decided to keep a table representing only the pieces and aggregate up to the pie level when pies are asked for.

I noticed this line in the renderer code: 

`('id', encoding.force_text(resource_instance.pk) if resource_instance else None),`
It's in the method `build_json_resource_obj`

Since my dataset has many repeats of the pie IDs, I can't use it as a pk, BUT once I aggregate the table the Pie Viewset, it IS unique.  Since it is not the actual pk on the model, however, I get the error `'dict' object has no attribute 'pk'`. 

The ideal situation would be that I can set the pie ID as the id in the JSON API representiation.

Is there a built-in way to handle this sort of use case?

sliverc:
I am not 100% sure about your use case. But when using a `ModelSerializer` this is currently not possible: look at discussion #155  But you could also use a python model and overwrite pk property see discussion at #704

Closing this as duplicate. If you think it is not a duplicate please comment or otherwise continue the discussion in the other issue if needed. Thanks.

mdettelson:
Ah ok, good to know!  Thanks for the references.

PR #1127, opened 2023-02-09
fix: serialization of non-model dict objects (fixes #155 #722 #1126)

### LINKED ISSUE #155: Support custom object keys

axieum:
Fixes #155 #722 #1126

## Description of the Change

Redirect logic used to extract the resource identifier from a serializer instance to a utility method. This utility method handles the edge-case where there is no model backing the serializer resulting in a "`dict` object has no attribute `pk`".

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/rest_framework_json_api/utils.py
+++ b/rest_framework_json_api/utils.py
@@ -308,6 +308,19 @@ def get_resource_type_from_serializer(serializer):
     )
 
 
+def get_resource_id_from_instance(instance):
+    """Returns the resource identifier for a given instance (`id` takes priority over `pk`)."""
+    if not instance:
+        return None
+    elif hasattr(instance, "id"):
+        return encoding.force_str(instance.id)
+    elif hasattr(instance, "pk"):
+        return encoding.force_str(instance.pk)
+    elif isinstance(instance, dict):
+        return instance.get("id", instance.get("pk", None))
+    return None
+
+
 def get_included_resources(request, serializer=None):
     """Build a list of included resources."""
     include_resources_param = request.query_params.get("include") if request else None
--- a/rest_framework_json_api/renderers.py
+++ b/rest_framework_json_api/renderers.py
@@ -459,10 +459,7 @@ class JSONRenderer(renderers.JSONRenderer):
             resource_name = utils.get_resource_type_from_instance(resource_instance)
         resource_data = [
             ("type", resource_name),
-            (
-                "id",
-                encoding.force_str(resource_instance.pk) if resource_instance else None,
-            ),
+            ("id", utils.get_resource_id_from_instance(resource_instance)),
             ("attributes", cls.extract_attributes(fields, resource)),
         ]
         relationships = cls.extract_relationships(fields, resource, resource_instance)
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -183,6 +183,28 @@ class TestAPIView:
             }
         }
 
+    @pytest.mark.urls(__name__)
+    def test_post_with_missing_id(self, client):
+        data = {
+            "data": {
+                "id": None,
+                "type": "custom",
+                "attributes": {"body": "hello"},
+            }
+        }
+
+        url = reverse("custom")
+
+        response = client.post(url, data=data)
+        assert response.status_code == status.HTTP_200_OK
+        assert response.json() == {
+            "data": {
+                "type": "custom",
+                "id": None,
+                "attributes": {"body": "hello"},
+            }
+        }
+
 
 # Routing setup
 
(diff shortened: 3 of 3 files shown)


sliverc:
Thanks for looking into our oldest open issue. It is great to tackle this.

I think it is a great idea to allow a DJA user to overwrite an `id` by configuring an `id` field on the serializer, as this is most intuitive. I think your code needs to be adjusted though, as in its current states it expects an `id` on the instance, and not on the serialized data.

A check would need to do the following:
1. check if there is an `id` in the serialized data (in the method `build_json_resource_obj` the serialized data is called `resource`).
2. if not fall back to the pk on the resource instance

That should be enough in my opinion and no more if/else are needed. It will also be important to document this behavior in our documentation, that setting an `id` field on the serializer is the way to have a custom `id` or when using generic serializers. To fully resolve #155 it also needs to be possible to configure a custom `pk` for a relation. When digging into the code, this is actually already possible by setting [pk_field](https://github.com/django-json-api/django-rest-framework-json-api/blob/main/rest_framework_json_api/relations.py#L251) on the `ResourceRelatedField`, so we would only need to document and add a test.

You think you could work on those things?

sliverc:
@axieum Do you think you can still work on this PR (open issues see comment above)? I very much like the implementation and would love to get it integrated.

axieum:
Hi, sorry about that @sliverc - got heavily distracted elsewhere.

I've gone ahead and quickly added some commits to do what you proposed, first it'll try the `resource` and then the `resource_instance`. In both scenarios, the `resource` (and `resource_instance`) may be a `dict` or `object`.

I couldn't think of how to write a test for the case where `resource` is preferred over `resource_instance`, could you assist here please?

Thanks,
Jonathan.

sliverc:
```python
def get_resource_id(resource_instance, resource):
   if resource and 'id' in resource:
      return force_str(resource['id'])
   return resource_instance and force_str(resource_instance.pk) or None
```

As we need to add a utility method anyway to get the resource id, force_str, checking for None etc. should all be in there. Above is how the method could look like (not tested, but as an idea). As passing on `resource_instance` and `resource` separately, it is actually much simpler.

For writing a test, you need to write a test where the serializer has a field called id. Easiest would then be to add a method get_id(self) to the serializer which returns a different id to see whether that id is actually returned. Then test it as you already do in the `test_views.py` file.

Hope this makes things clearer.

axieum:
I think we're going to need to explicitly add code to check for and execute a `get_id` method for that piece to work.

rest_framework_json_api/utils.py
```diff
@@ -304,6 +304,21 @@ def get_resource_type_from_serializer(serializer):
     )
 
 
+def get_resource_id_from_instance(resource_instance, resource):
+    """Returns the resource identifier for a given instance (`id` takes priority over `pk`)."""
+    if resource and "id" in resource:
+        return resource["id"] and encoding.force_str(resource["id"]) or None
+    if resource_instance:
+        if hasattr(resource_instance, "get_id") and callable(resource_instance.get_id):
```
sliverc, line 312: What is the use case to have a `get_id` method on the resource instance resp. model? The model is the database representation, whereas the `id` is the JSON:API spec representation. So if the `id` needs to be different from the database it should be defined on the serializer by adding a `id` field. That should be enough as far as I see it.

rest_framework_json_api/utils.py
```diff
@@ -304,6 +304,21 @@ def get_resource_type_from_serializer(serializer):
     )
 
 
+def get_resource_id_from_instance(resource_instance, resource):
```
sliverc, line 307: I would rename it now as it is not retrieving more than just an instance to `get_resource_id`.

sliverc (changes requested):
Thanks for continue working on this. See my inline comments + the tests do not seem to pass, therefore haven't reviewed those in detail yet.

sliverc:
This is a very important feature and as we are soon releasing a new version I would like to get this ready. So I am currently working on it to be able to merge it before our next release.

sliverc (approved):

sliverc:
I now added some documentation as well. For this feature to be complete, it is also necessary to be able to overwrite resource id for resource related fields. Added this now as well. So this is ready to merge. In case there are any comments of anything I missed, let me know.

PR #1132, opened 2023-02-21
Replaced OrderedDict with dict

sliverc:
Fixes #1094

## Description of the Change

As Python 3.7+ now ensures that dicts are ordered, there is no need to use `OrderedDict` anymore.

See in-depth explanation at https://github.com/django-json-api/django-rest-framework-json-api/pull/1094#issuecomment-1438038120

Using regular dicts also make the code more readable and easier debuggable.

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/rest_framework_json_api/renderers.py
+++ b/rest_framework_json_api/renderers.py
@@ -2,7 +2,7 @@
 Renderers
 """
 import copy
-from collections import OrderedDict, defaultdict
+from collections import defaultdict
 from collections.abc import Iterable
 
 import inflection
@@ -56,7 +56,7 @@ class JSONRenderer(renderers.JSONRenderer):
         """
         Builds the `attributes` object of the JSON:API resource object.
         """
-        data = OrderedDict()
+        data = {}
         for field_name, field in iter(fields.items()):
             # ID is always provided in the root of JSON:API so remove it from attributes
             if field_name == "id":
@@ -89,7 +89,7 @@ class JSONRenderer(renderers.JSONRenderer):
         # Avoid circular deps
         from rest_framework_json_api.relations import ResourceRelatedField
 
-        data = OrderedDict()
+        data = {}
 
         # Don't try to extract relationships from a non-existent resource
         if resource_instance is None:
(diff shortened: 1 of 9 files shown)


sliverc:
Removing `OrderedDict` is not an urgent task, however it makes the code more readable at a few instances. So I would love to remove it, but really want to avoid regressions. Codecov shows that there are a couple of code paths which I needed to change, which actually do not have any test coverage.

I would see three ways forward:
1. wait till #468 is completed
2. add test for those missing paths in this PR (could be cumbersome)
3. Force merge it as test coverage actually does not change. We would have to be fairly confident, that the source changes are not breaking anything.

@n2ygk Any thoughts?

n2ygk:
> Removing `OrderedDict` is not an urgent task, however it makes the code more readable at a few instances. So I would love to remove it, but really want to avoid regressions. Codecov shows that there are a couple of code paths which I needed to change, which actually do not have any test coverage.
> 
> I would see three ways forward:
> 
> 1. wait till [Convert all tests to pytest #468](https://github.com/django-json-api/django-rest-framework-json-api/issues/468) is completed
> 2. add test for those missing paths in this PR (could be cumbersome)
> 3. Force merge it as test coverage actually does not change. We would have to be fairly confident, that the source changes are not breaking anything.
> 
> @n2ygk Any thoughts?

@sliverc I think it will be pretty safe to replace `OrderedDict` and certainly improves readability, especially when it is invoked with a list of tuples vs. the much more readable dict syntax. A force merge in this case seems fine.

sliverc:
Sounds good. I made the PR ready then for merging. I went through all changes again, but please have a close look whether everything looks good to you.

rest_framework_json_api/renderers.py
```diff
@@ -457,37 +443,33 @@ def build_json_resource_obj(
...
+        resource_id = force_str(resource_instance.pk) if resource_instance else None
+        resource_data = {
+            "type": resource_name,
+            "id": resource_id,
+            "attributes": cls.extract_attributes(fields, resource),
+        }
         relationships = cls.extract_relationships(fields, resource, resource_instance)
         if relationships:
-            resource_data.append(("relationships", relationships))
+            resource_data["relationships"] = relationships
```
n2ygk, line 454: nice replacement of all these .append()s.

n2ygk (approved):
Looks good. Nice cleanup, tightening up the loops while eliminating `OrderedDict`.

n2ygk:
Per our conversation doing a force merge to get around failed codecov target.

PR #1137, opened 2023-03-07
Use /data as the pointer for non-field serializer errors

### LINKED ISSUE #1136: Object-level validation errors given an incorrect source.pointer

When raising a non-field `ValidationError` in a serializer [in accordance with the DRF documentation](https://www.django-rest-framework.org/api-guide/serializers/#object-level-validation), the error is attributed to a non-existing `/data/attributes/non_field_errors` path instead of `/data`.

```python
from rest_framework_json_api import serializers

class ProductSerializer(serializers.ModelSerializer):
    def validate(self, data):
        if data.get("best_before_date") and data.get("expiry_date"):
            raise serializers.ValidationError(
                "Specify either a best before date or an expiry date."
            )
        return data
```

```json
{
    "errors": [
        {
            "detail": "Specify either a best before date or an expiry date.",
            "status": "400",
            "source": {
                "pointer": "/data/attributes/non_field_errors"
            },
            "code": "invalid"
        }
    ]
}
```

Complete minimal example available [here](https://github.com/arttuperala/jsonapi_non_field_errors).

## Checklist

arttuperala:
Fixes #1136

## Description of the Change

Checks if a field matches [NON_FIELD_ERRORS_KEY](https://www.django-rest-framework.org/api-guide/settings/#non_field_errors_key) and sets the error pointer to `"/data"` if it does during DRF error formatting.

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -3,6 +3,7 @@ Adam Ziolkowski <adam@adsized.com>
 Alan Crosswell <alan@columbia.edu>
 Alex Seidmann <alex@leanpnt.de>
 Anton Shutik <shutikanton@gmail.com>
+Arttu Perälä <arttu@perala.me>
 Ashley Loewen <github@ashleycodes.tech>
 Asif Saif Uddin <auvipy@gmail.com>
 Beni Keller <beni@matraxi.ch>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ any parts of the framework not mentioned in the documentation should generally b
 ### Fixed
 
 * Refactored handling of the `sort` query parameter to fix duplicate declaration in the generated schema definition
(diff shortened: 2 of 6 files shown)


sliverc (changes requested):
Thanks for working on this. Appreciated. In general, looks good to me. Not sure that the test should belong there, but this needs to be cleaned up later anyway, so let's leave it like that.
Please add the non error field name to reserved field names https://github.com/django-json-api/django-rest-framework-json-api/blob/main/rest_framework_json_api/serializers.py#L158 even though it is very unlikely that someone uses it as a field name it is just good for completeness. Otherwise, good to merge.

sliverc (approved):
Thanks a lot for your great work on this issue and PR. Appreciated and any time again :smile:.

PR #1138, opened 2023-03-10
Revert SchemaGenerator description value from tuple to str

arttuperala:
## Description of the Change

Fixes a schema issue introduced in 9ced3be23c5b5c7772d11959148e350300e382d0 that would result in broken YAML schema.

```yaml
components:
  schemas:
    error:
      properties:
        source:
          properties:
            pointer:
              description: !!python/tuple
              - A [JSON Pointer](https://tools.ietf.org/html/rfc6901) to the associated
                entity in the request document [e.g. `/data` for a primary data object,
                or `/data/attributes/title` for a specific attribute.
```

Only present on the main branch so I didn't update the changelog.

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/example/tests/__snapshots__/test_openapi.ambr
+++ b/example/tests/__snapshots__/test_openapi.ambr
@@ -706,3 +706,548 @@
   }
   '''
 # ---
+# name: test_schema_construction
+  '''
+  {
+    "components": {
+      "parameters": {
+        "fields": {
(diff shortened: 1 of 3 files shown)


sliverc:
Great catch. I am confused though that our test_openapi.py did not catch this. Could you have a brief look at the test and see whether it is easy to adjust it so it covers this part of the schema as well?

arttuperala:
Converted the existing schema construction test to a snapshot one. Do you think that'll suffice?

sliverc (approved):
Awesome. Thanks, Yep, makes most sense to have this schema test as a snapshot test. Like this, it would have caught this error.

PR #1142, opened 2023-04-04
Add support for Django 4.2

Tenzer:
Allows library to be installed with Django 4.2.

## Description of the Change

This broadens the range of Django versions this library allows to be installed alongside to include Django 4.2 and any bugfix releases of it.

It also disables the use of the now deprecated `UnsaltedMD5PasswordHasher` hashing algorithm when running tests, as it otherwise would raise deprecation warnings. This will make the tests slightly slower since the default password hashing algorithm uses more resources, but it shouldn't have a big impact all in all as the tests run pretty quickly regardless.

## Checklist

- [ ] unit-test added
- [x] documentation updated

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,6 +2,7 @@
 envlist =
     py{37,38,39,310}-django32-drf{313,314,master},
     py{38,39,310}-django40-drf{313,314,master},
+    py{38,39,310,311}-django41-drf{314,master},
     py{38,39,310,311}-django42-drf{314,master},
     black,
     docs,
@@ -11,6 +12,7 @@ envlist =
 deps =
     django32: Django>=3.2,<3.3
     django40: Django>=4.0,<4.1
+    django41: Django>=4.1,<4.2
     django42: Django>=4.2,<4.3
     drf313: djangorestframework>=3.13,<3.14
     drf314: djangorestframework>=3.14,<3.15
@@ -52,6 +54,9 @@ commands =
 ignore_outcome = true
 
 [testenv:py{38,39,310}-django40-drfmaster]
+ignore_outcome = true
+
+[testenv:py{38,39,310,311}-django41-drfmaster]
 ignore_outcome = true
 
 [testenv:py{38,39,310,311}-django42-drfmaster]
--- a/tox.ini
+++ b/tox.ini
@@ -2,7 +2,7 @@
 envlist =
     py{37,38,39,310}-django32-drf{313,314,master},
     py{38,39,310}-django40-drf{313,314,master},
-    py{38,39,310,311}-django41-drf{314,master},
+    py{38,39,310,311}-django42-drf{314,master},
     black,
     docs,
     lint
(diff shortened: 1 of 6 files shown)


tox.ini
```diff
@@ -2,7 +2,7 @@
 envlist =
     py{37,38,39,310}-django32-drf{313,314,master},
     py{38,39,310}-django40-drf{313,314,master},
-    py{38,39,310,311}-django41-drf{314,master},
```
sliverc, line 5: we should not remove Django 4.1 from the testgrid

tox.ini
```diff
@@ -11,7 +11,7 @@ envlist =
 deps =
     django32: Django>=3.2,<3.3
     django40: Django>=4.0,<4.1
-    django41: Django>=4.1,<4.2
```
sliverc, line 14: as above Django 4.1 should remain

tox.ini
```diff
@@ -54,5 +54,5 @@ ignore_outcome = true
 [testenv:py{38,39,310}-django40-drfmaster]
 ignore_outcome = true
 
-[testenv:py{38,39,310,311}-django41-drfmaster]
```
sliverc, line 57: Django 4.1 should remain

sliverc (changes requested):
Thanks for working on this. Looks good overall just make sure you do not drop tests for Django 4.1.

Tenzer:
Thank you for the review. I have reinstated the Django 4.1 tests now.

sliverc (approved):
Thanks. It looks good now.

Tenzer:
Great, thanks! Do you know when a new release might be cut? This would be blocking any users of the library from upgrading to Django 4.2.

sliverc:
This kind of depends on when DRF will release a new version supporting Django 4.2. @auvipy Do you have more information when there will be a new DRF release?

Tenzer:
It looks like https://github.com/encode/django-rest-framework/pull/8794 is the relevant issue to track regards a new release.

auvipy:
> This kind of depends on when DRF will release a new version supporting Django 4.2. @auvipy Do you have more information when there will be a new DRF release?

I contacted with Tom but he seem to be busy. I hoe we can cut a release in late this month or early next month.

abmohan:
Thanks for working on this. I'm looking forward to upgrading to django 4.2. What's the release schedule looking like? How long do you think we'd have to wait for this?

sliverc:
@abmohan We are still waiting on DRF to release 3.15 with Django 4.2 support. @auvipy Do you have any update when this will be ready? Thanks.

elanqo:
@sliverc, When do you anticipate publishing these changes to PyPI?

max-wittig:
@sliverc We're using DRF with 4.2 in production from the latest release. According to their README it should be already supported. Sorry, if I missed something here and just jumped in without knowning something important. Just wanted to state the obvious (it works with 4.2) 😄

sliverc:
DRF 3.15 seems to take longer than anticipated. So I released it [now](https://github.com/django-json-api/django-rest-framework-json-api/releases/tag/v6.1.0).

PR #1156, opened 2023-06-08
Omit id field from OpenAPI schema attributes

arttuperala:
## Description of the Change

Omits the "id" field from `/data/attributes` the OpenAPI schema, as the `JSONRenderer` [explicitly removes it from the attributes](https://github.com/django-json-api/django-rest-framework-json-api/blob/main/rest_framework_json_api/renderers.py#L61-L63).

```python
{
    "type": "object",
    "required": ["type", "id"],
    "additionalProperties": False,
    "properties": {
        "type": {"$ref": "#/components/schemas/type"},
        "id": {"$ref": "#/components/schemas/id"},
        "links": { ... },
        "attributes": {
            "type": "object",
            "properties": {
                "id": {"type": "integer", "readOnly": True},  # <- Not real
                "name": {"type": "string", "maxLength": 100},
            },
            "required": ["name"],
        },
        "relationships": { ... },
    },
}
```

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ any parts of the framework not mentioned in the documentation should generally b
 
 * Refactored handling of the `sort` query parameter to fix duplicate declaration in the generated schema definition
 * Non-field serializer errors are given a source.pointer value of "/data".
+* Fixed "id" field being added to /data/attributes in the OpenAPI schema when it is not rendered there.
 
 ## [6.0.0] - 2022-09-24
 
--- a/example/tests/test_openapi.py
+++ b/example/tests/test_openapi.py
@@ -110,6 +110,21 @@ def test_schema_construction(snapshot):
     assert snapshot == json.dumps(schema, indent=2, sort_keys=True)
 
 
+def test_schema_id_field():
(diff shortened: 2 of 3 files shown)


sliverc (approved):

sliverc:
Thanks for working on this. In case others stumble upon this as well and as I mentioned in #1157. OpenAPI schema support will be deprecated soon in DJA (reason being is that DRF has also deprecated it). We will still add small changes like this, we do recommend to use [drf-spectecular-json-api](https://github.com/jokiefer/drf-spectecular-json-api) though. That library is just getting started, so it would certainly be great to support it.

arttuperala:
I'll have to look at migrating to that library at some point. We're currently only using the built-in DJA schema generation at my workplace and we figured that we'd contribute some of the fixes we've had to do during our implementation to upstream as well.

sliverc:
That is great. Very much appreciated your contribution. It would be great if you could take a look at [drf-spectecular-json-api](https://github.com/jokiefer/drf-spectecular-json-api) as it could certainly need some support.

PR #1161, opened 2023-06-27
Fix many serializer method fields having the incorrect relation schema

arttuperala:
## Description of the Change

`SerializerMethodResourceRelatedField(many=True)` (`ManySerializerMethodResourceRelatedField`) fields would be documented in the OpenAPI schema with "#/components/schemas/reltoone" reference instead of "#/components/schemas/reltomany" reference.

## Checklist

- [x] unit-test added
- [ ] documentation updated


sliverc (approved):
Thanks. Nice catch. Merging.

PR #1162, opened 2023-06-28
Include meta section for SerializerMethodResourceRelatedField(many=True)

### LINKED ISSUE #572: SerializerMethodResourceRelatedField does not include meta section in relationship response.
`SerializerMethodResourceRelatedField` does not include a meta section in the relationship response like `ResourceRelatedField` does. This doesn't make sense, because `SerializerMethodResourceRelatedField` extends `ResourceRelatedField` and both seem to return every related record in the response (for the record I've only tried with a little over a thousand related records).

So for example if you had an ObjectA with a has many relationship with ObjectB:

```python
    objectbs = SerializerMethodResourceRelatedField(
        source='get_objectbs',
        many=True,
        model=ObjectB,
        read_only=True
    )

    def get_objectbs(self, obj):
        return ObjectB.objects.filter(objecta=obj)
```

then your response would look like:

```json
"relationships": {
  "objectb": {
   "data": [...]
  }
}
```

instead of:

```json
"relationships": {
  "objectb": {
    "meta": {
      "count": 2
    },
   "data": [...]
  }
}
```

arttuperala:
Fixes #572

## Description of the Change

`SerializerMethodResourceRelatedField(many=True)` (`ManySerializerMethodResourceRelatedField`) fields are given a meta section with a count much like `ManyRelatedField`.

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,8 @@ any parts of the framework not mentioned in the documentation should generally b
             return value.name
     ```
 
+* `SerializerMethodResourceRelatedField(many=True)` relationship data now includes a meta section.
+
 ### Fixed
 
 * Refactored handling of the `sort` query parameter to fix duplicate declaration in the generated schema definition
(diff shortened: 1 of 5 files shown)


sliverc (approved):
Thanks for working on this very old issue. Appreciated. Looks good. Merging.

PR #1165, opened 2023-07-03
Add HTTP 400 Bad Request as a possible generic error response

arttuperala:
## Description of the Change

Adds `400 Bad Request` as a possible error response, as JSON:API specification says to use it for requests with bad `include` / `sort` or otherwise illegal query parameters, and advocates using it as an appropriate status code for multiple 4xx errors. DJA's documentation also mentions `400 Bad Request` multiple times, so it's a bit weird if it's not documented as a possible response.

Discovered during an OpenAPI schema validation run as our generic error responses were not found in the schema.

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ any parts of the framework not mentioned in the documentation should generally b
 
 * Added support for Python 3.11.
 * Added support for Django 4.2.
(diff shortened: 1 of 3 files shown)


sliverc (approved):
Thanks. Indeed, that makes sense. Merging.

PR #1167, opened 2023-07-12
Exclude callable field defaults from the OpenAPI schema

arttuperala:
## Description of the Change

Implements [the same sort of callable field.default exclusion that DRF implements](https://github.com/encode/django-rest-framework/blob/master/rest_framework/schemas/openapi.py#L538), as having a callable field default in the OpenAPI schema results in broken YAML output.

```yaml
field:
  type: string
  writeOnly: true
  default: !!python/object:rest_framework.fields.CreateOnlyDefault
    default: null
```

## Checklist

- [x] unit-test added
- [ ] documentation updated


tests/test_schema.py
```diff
@@ -0,0 +1,10 @@
+from rest_framework_json_api.schemas.openapi import AutoSchema
```
sliverc, line 1: Great that you added a test to our new tests folder.

What we try to do here is that the test file is in the same module structure as the python file it is testing.

So this file should be `tests/schemas/test_openapi.py`

tests/test_schema.py
```diff
@@ -0,0 +1,10 @@
...
+from tests.serializers import CallableDefaultSerializer
+
+
+def test_schema_callable_default():
```
sliverc, line 5: If the test is testing part of a class we also use a class to group the test.

So the test would look like this:

```
class TestAutoSchema:
    def test_map_serializer_with_callable_default(self):
        ...
```

sliverc (approved):
Thanks for this addition. Well spotted. See my inline comments for some small adjustments.

tests/test_schema.py
arttuperala: Moved.

tests/test_schema.py
arttuperala: Punted into a class.

sliverc:
Great. Thanks for your work. Merged.

PR #1175, opened 2023-09-07
Fix Serializer schema generation when used as a ListField child

arttuperala:
## Description of the Change

Fixes schema generation for `Serializer` when it's used as a child of a `ListField`. Previously it'd use the full JSON:API serializer schema even though it doesn't render the "type", "id", "attributes", etc. fields in the response.

Also added a unit test for API response when using the `ListField` with a `Serializer` child while I was at it since there didn't seem to be one. This way it'll be a bit better defined how these work in DJA.

## Checklist

- [x] unit-test added
- [ ] documentation updated


sliverc:
Thanks for the contribution. At a look of it, this exact problem also occurs with nested serializers right? If yes, could you adjust the PR, so that is covered as well?

arttuperala:
Yeah, we ran across the nested serializer issue right aftewards. I had actually fixed it on another branch, but since I wasn't sure if it met the "PR only contains one change" criteria and since it was based off of this branch, I didn't have a PR open for it. Submitted that branch separately as #1177.

PR #1177, opened 2023-09-19
Fix schema generation for nested serializers

arttuperala:
## Description of the Change

Fixes schema generation for `Serializer` when it's used as a nested serializer inside another serializer or as a child of a `ListField`. Previously it'd use the full JSON:API serializer schema even though it doesn't render the "type", "id", "attributes", etc. fields in the response.

## Checklist

- [x] unit-test added
- [ ] documentation updated


sliverc (approved):
Awesome. Thanks for your work. LGTM.

PR #1185, opened 2023-10-19
Added Python 3.12 support

sliverc:
## Description of the Change

Added support for Python 3.12.

## Checklist

- [ ] unit-test added
- [x] documentation updated

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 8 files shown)


sliverc:
@n2ygk 
Changes to support Python 3.12 have been merged in polymorphic, However I do not think a version will be released soon there.

Therefore, I temporarily set the requirement pointing to master branch of django polymorphic. This way, we are not getting blocked with adding Python 3.12 support. Once a new polymorphic version has been released, we can then pin it again.

Ready for review.

n2ygk (approved):

PR #1221, opened 2024-04-19
Ensured that sparse fields only applies when rendering not when parsing

### LINKED ISSUE #332: Sparse fields nulling meta fields instead of removing them

### LINKED ISSUE #1216: Sparse fields set should only drop fields when reading

As discussed in #1214 there is an issue with `SparseFieldsetsMixin` serializer mixin. It drops fields when reading and writing, however it should only do it when reading and not when writing as per [specification](https://jsonapi.org/format/#fetching-sparse-fieldsets).

When analyzing the code, it seems the best to implement is to move the exclude logic from `__init__` and overwrite `_readable_fields`. 

Related to #1053 

## Checklist

sliverc:
Fixes #1216
Fixes #332

## Description of the Change

The rendering of objects is flawed as the DJA renderer does most of the serialization again which should actually only be done by the serializer. Even though I reduced the readable fields in the serializer when sparse fields is used, it did not work.

I already adjusted `extract_attributes` that it uses the fields provided by the serializer. To adjust `extract_relationships` is more complex. Therefore, to get this critical issue fixed I also added code to exclude fields in the renderer when sparse fields is used.

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ any parts of the framework not mentioned in the documentation should generally b
 * Avoided that an empty attributes dict is rendered in case serializer does not
   provide any attribute fields.
 * Avoided shadowing of exception when rendering errors (regression since 4.3.0).
+* Ensured that sparse fields only applies when rendering, not when parsing.
+* Adjusted that sparse fields properly removes meta fields when not defined.
 
 ### Removed
 
--- a/example/tests/integration/test_non_paginated_responses.py
+++ b/example/tests/integration/test_non_paginated_responses.py
@@ -14,7 +14,7 @@ def test_multiple_entries_no_pagination(multiple_entries, client):
     expected = {
         "data": [
             {
-                "type": "posts",
+                "type": "entries",
                 "id": "1",
                 "attributes": {
                     "headline": multiple_entries[0].headline,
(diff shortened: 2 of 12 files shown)


sliverc:
Codecov does not pass as it highlights two lines which I did change in this pull request but only changing the import not the functionality. There was no test for these lines before. I would say we ignore it, but I am open to revert the change to clean up the utils import in the renderer.

n2ygk (approved):

PR #1121, opened 2023-01-19
Fix: TypeError: object of type 'NoneType' has no len()

ericinfins:
TypeError: object of type 'NoneType' has no len()

Fixes #

Changed the sequence

## Checklist

- [ ] PR only contains one change (considered splitting up PR)
- [ ] unit-test added
- [ ] documentation updated
- [ ] `CHANGELOG.md` updated (only for user relevant changes)
- [ ] author name in `AUTHORS`

PR #1229, opened 2024-05-31
fix the format_errors bug when data is None

ericinfins:
Fixes #

Related to the PR: https://github.com/django-json-api/django-rest-framework-json-api/pull/1121
Add the test case to support it

AUTHORS:
 - Eric

## Description of the Change

Fix: TypeError: object of type 'NoneType' has no len()

## Checklist

- [ x ] PR only contains one change (considered splitting up PR)
- [x  ] unit-test added
- [ ] documentation updated
- [x ] author name in `AUTHORS`

### DIFF
--- a/example/tests/unit/test_default_drf_serializers.py
+++ b/example/tests/unit/test_default_drf_serializers.py
@@ -1,5 +1,6 @@
 import json
 from datetime import datetime
+import logging
 
 import pytest
 from django.urls import reverse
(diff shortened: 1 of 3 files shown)


sliverc:
Thanks for following up on this. I will take a look at it at some point and will give you feedback. Just writing to let you know that I saw it.

sliverc:
I have finally got around to take a look at this. Great to see how this can be reproduced by using vanilla DRF classes.

It is though as I expected that this fix does avoid the exception, however the output of the fix is not [JSON:API compliant](https://jsonapi.org/format/#errors). It simply returns
```python
{
  "errors": None
}
```
but the value of errors needs to be a list and each error object must at least have one member. I think the easiest is to add a single error object with the status code as a string.

PR #1240, opened 2024-06-27
Ensured that no fields are returned when sparse fields is set to an empty value

### LINKED ISSUE #1238: Sparse field set should not return all fields when value is empty

The [JSON:API spec](https://jsonapi.org/format/#fetching-sparse-fieldsets) states

> The value of any fields[TYPE] parameter MUST be a comma-separated (U+002C COMMA, “,”) list that refers to the name(s) of the fields to be returned. An empty value indicates that no fields should be returned.

Currently, DJA returns all fields though when value is empty, but it should return none. None meaning no attributes and relationships but id and type remaining.

## Checklist

sliverc:
Fixes #1238

## Description of the Change

When a sparse field is set to an empty value no fields (attributes/relationships) should be rendered whereas currently it rendered all fields.

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ any parts of the framework not mentioned in the documentation should generally b
 ### Fixed
 
 * Re-enabled overwriting of url field (regression since 7.0.0)
-* Ensured that no fields are rendered when sparse fields is set to an empty value.
+* Ensured that no fields are rendered when sparse fields is set to an empty value. (regression since 7.0.0)
 
 ## [7.0.1] - 2024-06-06
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ any parts of the framework not mentioned in the documentation should generally b
 ### Fixed
 
 * Re-enabled overwriting of url field (regression since 7.0.0)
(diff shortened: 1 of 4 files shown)


n2ygk (commented):
I think the current behavior is supported by the spec:
```

If a client does not specify the set of fields for a given resource type, the server MAY send all fields, a subset of fields, or no fields for that resource type.
```

https://jsonapi.org/format/
8.4 Sparse Fieldsets

It could break a lot of implementations. Perhaps add a setting.

sliverc:
I agree that the sentence you mentioned is a bit ambiguous, however it would not make any sense to first specify that `The value of any fields[TYPE] parameter MUST` and then afterward say that the `server MAY`. So the sentence you mention is, as far as I see it, if there is no query parameter set at all for a given resource type, but not if it is explicitly set to an empty value.

In any case, I have double-checked what was the behavior before change #1221. It turned out when defining an empty value it would also return no fields. So this means it is a regression and this PR will revert to the behavior we had before version 7.0.0. I have clarified this in the changelog as well now.

n2ygk (approved):

sliverc:
Thanks @n2ygk for merging. Let me prepare a release to get those regressions out of the way.

PR #1245, opened 2024-07-22
Handle zero as valid pk/id in get_resource_id util method

### LINKED ISSUE #1244: id column integer primary key of 0 gets returned as null
If a model has an id column that's an integer primary key
```python
class Person(models.Model):
    id = models.IntegerField(primary_key=True)
    name = models.CharField(max_length=200)
```
and a fixture is loaded where the id has value 0
```json
[
    {"model": "project.person", "fields": {"id": 0, "name": "Example"}},
    {"model": "project.person","fields": {"id": 1, "name": "Someone"}}
]
```
then the JSON response returned from the API replaces the id of 0 with null
```json
    "data": [
        {
            "type": "person",
            "id": null,
            "attributes": {
                "name": "Example"
            }
        },
        {
            "type": "person",
            "id": "1",
            "attributes": {
                "name": "Someone"
            }
        }
    ],
```

humayunah:
Fixes #1244 

## Description of the Change
- handle zero as valid pk/id in get_resource_id util method
- update tests to include cases where ID is zero

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -15,6 +15,7 @@ David Guillot, for Contexte <dguillot@contexte.com>
 David Vogt <david.vogt@adfinis-sygroup.ch>
 Felix Viernickel <felix@gedankenspieler.org>
 Greg Aker <greg@gregaker.net>
+Humayun Ahmad <humayunahbh@gmail.com>
 Jamie Bliss <astronouth7303@gmail.com>
 Jason Housley <housleyjk@gmail.com>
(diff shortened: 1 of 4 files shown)


humayunah:
@sliverc please review this and let me know if anything needs to be changed/added.

sliverc (approved):
Great job. Thanks for your work. I have only updated it slightly to avoid double lookup (hasattr) for performance reasons. Ready for merging.

PR #1251, opened 2024-09-10
Ensured that patching a To-Many relationship correctly raises request error

sliverc:
## Description of the Change

As of DRF upstream change https://github.com/encode/django-rest-framework/pull/9455 our current tests with DRF master are failing. The reason being is that previous to the above change, parser attribute errors would get swallowed.

After this upstream change it occurred that we actually have an error in our code that an attribute error gets hidden and instead of raising a 500-error the code would just continue.

This happened when patching a To-Many relationship with `RelationshipView` with `None`. However, the [JSON:API spec](https://jsonapi.org/format/#crud-updating-to-many-relationships) defines that for emptying a To-Many relationship, caller needs to pass empty array instead of None.

So this PR ensures that when passing on `None` when patching To-Many relationship, a proper request error (400) is returned instead of 500.

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ any parts of the framework not mentioned in the documentation should generally b
 ### Fixed
 
 * Handled zero as a valid ID for resource (regression since 6.1.0)
+* Ensured that patching a To-Many relationship with the `RelationshipView` correctly raises request error when passing in `None`.
+  For emptying a To-Many relationship an empty array should be used as per [JSON:API spec](https://jsonapi.org/format/#crud-updating-to-many-relationships)
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


sliverc:
To see failing drf master tests, go to https://github.com/django-json-api/django-rest-framework-json-api/actions/runs/10648325051/job/29517253464 open Run tox targets and scroll down to drfmaster tests.

These tests have been failing for a while and as we ignore failing drfmaster tests, it is hard to notice it. Possibly have to think about it not to ignore it (as it hardly ever happens that it fails). And if one PR suddenly fails because of DRF changes, we can always force merge it.

Anyway, there might be also other approaches, but that would be part of another PR/discussion.

n2ygk (approved):

PR #1178, opened 2023-09-19
Removed support for Python 3.7

sliverc:
## Description of the Change

Python 3.7 is end of life since 27 Jun 2023, so we can remove support for it.

## Checklist

- [x] PR only contains one change (considered splitting up PR)
- [ ] unit-test added
- [x] documentation updated
- [x] `CHANGELOG.md` updated (only for user relevant changes)
- [x] author name in `AUTHORS`

PR #1266, opened 2025-01-07
Removed support for Python 3.8

n2ygk:
## Description of the Change
Remove Python 3.8 which is EOL per [PEP 569](https://peps.python.org/pep-0569/)

## Checklist

- [ ] unit-test added
- [x] documentation updated

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
     env:
       PYTHON: ${{ matrix.python-version }}
     steps:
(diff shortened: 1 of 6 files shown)


n2ygk:
drfmaster tests are throwing an error for TypeError for encode of None. Maybe related to #1229?

sliverc:
There is an upstream change in DRF which caused these errors. I have submitted a PR here https://github.com/encode/django-rest-framework/pull/9615 hope it will be merged soon.

sliverc (changes requested):
Thanks for working on this. getting-started.md needs also to be adjusted I think and also add a note in changelog to version 7.1.0 that states that it is the last version supporting 3.7. See how we did it in #1178

n2ygk:
@sliverc 

> Thanks for working on this. getting-started.md needs also to be adjusted I think and also add a note in changelog to version 7.1.0 that states that it is the last version supporting 3.7. See how we did it in #1178

Done.

Also, perhaps as a separate PR, should we make a failed drfmaster test a non-fatal workflow error? That's how we do it in DOT, for example: https://github.com/jazzband/django-oauth-toolkit/blob/master/tox.ini#L57-L59. Right now, errors in non-released DRF are blocking successful PRs here.

n2ygk:
See #1267

sliverc (approved):

sliverc:
Thanks looks good.

PR #1267, opened 2025-01-09
ignore drfmaster-induced test failures

n2ygk:
## Description of the Change

Errors introduced into unreleased drfmaster are causing test failures. This splits out the drfmaster tests to ignore
errors. 

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,8 +1,10 @@
 [tox]
 envlist =
-    py{38,39,310,311,312}-django42-drf{314,315,master},
-    py{310,311,312}-django{50,51}-drf{314,315,master},
-    py313-django51-drf{master},
+    py{38,39,310,311,312}-django42-drf{314,315},
+    py{310,311,312}-django{50,51}-drf{314,315},
+    py{38,39,310,311,312}-django42-drfmaster,
+    py{310,311,312}-django{50,51}-drfmaster,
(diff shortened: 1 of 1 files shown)


sliverc:
Thanks for the suggestion. We actually removed those ignore outcome deliberately in PR #1256. The reason is that with this approach when our tests fail against drfmaster it is really hard to notice. An error is hidden in a very long log of a job which shows it is successful, so no need to browse through it normally

Previously, our tests were failing against drfmaster, and we did not notice it for many weeks. Thankfully, by chance, I noticed it and was able to fix it in #1251 before DRF got a released. So this ignore_outcome is not an option for me as it is almost the same as not testing against drfmaster at all.

We could of course change our configuration the way, that the drfmaster tests will have its own check in the GitHub check list, which we mark as optional. That way we quickly see when drfmaster is failing but it would still be green to merge.

However, this, I feel, is a lot of maintenance work as we have to duplicate configuration from tox into the GitHub action.

Actually this time we got the error quickly fixed upstream and there were no urgent PRs to merge on our end so we were not really blocked. And if we had an urgent PR we could still force merge it.

So I suggest we leave it as is and deal with the error once it occurs (which is usually very rarely).

PR #1327, opened 2026-07-19
Fix broken commands in getting-started docs

### LINKED ISSUE #1326: ModuleNotFoundError on running the example app
Trying to run the example app as described in
https://django-rest-framework-json-api.readthedocs.io/en/stable/getting-started.html#running-the-example-app
throws the error
`ModuleNotFoundError: No module named 'example'`
while executing the step
`django-admin migrate --settings=example.settings`

danialsiavashani:
django-admin isn't always on PATH after a fresh venv install, and requirements.txt doesn't exist anymore. Switched to python -m django and pip install -e ., which I tested and confirmed work.

Fixes #1326

## Description of the Change
The "Running the example app" section in getting-started.md instructed
users to run `django-admin migrate`, but `django-admin` is not
guaranteed to be on PATH after a fresh venv install, which causes
`ModuleNotFoundError: No module named 'example'`. It also referenced
`pip install -Ur requirements.txt`, which no longer exists in this repo.

I reproduced the bug locally, confirmed `python -m django migrate` and
`pip install -e .` work correctly instead, and updated the docs
accordingly.

## Checklist
- [ ] unit-test added
- [x] documentation updated

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -91,10 +91,10 @@ and add `rest_framework_json_api` to your `INSTALLED_APPS` setting below `rest_f
 	cd django-rest-framework-json-api
 	python3 -m venv env
 	source env/bin/activate
-	pip install -Ur requirements.txt
-	django-admin migrate --settings=example.settings
-	django-admin loaddata drf_example --settings=example.settings
-	django-admin runserver --settings=example.settings
+	pip install -e .
+	python -m django migrate --settings=example.settings
+	python -m django loaddata drf_example --settings=example.settings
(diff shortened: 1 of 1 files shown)


sliverc:
Thanks for opening this PR. This is not a fix though because if requirements are not installed it won't work (e.g. tests etc.)

Anyhow on the [README.md](https://github.com/django-json-api/django-rest-framework-json-api#running-the-example-app) there is a section on how to run the example app which should work and this section should be synchronized to getting-started.md so it is the same in both files.

sliverc (approved):
Thanks.

